Issue #54, opened 2026-05-11
Marketing / landing: expand feature set and strengthen CTAs with full product story


ZappoMan:
## Summary

The public-facing marketing experience (e.g. B2C landing at `/` for guests) should **accurately and persuasively communicate BeakerStack’s value**—not only as a “starter app,” but as a **monorepo + Supabase + web/mobile + delivery pipeline** template. This issue covers **adding or refreshing on-page “feature” content** and **improving primary/secondary CTAs** so they reflect the full story, including items we already want to highlight and additional differentiators from the template docs.

---

## Goals

1. **Feature sections** (or equivalent blocks) should cover, at minimum, the themes below—using clear, non-jargony copy and consistent structure (e.g. config-driven sections if we keep the `landing.ts` pattern).
2. **CTAs** (hero, mid-page, footer, pricing, final CTA) should **tie actions to outcomes** (e.g. “See the repo,” “Read the quickstart,” “View architecture,” “Try preview deploys”) and **surface the same pillars** so no single CTA reads like a narrow “sign up only” funnel unless that is intentional.
3. **Avoid stale or unverifiable claims:** tie numeric claims (e.g. test coverage) to **what CI publishes** or soften to “high unit-test coverage with documented gates” if we don’t snapshot a number on the page.

---

### Already prioritized (must ship)

- **CI/CD and environments:** PR **preview**, **staging**, and **production** flows aligned with the repo’s branching/deployment model.
- **Test discipline:** **high unit-test coverage** (e.g. **>95%** where accurate per current CI/coverage reports), plus the broader story that the template also uses **integration, E2E, and database** tests as documented.

### Additional pillars (should ship)

- **One product, two surfaces:** **Web (React + Vite)** and **mobile (React Native + Expo)** from a **monorepo**, with **shared packages** so core logic and types don’t drift.
- **Supabase-centric backend:** Postgres, Auth, Storage, Edge Functions; **migrations at repo root**; **generated TypeScript types** from the schema; **local Supabase (Docker)** for parity with remotes.
- **Environment-safe data story:** **Local → PR testing → staging → production** Supabase projects (and why that matters for **migrations** and safe iteration).
- **Review and deploy UX:** **Path-based PR previews** on AWS static hosting; **EAS Update channels** for mobile aligned to preview/staging/production when configured.
[...]

PR #64, opened 2026-05-11
fix: AI code quality findings — landing copy, test mocks, jest regex (closes #62)

### LINKED ISSUE #62: Fix AI code quality findings
in apps/web/src/config/landing.ts:

This description is overly long and contains multiple distinct points. Consider breaking it into separate sentences or condensing the content to improve readability and maintainability.

        body: 'Postgres, Auth, Storage, and Edge Functions — with RLS-oriented patterns, migrations at repo root, generated TypeScript types from the schema, and local Supabase via Docker for dev/test parity.',

recommended:
        body: 'Postgres, Auth, Storage, and Edge Functions with RLS-first patterns, schema-generated TypeScript types, and local Docker parity for development and testing.',

This description contains multiple sentences that could be separated for better readability. Consider splitting into distinct points or using a more structured format.

      body: 'Shared auth logic, shared billing hooks, shared business rules — your React web app and React Native mobile app stay in sync without duplicating code. 40–60% of the codebase is shared across platforms.',

recommended:

mobile app stay in sync without duplicating code. 40–60% of the codebase is shared across platforms.', | body: 'Shared auth logic. Shared billing hooks. Shared business rules. Your React web app and React Native mobile app stay in sync without duplicated work. About 40–60% of the codebase is shared across platforms.',

This description is dense with multiple complex ideas. Consider breaking it into separate sentences or simplifying to improve clarity and maintainability.

      body: 'Three Supabase databases — local Docker, shared PR testing, and staging — mirror your branching model so migrations are validated before they touch production. PR previews deploy to path-based S3/CloudFront URLs automatically. EAS Update channels give mobile the same preview → staging → production story.',

recommended:

      body: 'Three Supabase databases—local Docker, shared PR testing, and staging—mirror your branching model. Migrations are validated before they touch production. PR previews deploy automatically to path-based S3/CloudFront URLs. EAS Update channels give mobile the same preview → staging → production flow.',

This description contains multiple distinct points. Consider breaking into separate sentences or condensing for better readability.

      body: 'A single `npm run setup` command provisions your local environment. QUICKSTART, ARCHITECTURE, and per-topic guides (OAuth, Stripe, testing) cover every decision. Environment variables and secrets follow a consistent layout that maps directly to GitHub Actions — no hunting for where a key goes.',

recommended:

      body: 'Run `npm run setup` to provision your local environment in one step. QUICKSTART, ARCHITECTURE, and focused guides (OAuth, Stripe, testing) explain key decisions. Env vars and secrets use a consistent layout that maps directly to GitHub Actions.',

=====

in apps/web/src/pages/__tests__/HomePage.test.tsx

The subscribe mock callback returns an empty object, but Supabase channel subscriptions typically return an object with an `unsubscribe` method. This could lead to issues if the code under test attempts to call `unsubscribe()` on the returned subscription. Consider returning `{ unsubscribe: vi.fn() }` instead.

      subscribe: vi.fn(cb => { cb('SUBSCRIBED'); return {}; 

recommended:

      subscribe: vi.fn(cb => { cb('SUBSCRIBED'); return { unsubscribe: vi.fn() }; }),

This subscribe mock has the same issue as line 25 - it returns an empty object instead of an object with an `unsubscribe` method. This duplication suggests the mock setup could be centralized to avoid inconsistencies.

Using `setTimeout(resolve, 0)` as a way to wait for state updates is fragile and can lead to flaky tests. Consider using `waitFor` from testing-library or `vi.waitFor` to wait for specific conditions instead of arbitrary delays.

      await new Promise(resolve => setTimeout(resolve, 0));

recommended:

await waitFor(() => {
--
expect(
    screen.queryByText('Build the full stack. Not the scaffolding.')
    screen.queryByText('Dashboard')
    ).toBeTruthy();
});

=====

in packages/shared-tests/jest.web.cjs

The regex pattern `(?!^$)` creates a negative lookahead for an empty string, which means this pattern will match all node_modules except those with an empty module name (which don't exist). This effectively transpiles all node_modules, which is likely unintended and could cause performance issues. The standard pattern to exclude all node_modules is `node_modules/(?!.*)` or simply omit this configuration to use Jest's default.

"node_modules/(?!^$)" // default; adjust to transpile specific libs if needed

recommended:

  "/node_modules/" // default; adjust to transpile specific libs if needed

diego-artificerinnovations:
## Summary

- **`apps/web/src/config/landing.ts`** — condense 4 feature row body strings per AI recommendations: shorter sentences, active voice, removed redundant clauses
- **`apps/web/src/pages/__tests__/HomePage.test.tsx`** — two fixes:
  - Subscribe mock now returns `{ unsubscribe: vi.fn().mockResolvedValue(undefined) }` so cleanup code that chains `.catch()` on `unsubscribe()` doesn't throw (the empty `{}` return was a latent crash in the authenticated redirect test)
  - Replace `setTimeout(resolve, 0)` flush with `waitFor()` outside `act()` — retries until the DOM settles rather than relying on an arbitrary delay
- **`packages/shared-tests/jest.web.cjs`** — clean up `transformIgnorePatterns` regex: `node_modules/(?!^$)` → `/node_modules/`. The original pattern was functionally correct (node_modules was excluded from transpilation) but the negative-lookahead on an empty string anchor was confusing and non-idiomatic. Replaced with the standard Jest pattern for readability.

Closes #62

## Test plan

- [x] `HomePage.test.tsx` — 5/5 passing locally after fixes
- [ ] CI green

### DIFF
--- a/apps/web/src/config/landing.ts
+++ b/apps/web/src/config/landing.ts
@@ -204,7 +204,7 @@ export const landingConfig: LandingConfig = {
     },
     {
       title: 'Web and mobile from a single codebase.',
-      body: 'Shared auth logic. Shared billing hooks. Shared business rules. Your React web app and React Native mobile app stay in sync without duplicated work. About 40–60% of the codebase is shared across platforms.',
+      body: 'Shared auth logic. Shared billing hooks. Shared business rules. Your React web app and React Native mobile app stay in sync without duplicating work. About 40–60% of the codebase is shared across platforms.',
       ctaLabel: 'Learn about mobile',
       ctaHref: '#features',
       mediaSrc: 'https://placehold.co/560x315?text=Mobile+App',
--- a/apps/web/src/config/landing.ts
+++ b/apps/web/src/config/landing.ts
@@ -161,7 +161,7 @@ export const landingConfig: LandingConfig = {
       {
         icon: Database,
         title: 'Supabase backend',
-        body: 'Postgres, Auth, Storage, and Edge Functions — with RLS-oriented patterns, migrations at repo root, generated TypeScript types from the schema, and local Supabase via Docker for dev/test parity.',
+        body: 'Postgres, Auth, Storage, and Edge Functions with RLS-first patterns, schema-generated TypeScript types, and local Docker parity for development and testing.',
       },
       {
         icon: Code2,
@@ -204,7 +204,7 @@ export const landingConfig: LandingConfig = {
     },
     {
       title: 'Web and mobile from a single codebase.',
-      body: 'Shared auth logic, shared billing hooks, shared business rules — your React web app and React Native mobile app stay in sync without duplicating code. 40–60% of the codebase is shared across platforms.',
+      body: 'Shared auth logic. Shared billing hooks. Shared business rules. Your React web app and React Native mobile app stay in sync without duplicated work. About 40–60% of the codebase is shared across platforms.',
       ctaLabel: 'Learn about mobile',
       ctaHref: '#features',
       mediaSrc: 'https://placehold.co/560x315?text=Mobile+App',
@@ -213,7 +213,7 @@ export const landingConfig: LandingConfig = {
     },
     {
       title: 'Environments that match your shipping workflow.',
-      body: 'Three Supabase databases — local Docker, shared PR testing, and staging — mirror your branching model so migrations are validated before they touch production. PR previews deploy to path-based S3/CloudFront URLs automatically. EAS Update channels give mobile the same preview → staging → production story.',
+      body: 'Three Supabase databases—local Docker, shared PR testing, and staging—mirror your branching model. Migrations are validated before they touch production. PR previews deploy automatically to path-based S3/CloudFront URLs. EAS Update channels give mobile the same preview → staging → production flow.',
       ctaLabel: 'Read the architecture docs',
       ctaHref: 'https://github.com/Artificer-Innovations/BeakerStack/blob/main/ARCHITECTURE.md',
       mediaSrc: 'https://placehold.co/560x315?text=Environments',
@@ -222,7 +222,7 @@ export const landingConfig: LandingConfig = {
     },
     {
       title: 'Set up in minutes. Documented for the long haul.',
-      body: 'A single `npm run setup` command provisions your local environment. QUICKSTART, ARCHITECTURE, and per-topic guides (OAuth, Stripe, testing) cover every decision. Environment variables and secrets follow a consistent layout that maps directly to GitHub Actions — no hunting for where a key goes.',
+      body: 'Run `npm run setup` to provision your local environment in one step. QUICKSTART, ARCHITECTURE, and focused guides (OAuth, Stripe, testing) explain key decisions. Env vars and secrets use a consistent layout that maps directly to GitHub Actions.',
       ctaLabel: 'Read the quickstart',
       ctaHref: 'https://github.com/Artificer-Innovations/BeakerStack/blob/main/QUICKSTART.md',
       mediaSrc: 'https://placehold.co/560x315?text=Setup',
--- a/packages/shared-tests/jest.web.cjs
+++ b/packages/shared-tests/jest.web.cjs
@@ -22,7 +22,7 @@ module.exports = {
     ]
   },
   transformIgnorePatterns: [
-    "node_modules/(?!^$)" // default; adjust to transpile specific libs if needed
+    "/node_modules/" // default; adjust to transpile specific libs if needed
   ],
   moduleFileExtensions: ["ts", "tsx", "js", "jsx", "json"],
   collectCoverageFrom: [
--- a/apps/web/src/pages/__tests__/HomePage.test.tsx
+++ b/apps/web/src/pages/__tests__/HomePage.test.tsx
@@ -22,7 +22,7 @@ vi.mock('@/lib/supabase', () => ({
     }),
     channel: vi.fn().mockReturnValue({
       on: vi.fn().mockReturnThis(),
-      subscribe: vi.fn(cb => { cb('SUBSCRIBED'); return {}; }),
+      subscribe: vi.fn(cb => { cb('SUBSCRIBED'); return { unsubscribe: vi.fn().mockResolvedValue(undefined) }; }),
     }),
     removeChannel: vi.fn().mockResolvedValue({ status: 'ok', error: null }),
   },
@@ -84,7 +84,7 @@ describe('HomePage', () => {
       }),
       channel: vi.fn().mockReturnValue({
         on: vi.fn().mockReturnThis(),
-        subscribe: vi.fn(cb => { cb('SUBSCRIBED'); return {}; }),
+        subscribe: vi.fn(cb => { cb('SUBSCRIBED'); return { unsubscribe: vi.fn().mockResolvedValue(undefined) }; }),
       }),
       removeChannel: vi.fn().mockResolvedValue({ status: 'ok', error: null }),
     } as unknown as SupabaseClient;
@@ -113,8 +113,13 @@ describe('HomePage', () => {
           </AuthProvider>
         </MemoryRouter>
       );
-      await new Promise(resolve => setTimeout(resolve, 0));
     });
+    await waitFor(() =>
+      expect(
+        screen.queryByText('Build the full stack. Not the scaffolding.') ??
+          screen.queryByText('Dashboard')
+      ).toBeTruthy()
+    );
     return result!;
   };
 


rahul-artificerinnovations (commented):
Reviewed the diff. All three fixes are correct — holding formal approval for CI to go green on `tests`.

**`landing.ts` copy:** Clean improvements — active voice, shorter sentences, em-dashes. No concerns.

**Subscribe mock:** Correct fix. `useProfile.ts` calling `unsubscribe().catch(...)` means the mock's `subscribe()` return needs an `unsubscribe` method that returns a Promise. The old `{}` return left `unsubscribe` as `undefined`, causing a silent `TypeError: unsubscribe is not a function` during effect cleanup. `vi.fn().mockResolvedValue(undefined)` matches the real Supabase shape exactly.

**`waitFor()` flush:** Better pattern. `setTimeout(resolve, 0)` was a timing hack; `waitFor()` retries until the DOM settles and makes the test intent explicit. The dual-query `queryByText('headline') ?? queryByText('Dashboard')` is correct — `null ?? null` would fail with `expected null to be truthy`, triggering a retry. Minor: if this assertion ever needs a clearer failure message, a `getByText` paired with a `.toBeInTheDocument()` inside a try/catch would be more descriptive, but not a blocker.

**`transformIgnorePatterns`:** The change to `/node_modules/` is the right idiomatic Jest default. One small clarification on the PR description: `node_modules/(?!^$)` was actually still excluding all node_modules paths from transpilation (the lookahead `(?!^$)` always succeeds mid-string since `^` only anchors at position 0), so the old behavior was functionally equivalent — just unreadable and likely to mislead anyone trying to add package allowlisting. The fix is the correct call regardless.

Will approve once `tests` check clears.

rahul-artificerinnovations (commented):
Reviewed the diff. All three fixes are correct — holding formal approval for CI to go green on `tests`.

**`landing.ts` copy:** Clean improvements — active voice, shorter sentences, em-dashes. No concerns.

**Subscribe mock:** Correct fix. `useProfile.ts` calling `unsubscribe().catch(...)` means the mock's `subscribe()` return needs an `unsubscribe` method that returns a Promise. The old `{}` return left `unsubscribe` as `undefined`, causing a silent `TypeError: unsubscribe is not a function` during effect cleanup. `vi.fn().mockResolvedValue(undefined)` matches the real Supabase shape exactly.

**`waitFor()` flush:** Better pattern. `setTimeout(resolve, 0)` was a timing hack; `waitFor()` retries until the DOM settles and makes the test intent explicit. The dual-query `queryByText('headline') ?? queryByText('Dashboard')` is correct — `null ?? null` would fail with `expected null to be truthy`, triggering a retry. Minor: if this assertion ever needs a clearer failure message, a `getByText` paired with a `.toBeInTheDocument()` inside a try/catch would be more descriptive, but not a blocker.

**`transformIgnorePatterns`:** The change to `/node_modules/` is the right idiomatic Jest default. One small clarification on the PR description: `node_modules/(?!^$)` was actually still excluding all node_modules paths from transpilation (the lookahead `(?!^$)` always succeeds mid-string since `^` only anchors at position 0), so the old behavior was functionally equivalent — just unreadable and likely to mislead anyone trying to add package allowlisting. The fix is the correct call regardless.

Will approve once `tests` check clears.

mei-artificerinnovations (approved):
CI green. All three fixes are correct.

**`jest.web.cjs` — `transformIgnorePatterns` regex** — `node_modules/(?!^$)` is a broken negative lookahead: module names after `node_modules/` are never empty, so the lookahead always passes and the pattern matches nothing — effectively transpiling all of node_modules on every test run. `/node_modules/` is the correct standard pattern. This was likely causing noticeably slow shared-test runs. ✓

**`HomePage.test.tsx` — subscribe mock** — empty `{}` return is a latent crash if anything in cleanup calls `channel.unsubscribe().catch(...)`. Adding `{ unsubscribe: vi.fn().mockResolvedValue(undefined) }` is the right fix, and applying it in both mock locations is correct. ✓

**`HomePage.test.tsx` — `waitFor` flush** — `waitFor` with a real DOM assertion is strictly better than `setTimeout(resolve, 0)`: it retries until the DOM settles rather than relying on an arbitrary delay, and it fails with a useful message if the expected content never appears. ✓

**Copy edits** — shorter, punchier. No issues.

Approved.

packages/shared-tests/jest.web.cjs
```diff
@@ -22,7 +22,7 @@ module.exports = {
     ]
   },
```
sarah-artificerinnovations, line 23: The new `/node_modules/` pattern is right and cleaner, but the stated reason is slightly off. Verified: `node_modules/(?!^$)` *does* match real file paths — `^$` inside a lookahead in non-multiline mode tests whether the *entire input* is empty, which is never true for a real file path. So the original was behaving correctly (not transpiling node_modules), just via a confusing regex. The bug was unreadable code, not broken behavior. Worth amending the commit message or PR description so the history reflects the actual change accurately.

sarah-artificerinnovations (approved):
## PR #64 review — Approving ✅

CI green. Two of the three fixes are straightforward; one has a small inaccuracy in the stated reasoning (not blocking — the change itself is still correct).

**Subscribe mock** ✅ — `{ unsubscribe: vi.fn().mockResolvedValue(undefined) }` is the right fix. `useProfile.ts` chains `.catch()` on the return of `unsubscribe()`, which means the return must be a Promise. The empty `{}` was silently swallowing the error in prior tests that didn't hit the cleanup path — good catch.

**`waitFor` refactor** ✅ — Replacing the arbitrary `setTimeout(resolve, 0)` flush with a `waitFor` outside `act` is the right move. The `??` operator works correctly here: `queryByText` returns `null` (nullish) when not found, so the coalescing behaves identically to `||` for DOM queries. Handles both unauthenticated (landing headline) and authenticated (Dashboard) render paths cleanly.

**`transformIgnorePatterns` — fix is right, stated reason needs a small correction (see inline)**

**nit: "migrations at repo root" removed from Supabase grid item**

The Supabase backend grid body originally read: `"…migrations at repo root, generated TypeScript types from the schema, and local Supabase via Docker for dev/test parity."` — this specific detail was called out in the issue #54 plan review as a required part of the Supabase backend story. The condensed version drops it. Not blocking (it's copy), but worth being intentional about — if it's gone for length, that's fine; just noting it was there for a reason.

apps/web/src/config/landing.ts
```diff
@@ -204,7 +204,7 @@ export const landingConfig: LandingConfig = {
     },
     {
       title: 'Web and mobile from a single codebase.',
-      body: 'Shared auth logic, shared billing hooks, shared business rules — your React web app and React Native mobile app stay in sync without duplicating code. 40–60% of the codebase is shared across platforms.',
+      body: 'Shared auth logic. Shared billing hooks. Shared business rules. Your React web app and React Native mobile app stay in sync without duplicated work. About 40–60% of the codebase is shared across platforms.',
```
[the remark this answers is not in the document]
ZappoMan: Agreed @diego-artificerinnovations - I like copilot's recommendation.

packages/shared-tests/jest.web.cjs
ZappoMan: @diego-artificerinnovations - I agree with @sarah-artificerinnovations on this, let's make sure the PR description is correctly explaining the change. Our future selves will thank us.

ZappoMan: @diego-artificerinnovations - also I agree with @sarah-artificerinnovations that we don't want to drop meaningful information about Supabase.

rahul-artificerinnovations (approved):
CI green. All three fixes are correct — approving.

(See earlier comment for full review notes on the subscribe mock, `waitFor()` flush, and `transformIgnorePatterns` change.)

apps/web/src/config/landing.ts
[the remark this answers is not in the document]
diego-artificerinnovations: Good catch — "without duplicated work" reads awkwardly. Changed to "without duplicating work" in the latest commit.

packages/shared-tests/jest.web.cjs
diego-artificerinnovations: You are correct, and I overstated this in the PR description. The original `node_modules/(?!^$)` was functionally equivalent to `/node_modules/` — it was excluding node_modules from transpilation, just via an opaque negative-lookahead. The change is a readability fix, not a behavioral one. Updated the PR description to reflect this accurately.

apps/web/src/config/landing.ts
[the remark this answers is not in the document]
diego-artificerinnovations: Fixed — changed to "without duplicating work".

packages/shared-tests/jest.web.cjs
diego-artificerinnovations: Agreed, and the PR description is updated to say the original was functionally correct but non-idiomatic. The change is about readability, not fixing broken behavior.

PR #69, opened 2026-05-11
feat(web): post-signup upgrade funnel — plan intent, billing redirect, OAuth/email storage

### LINKED ISSUE #68: Post-pricing signup: preserve plan intent, contextual signup UI, and route paid signups to billing checkout
## Business case

Landing pricing already sends users to signup with intent in the URL (`/signup?plan=…&cadence=…` from [`PricingSection`](apps/web/src/components/landing/sections/PricingSection.tsx)), but after account creation they always land on the **dashboard** and the signup screen shows **no plan context**. That creates two conversion problems:

1. **Broken upgrade path** — High-intent clicks (“Get started with Pro”) do not continue into checkout; momentum stops at a generic home surface instead of `/billing/plans` where Stripe checkout already exists.
2. **Friction and doubt at signup** — Users cannot confirm which tier or price cadence they chose while entering credentials. Research and common SaaS practice support keeping a **lightweight plan summary** (name, price, cadence, a few value bullets) and **intentional CTA copy** (e.g. “Continue with Pro”) to reduce abandonment and post-signup regret.

**Expected outcome:** Higher completion from pricing → signup → paid checkout, with clearer reassurance and no open redirects or callback URL allow-list surprises.

---

## Current behavior (root cause)

- [`apps/web/src/pages/SignupPage.tsx`](apps/web/src/pages/SignupPage.tsx) always navigates to `/dashboard` after email signup; it does not read `plan` / `cadence`.
- [`apps/web/src/pages/AuthCallbackPage.tsx`](apps/web/src/pages/AuthCallbackPage.tsx) always navigates to `/dashboard` after Google OAuth.
- OAuth `redirectTo` is fixed in [`packages/shared/src/hooks/useAuth.ts`](packages/shared/src/hooks/useAuth.ts); arbitrary `?next=` on the callback URL is not a safe default because Supabase redirect allow-lists are usually exact URLs.

**Free tier** (`beakerstack_free`): post-auth destination should remain **dashboard** (no paid funnel).

---

## Target user flow

1. User selects cadence on marketing pricing and clicks a **paid** plan CTA.
2. **Signup** shows a **plan summary** (name, price aligned with marketing, 2–3 bullets from billing metadata) and primary button **“Continue with {Plan}”** when `?plan=` is a valid paid tier.
3. After **email/password** or **Google** auth, user lands on **`/billing/plans`** with the same `plan` + `cadence`, a **welcome** query flag for UI, scroll/highlight on the chosen plan, and a short banner; they confirm and use existing **Stripe checkout** (no auto-redirect to Stripe on load).

```mermaid
flowchart LR
  Landing[Pricing CTA] --> Signup["/signup?plan=&cadence="]
  Signup --> EmailFlow[Email signup]
  Signup --> OAuthFlow[Google OAuth]
  EmailFlow --> Plans["/billing/plans?plan=&cadence=&welcome=1"]
  OAuthFlow --> Callback["/auth/callback"]
  Callback --> Plans
  Plans --> Stripe[Stripe Checkout]
```

---

### 1. Post-auth path resolver (web)

- Add a small helper (e.g. `apps/web/src/auth/postSignupRedirect.ts`).
- **Rules:** unknown / missing `plan` → `/dashboard`; free plan → `/dashboard`; valid **paid** public plan → `/billing/plans?plan=…&cadence=…&welcome=1` (cadence rules aligned with [`PricingSection`](apps/web/src/components/landing/sections/PricingSection.tsx)).
- **Security:** only allow internal paths (e.g. must start with `/`); never pass through external URLs.

### 2. Signup plan summary (MVP)

- When `?plan=` resolves to a **paid** catalog plan: show summary **card** (desktop: two-column with form; mobile: summary above form).
- **Data:** `BillingProvider` + `usePlanCatalog` + [`beakerstackBillingConfig`](apps/web/src/billing/beakerstackBillingConfig.ts); reuse same price formatting as [`PricingSection.tsx`](apps/web/src/components/landing/sections/PricingSection.tsx) ([`getCadenceFromSearch`](apps/web/src/components/billing/CadenceToggle.web.tsx), [`annualListCentsFromSync`](apps/web/src/billing/billingSyncDisplay.ts)).
- **Bullets:** derive 2–3 lines from config (`features`, `usageLimits`, boolean `planFeatureRows`) — single source of truth, not duplicate marketing copy.
- **Copy:** e.g. “Create your account to continue with **Pro**”; submit label **“Continue with {display_name}”** for paid intent; loading/invalid plan → generic signup.

### 3. Signup / login wiring

- [`SignupPage.tsx`](apps/web/src/pages/SignupPage.tsx): `useSearchParams`; navigate using resolver after successful signup; preserve query on link to **Sign in**.
- [`LoginPage.tsx`](apps/web/src/pages/LoginPage.tsx): same resolver after sign-in; optional compact reuse of plan summary when `?plan=` is paid.

### 4. Google OAuth continuation

- Before `signInWithGoogle`, when post-auth target is not dashboard, set a **namespaced** `sessionStorage` key with **pathname + search** only (respect PR-preview base path the same way OAuth already does).
- [`AuthCallbackPage.tsx`](apps/web/src/pages/AuthCallbackPage.tsx): after session is established, read + clear storage and `navigate` there, else `/dashboard`.

### 5. Billing plans “welcome” UX

- [`BillingPlansPage.tsx`](apps/web/src/pages/billing/BillingPlansPage.tsx): read `plan`, `welcome`, `cadence`; banner + scroll to plan card; stable `id` on [`PlanCard`](apps/web/src/components/billing/PlanCard.web.tsx) wrapper; optionally strip `welcome` from URL after first paint.

### 6. Tests

- Signup: paid `plan` → `/billing/plans` with expected query (mock auth).
- Signup: summary + “Continue with Pro” when `?plan=beakerstack_pro`; generic when no plan.
- Auth callback: sessionStorage consumed and navigation correct.
- Billing plans: `welcome=1` + `plan` shows banner (scroll optional / brittle in JSDOM).

### 7. Follow-ups (out of scope for this issue unless scoped)

- [`apps/mobile/src/screens/SignupScreen.tsx`](apps/mobile/src/screens/SignupScreen.tsx) parity if mobile gets pricing deep links.
- **Email confirmation** enabled in production: no navigate without session; optional phase-2 stash of intent for first session after confirm.

---

## Acceptance criteria

- [ ] From marketing pricing, paid CTA → signup URL still includes `plan` (and `cadence` when annual).
- [ ] Signup with valid paid `plan` shows plan summary (name, price/cadence consistent with pricing) and primary CTA **“Continue with {display_name}”**.
- [ ] After email signup with paid intent, user lands on `/billing/plans` with matching query and welcome UX; free plan still goes to dashboard.
- [ ] After Google signup with paid intent, user lands on same billing plans URL (via sessionStorage + callback).
- [ ] Login with same query params behaves consistently when implemented.
- [ ] No open redirect; invalid `plan` falls back to generic signup + dashboard behavior.
- [ ] Tests cover redirect, signup plan UI, and callback storage behavior at minimum.

---

## Work checklist (maps to implementation)

- [ ] Path resolver (`postSignupRedirect` or equivalent)
- [ ] Signup page: params, navigation, plan summary UI, CTA copy, Sign-in link preserves query
- [ ] Login page: params, navigation, optional compact summary, Google sessionStorage
- [ ] Auth callback: read/clear `sessionStorage`, navigate
- [ ] Billing plans: welcome banner, scroll target, card `id`
- [ ] Tests (web)

ZappoMan:
Closes #68 

## Summary

Implements the **post-pricing signup funnel**: preserve plan intent from landing pricing through signup/login, route paid-intent users to **`/billing/plans`** after authentication (with welcome UX), and show **plan context on signup** using the same Supabase-backed catalog as the home page.

## What changed

- **`resolvePostAuthDestination`** + **`validateInternalPostAuthPath`** — synchronous resolver against `beakerstackBillingConfig`; unknown/free → `/dashboard`; paid public → `/billing/plans?plan=…&welcome=1` (+ cadence); rejects protocol-relative and external URLs.
- **`beakerstack:post_auth_redirect`** — JSON payload `{ path, ts }` with **30-minute TTL** in **`sessionStorage`** (Google OAuth) and **`localStorage`** (email confirmation when signup returns no session).
- **`AuthCallbackPage`** — read both stores, **unconditionally `removeItem` both**, parse/TTL/validate, then `navigate(..., { replace: true })`.
- **`SignupPage` / `LoginPage`** — `BillingProvider` + plan summary; preserve query between signup ↔ login; clear stale `localStorage` on successful email auth; Google stashes intent before OAuth.
- **`BillingPlansPage`** — welcome banner, scroll to **`plan-card-{plan.id}`**, strip **`welcome`** from URL with **`replace: true`**.
- Tests for redirect helper, callback coverage, billing welcome, plan bullets.

## Type of change

- [x] Feature
- [ ] Bug fix
- [ ] Documentation
- [ ] Refactoring
- [ ] Other

## Merge strategy (targets `develop`)

Use **Squash and merge** when merging into `develop`.

## Checklist

- [x] Code follows the project's style guidelines
- [x] Self-review completed
- [ ] Comments added for complex code (only where needed)
- [ ] Documentation updated (if needed)
- [x] No new warnings generated (address any CI findings)
- [x] Tests added/updated
- [x] All tests pass locally (`apps/web` vitest, `tsc --noEmit`)

## Testing

- `cd apps/web && npm test`
- `cd apps/web && npx tsc --noEmit`

## Related

Closes / implements **[#68 — Post-pricing signup: preserve plan intent, contextual signup UI, and route paid signups to billing checkout](https://github.com/Artificer-Innovations/BeakerStack/issues/68)** (if that matches scope—adjust if not).

### DIFF
--- a/apps/web/src/auth/__tests__/planSignupBullets.test.ts
+++ b/apps/web/src/auth/__tests__/planSignupBullets.test.ts
@@ -2,18 +2,48 @@ import { describe, expect, it } from 'vitest';
 import { planSignupBullets } from '../planSignupBullets';
 
 describe('planSignupBullets', () => {
-  it('returns bullets for Pro', () => {
+  it('returns non-empty strings for known plans', () => {
     const b = planSignupBullets('beakerstack_pro');
     expect(b.length).toBeGreaterThan(0);
-    expect(b.some(x => /Feature A/i.test(x))).toBe(true);
+    expect(b.length).toBeLessThanOrEqual(3);
+    expect(b.every(x => typeof x === 'string' && x.length > 0)).toBe(true);
   });
 
-  it('includes trial copy for Max', () => {
+  it('includes trial copy when trial days > 0', () => {
     const b = planSignupBullets('beakerstack_max');
     expect(b.some(x => /trial/i.test(x))).toBe(true);
   });
 
   it('returns empty for unknown plan', () => {
     expect(planSignupBullets('unknown')).toEqual([]);
+  });
+
+  it('lists limited numeric features for Free plan', () => {
+    const b = planSignupBullets('beakerstack_free');
+    expect(b.some(x => /Up to 2 collections/i.test(x))).toBe(true);
+    expect(b.some(x => /Up to 3 items per collection/i.test(x))).toBe(true);
+    expect(b.some(x => /AI summarize/i.test(x))).toBe(true);
+    expect(b.length).toBeLessThanOrEqual(3);
+  });
+
+  it('includes unlimited collection copy when capped at -1', () => {
+    const b = planSignupBullets('beakerstack_pro');
+    expect(b.some(x => /Unlimited collections/i.test(x))).toBe(true);
+    expect(b.some(x => /Feature A/i.test(x))).toBe(true);
+  });
+
+  it('caps at three lines on Max (trial + booleans fill quota before meter)', () => {
+    const b = planSignupBullets('beakerstack_max');
+    expect(b.length).toBe(3);
+    expect(b.some(x => /trial/i.test(x))).toBe(true);
+    expect(b.some(x => /Feature A/i.test(x))).toBe(true);
+    expect(b.some(x => /Feature B/i.test(x))).toBe(true);
+  });
+
+  it('adds meter copy when feature rows leave room (Free)', () => {
+    const b = planSignupBullets('beakerstack_free');
+    expect(
+      b.some(x => /\d+\s+AI summarize\s+\/\s+billing period/i.test(x))
+    ).toBe(true);
   });
 });
--- a/apps/web/src/auth/__tests__/planSignupBullets.test.ts
+++ b/apps/web/src/auth/__tests__/planSignupBullets.test.ts
@@ -1,0 +1,19 @@
+import { describe, expect, it } from 'vitest';
+import { planSignupBullets } from '../planSignupBullets';
+
+describe('planSignupBullets', () => {
+  it('returns bullets for Pro', () => {
+    const b = planSignupBullets('beakerstack_pro');
+    expect(b.length).toBeGreaterThan(0);
+    expect(b.some(x => /Feature A/i.test(x))).toBe(true);
+  });
+
+  it('includes trial copy for Max', () => {
+    const b = planSignupBullets('beakerstack_max');
+    expect(b.some(x => /trial/i.test(x))).toBe(true);
+  });
+
+  it('returns empty for unknown plan', () => {
+    expect(planSignupBullets('unknown')).toEqual([]);
+  });
+});
--- a/apps/web/src/auth/planSignupBullets.ts
+++ b/apps/web/src/auth/planSignupBullets.ts
@@ -12,16 +12,17 @@ export function planSignupBullets(planId: string): string[] {
     );
   }
 
-  const feats = cfg.features as Record<string, number | boolean>;
+  const feats = cfg.features as Record<string, number | boolean | undefined>;
   for (const row of beakerstackBillingConfig.planFeatureRows) {
     if (out.length >= 3) break;
     const val = feats[row.featureKey];
     if (row.kind === 'boolean') {
       if (val === true) out.push(row.label);
     } else {
-      const n = val as number;
-      if (n === -1) out.push(row.unlimitedLabel);
-      else out.push(row.limitedLabelTemplate.replace('{count}', String(n)));
+      if (typeof val !== 'number') continue;
+      if (val === -1) out.push(row.unlimitedLabel);
+      else
+        out.push(row.limitedLabelTemplate.replace('{count}', String(val)));
     }
   }
 
@@ -39,5 +40,5 @@ export function planSignupBullets(planId: string): string[] {
     }
   }
 
-  return out.slice(0, 3);
+  return out;
 }
--- a/apps/web/src/auth/planSignupBullets.ts
+++ b/apps/web/src/auth/planSignupBullets.ts
@@ -1,0 +1,43 @@
+import { beakerstackBillingConfig } from '../billing/beakerstackBillingConfig';
+
+/** Up to 3 human-readable lines for signup plan context (from template billing config). */
+export function planSignupBullets(planId: string): string[] {
+  const cfg = beakerstackBillingConfig.plans.find(p => p.id === planId);
+  if (!cfg) return [];
+
+  const out: string[] = [];
+  if (cfg.trialPeriodDays > 0) {
+    out.push(
+      `Includes a ${cfg.trialPeriodDays}-day trial, then billed at this rate.`
+    );
+  }
+
+  const feats = cfg.features as Record<string, number | boolean>;
+  for (const row of beakerstackBillingConfig.planFeatureRows) {
+    if (out.length >= 3) break;
+    const val = feats[row.featureKey];
+    if (row.kind === 'boolean') {
+      if (val === true) out.push(row.label);
+    } else {
+      const n = val as number;
+      if (n === -1) out.push(row.unlimitedLabel);
+      else out.push(row.limitedLabelTemplate.replace('{count}', String(n)));
+    }
+  }
+
+  const ai = cfg.usageLimits?.ai_summarize;
+  if (out.length < 3 && typeof ai === 'number') {
+    const label = beakerstackBillingConfig.usageMeterCopy.ai_summarize?.label;
+    if (ai === -1) {
+      out.push(label ? `Unlimited ${label}` : 'Unlimited AI summarize');
+    } else {
+      out.push(
+        label
+          ? `${ai} ${label} / billing period`
+          : `${ai} AI summarize requests / billing period`
+      );
+    }
+  }
+
+  return out.slice(0, 3);
+}
--- a/apps/web/src/auth/planSignupBullets.ts
+++ b/apps/web/src/auth/planSignupBullets.ts
@@ -21,8 +21,7 @@ export function planSignupBullets(planId: string): string[] {
     } else {
       if (typeof val !== 'number') continue;
       if (val === -1) out.push(row.unlimitedLabel);
-      else
-        out.push(row.limitedLabelTemplate.replace('{count}', String(val)));
+      else out.push(row.limitedLabelTemplate.replace('{count}', String(val)));
     }
   }
 
--- a/apps/web/src/auth/postAuthRedirect.ts
+++ b/apps/web/src/auth/postAuthRedirect.ts
@@ -48,7 +48,7 @@ export function resolvePostAuthDestination(
   if (!planId || !PUBLIC_PLAN_IDS.has(planId)) return '/dashboard';
 
   const cfg = configPlanById(planId);
-  if (!cfg || !cfg.isPublic || cfg.priceCents === 0) return '/dashboard';
+  if (!cfg || cfg.priceCents === 0) return '/dashboard';
 
   const cadence =
     searchParams.get('cadence') === 'annual' ? 'annual' : 'monthly';
@@ -65,6 +65,7 @@ export function serializePostAuthRedirectPayload(path: string): string {
 
 /**
  * Parse stored JSON `{ path, ts }`, enforce TTL, validate path. Returns null if invalid/expired.
+ * Non-JSON or malformed payloads return null (TTL always enforced; no legacy raw-path fallback).
  */
 export function parseStoredPostAuthRedirect(
   raw: string | null,
@@ -80,8 +81,34 @@ export function parseStoredPostAuthRedirect(
     }
     return null;
   } catch {
-    return validateInternalPostAuthPath(raw, origin);
+    return null;
   }
+}
+
+/** Remove intent from both storages (after successful email auth or when clearing stale OAuth stash). */
+export function clearPostAuthRedirectKeys(): void {
+  if (typeof sessionStorage !== 'undefined') {
+    sessionStorage.removeItem(POST_AUTH_REDIRECT_KEY);
+  }
+  if (typeof localStorage !== 'undefined') {
+    localStorage.removeItem(POST_AUTH_REDIRECT_KEY);
+  }
+}
+
+/** Read intent once, clear both stores unconditionally, parse + TTL + validate. */
+export function readAndClearPostAuthRedirect(): string | null {
+  const rawS =
+    typeof sessionStorage !== 'undefined'
+      ? sessionStorage.getItem(POST_AUTH_REDIRECT_KEY)
+      : null;
+  const rawL =
+    typeof localStorage !== 'undefined'
+      ? localStorage.getItem(POST_AUTH_REDIRECT_KEY)
+      : null;
+  clearPostAuthRedirectKeys();
+  return (
+    parseStoredPostAuthRedirect(rawS) ?? parseStoredPostAuthRedirect(rawL)
+  );
 }
 
 export function hasPaidPlanIntent(searchParams: URLSearchParams): boolean {
--- a/apps/web/src/auth/postAuthRedirect.ts
+++ b/apps/web/src/auth/postAuthRedirect.ts
@@ -1,0 +1,89 @@
+import { beakerstackBillingConfig } from '../billing/beakerstackBillingConfig';
+
+/** Namespaced key — sessionStorage (OAuth) + localStorage (email confirmation). */
+export const POST_AUTH_REDIRECT_KEY = 'beakerstack:post_auth_redirect';
+
+export const POST_AUTH_REDIRECT_TTL_MS = 30 * 60 * 1000;
+
+const PUBLIC_PLAN_IDS: Set<string> = new Set(
+  beakerstackBillingConfig.plans
+    .filter(p => p.isPublic)
+    .map(p => p.id)
+);
+
+function configPlanById(planId: string) {
+  return beakerstackBillingConfig.plans.find(p => p.id === planId);
+}
+
+/**
+ * Returns pathname + search for an in-app redirect, or null if unsafe / external.
+ * Rejects protocol-relative URLs (e.g. //evil.com/...) and non-same-origin paths.
+ */
+export function validateInternalPostAuthPath(
+  path: string,
+  origin: string = typeof window !== 'undefined'
+    ? window.location.origin
+    : 'http://localhost'
+): string | null {
+  if (!path || typeof path !== 'string') return null;
+  const trimmed = path.trim();
+  if (!trimmed.startsWith('/') || trimmed.startsWith('//')) return null;
+  try {
+    const u = new URL(trimmed, origin);
+    if (u.origin !== new URL(origin).origin) return null;
+    return u.pathname + u.search;
+  } catch {
+    return null;
+  }
+}
+
+/**
+ * Synchronous post-auth path from signup/login URL search params.
+ * Unknown or free plan → /dashboard; paid public plan → /billing/plans?…
+ */
+export function resolvePostAuthDestination(
+  searchParams: URLSearchParams
+): string {
+  const planId = searchParams.get('plan');
+  if (!planId || !PUBLIC_PLAN_IDS.has(planId)) return '/dashboard';
+
+  const cfg = configPlanById(planId);
+  if (!cfg || !cfg.isPublic || cfg.priceCents === 0) return '/dashboard';
+
+  const cadence =
+    searchParams.get('cadence') === 'annual' ? 'annual' : 'monthly';
+  const q = new URLSearchParams();
+  q.set('plan', planId);
+  q.set('welcome', '1');
+  if (cadence === 'annual') q.set('cadence', 'annual');
+  return `/billing/plans?${q.toString()}`;
+}
+
+export function serializePostAuthRedirectPayload(path: string): string {
+  return JSON.stringify({ path, ts: Date.now() });
+}
+
+/**
+ * Parse stored JSON `{ path, ts }`, enforce TTL, validate path. Returns null if invalid/expired.
+ */
+export function parseStoredPostAuthRedirect(
+  raw: string | null,
+  now: number = Date.now(),
+  origin?: string
+): string | null {
+  if (!raw) return null;
+  try {
+    const o = JSON.parse(raw) as { path?: string; ts?: number };
+    if (typeof o.path === 'string' && typeof o.ts === 'number') {
+      if (now - o.ts > POST_AUTH_REDIRECT_TTL_MS) return null;
+      return validateInternalPostAuthPath(o.path, origin);
+    }
+    return null;
+  } catch {
+    return validateInternalPostAuthPath(raw, origin);
+  }
+}
+
+export function hasPaidPlanIntent(searchParams: URLSearchParams): boolean {
+  return resolvePostAuthDestination(searchParams) !== '/dashboard';
+}
--- a/apps/web/src/auth/postAuthRedirect.ts
+++ b/apps/web/src/auth/postAuthRedirect.ts
@@ -6,9 +6,7 @@ export const POST_AUTH_REDIRECT_KEY = 'beakerstack:post_auth_redirect';
 export const POST_AUTH_REDIRECT_TTL_MS = 30 * 60 * 1000;
 
 const PUBLIC_PLAN_IDS: Set<string> = new Set(
-  beakerstackBillingConfig.plans
-    .filter(p => p.isPublic)
-    .map(p => p.id)
+  beakerstackBillingConfig.plans.filter(p => p.isPublic).map(p => p.id)
 );
 
 function configPlanById(planId: string) {
@@ -106,9 +104,7 @@ export function readAndClearPostAuthRedirect(): string | null {
       ? localStorage.getItem(POST_AUTH_REDIRECT_KEY)
       : null;
   clearPostAuthRedirectKeys();
-  return (
-    parseStoredPostAuthRedirect(rawS) ?? parseStoredPostAuthRedirect(rawL)
-  );
+  return parseStoredPostAuthRedirect(rawS) ?? parseStoredPostAuthRedirect(rawL);
 }
 
 export function hasPaidPlanIntent(searchParams: URLSearchParams): boolean {
--- a/apps/web/src/components/auth/SignupPlanSummary.tsx
+++ b/apps/web/src/components/auth/SignupPlanSummary.tsx
@@ -119,8 +119,8 @@ export function PlanIntentSummary({
       </div>
       {bullets.length > 0 ? (
         <ul className='mt-4 list-inside list-disc space-y-1 text-sm text-gray-700'>
-          {bullets.map(line => (
-            <li key={line}>{line}</li>
+          {bullets.map((line, i) => (
+            <li key={`${mode}-bullet-${i}`}>{line}</li>
           ))}
         </ul>
       ) : null}
--- a/apps/web/src/components/auth/SignupPlanSummary.tsx
+++ b/apps/web/src/components/auth/SignupPlanSummary.tsx
@@ -1,0 +1,137 @@
+import { usePlanCatalog } from '@beakerstack/billing';
+import { useMemo } from 'react';
+import { useSearchParams } from 'react-router-dom';
+import { beakerstackBillingConfig } from '../../billing/beakerstackBillingConfig';
+import {
+  annualListCentsFromSync,
+  formatSavingsCalloutFromCopy,
+  planAnnualSavingsCopy,
+} from '../../billing/billingSyncDisplay';
+import { getCadenceFromSearch } from '../billing/CadenceToggle.web';
+import { hasPaidPlanIntent } from '../../auth/postAuthRedirect';
+import { planSignupBullets } from '../../auth/planSignupBullets';
+
+type PlanIntentMode = 'signup' | 'login';
+
+/**
+ * Plan context when `?plan=` is a paid tier (Supabase catalog + config bullets).
+ * Login uses a compact variant; primary CTA wording stays on the page, not here.
+ */
+export function PlanIntentSummary({
+  mode = 'signup',
+}: {
+  mode?: PlanIntentMode;
+}): JSX.Element | null {
+  const [search] = useSearchParams();
+  const planId = search.get('plan');
+  const cadence = getCadenceFromSearch(search);
+  const { plans, loading } = usePlanCatalog<typeof beakerstackBillingConfig>();
+
+  const catalogPlan = useMemo(
+    () => (planId ? plans.find(p => p.id === planId) : undefined),
+    [plans, planId]
+  );
+
+  if (!planId || !hasPaidPlanIntent(search)) return null;
+
+  if (loading) {
+    return (
+      <div className='rounded-xl border border-gray-200 bg-white p-6 shadow-sm'>
+        <p className='text-sm text-gray-500'>Loading plan…</p>
+      </div>
+    );
+  }
+
+  if (!catalogPlan || catalogPlan.price_cents === 0) return null;
+
+  const isAnnual = cadence === 'annual' && catalogPlan.price_cents > 0;
+  const annualCents = isAnnual
+    ? annualListCentsFromSync(catalogPlan.id, catalogPlan.price_cents)
+    : null;
+
+  const fmt = (cents: number) =>
+    new Intl.NumberFormat('en-US', {
+      style: 'currency',
+      currency: 'USD',
+      maximumFractionDigits: 0,
+    }).format(cents / 100);
+
+  const priceHeadline =
+    isAnnual && annualCents != null
+      ? fmt(annualCents)
+      : fmt(catalogPlan.price_cents);
+
+  const priceSubline =
+    cadence === 'annual' ? 'per year, billed annually' : 'per month';
+
+  const savingsCopy = isAnnual
+    ? planAnnualSavingsCopy(catalogPlan.id, catalogPlan.price_cents)
+    : null;
+  const savingsCallout = savingsCopy
+    ? formatSavingsCalloutFromCopy(savingsCopy)
+    : null;
+
+  const bullets = planSignupBullets(catalogPlan.id).slice(
+    0,
+    mode === 'login' ? 2 : 3
+  );
+
+  return (
+    <div
+      className={
+        mode === 'login'
+          ? 'rounded-lg border border-gray-200 bg-white p-4 shadow-sm'
+          : 'rounded-xl border border-indigo-100 bg-indigo-50/80 p-6 shadow-sm ring-1 ring-indigo-100'
+      }
+    >
+      <p
+        className={
+          mode === 'login'
+            ? 'text-xs font-medium uppercase tracking-wide text-gray-500'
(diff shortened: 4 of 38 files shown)


mei-artificerinnovations (approved):
Solid implementation — the plan is faithfully executed. CI is green, the security-sensitive pieces are correct, and the email-confirmation path is properly handled with `localStorage` + TTL. A few things worth noting for follow-up or future maintenance:

---

### nit: `stashOAuthIntent` not cleared on OAuth failure

In both `SignupPage` and `LoginPage`, `stashOAuthIntent()` writes to `sessionStorage` synchronously before calling `signInWithGoogle()`. If the user cancels the Google OAuth popup or if `signInWithGoogle` throws, the stash remains set. The next time any OAuth callback fires from this tab, `AuthCallbackPage` will pick up and consume the stale intent. In practice the 30-minute TTL bounds the damage, but adding `sessionStorage.removeItem(POST_AUTH_REDIRECT_KEY)` in the `catch` block would be cleaner:

```ts
} catch (err) {
  sessionStorage.removeItem(POST_AUTH_REDIRECT_KEY); // clear stale stash on failure
  setError(...);
}
```

---

### nit: shared backing store in `AuthCallbackPage` test

In `AuthCallbackPage.coverage.test.tsx`, both `sessionStorage` and `localStorage` are stubbed with the same `storageMock` / `memStore` object. In production the two stores are independent; in tests they're aliased. A test that sets a value via `sessionStorage.setItem` will also see it via `localStorage.getItem`, which means the test can't accurately verify the fallback path (localStorage only, no sessionStorage). Current tests pass because the read-and-clear logic works either way, but a future regression where, say, only one store is cleared would go undetected. Two separate `memStore` objects would fix this.

---

### nit: bullet `key={line}` in `PlanIntentSummary`

Bullets are rendered with `key={line}` (the text content). If two bullets ever have identical text, React will warn. `key={i}` (index) is fine here since the list is static during a render.

---

### minor: redundant `out.slice(0, 3)` in `planSignupBullets`

The loop already `break`s at `out.length >= 3`, and the trial bullet is added before the loop, so `out` is at most 3 items by the time `return` is reached. The `.slice(0, 3)` is a no-op. Remove or keep as a defensive guard — either is fine, just inconsistent with the `break` logic above it.

---

### confirmed correct

- `validateInternalPostAuthPath`: `//evil.com` rejection, `new URL(path, origin)` origin check, and `pathname + search` return (drops hash). ✓
- `parseStoredPostAuthRedirect` catch fallback is safe — `validateInternalPostAuthPath` is strict enough that an attacker controlling storage can't redirect off-origin. ✓
- `AuthCallbackPage` unconditional read+clear of both stores before navigation. ✓
- `welcomeSnapshot` captures initial URL state so the banner survives the `welcome` strip. ✓
- `requestAnimationFrame` before `scrollIntoView` + `cancelAnimationFrame` cleanup. ✓
- `BillingProvider` added to both `SignupPage` and `LoginPage`. ✓
- `localStorage.removeItem` cleared on successful email login (prevents stale email-confirmation stash). ✓

apps/web/src/pages/AuthCallbackPage.tsx
```diff
@@ -22,39 +26,58 @@ export default function AuthCallbackPage() {
...
-          // If we got a token but no user after loading completes, there might be an issue
-          setError(
-            'Authentication completed but session not established. Please try again.'
-          );
-          setTimeout(() => {
-            navigate('/login', { replace: true });
-          }, 3000);
-        }
-      }, 1000);
+    if (auth.loading) return;
```
sarah-artificerinnovations, line 37: nit: this read + clear + navigate block is repeated verbatim inside the `setTimeout` callback below. Consider extracting to a local `doNavigate()` inside the `useEffect`:

```ts
const doNavigate = () => {
  const rawS = sessionStorage.getItem(POST_AUTH_REDIRECT_KEY);
  const rawL = localStorage.getItem(POST_AUTH_REDIRECT_KEY);
  sessionStorage.removeItem(POST_AUTH_REDIRECT_KEY);
  localStorage.removeItem(POST_AUTH_REDIRECT_KEY);
  const stored = parseStoredPostAuthRedirect(rawS) ?? parseStoredPostAuthRedirect(rawL);
  navigate(stored ?? '/dashboard', { replace: true });
};
```

Then both branches just call `doNavigate()`. Makes the two code paths easier to audit.

sarah-artificerinnovations (approved):
## PR #69 review — Approving ✅

CI green. The implementation correctly addresses all the plan-review feedback. A few notes inline and below.

**What's solid:**
- `resolvePostAuthDestination` is synchronous, pure, operates on static config only — correct
- `validateInternalPostAuthPath` rejects `//`-prefixed protocol-relative paths via `new URL().origin` comparison — correct
- 30-minute TTL on stored payloads is a nice defensive addition
- `navigatedRef` in `AuthCallbackPage` correctly prevents double-navigation on dep array re-runs
- `authRef` avoids the stale closure in the timer callback
- `BillingProvider` wrapping both `SignupPage` and `LoginPage` — correct
- `welcomeSnapshot` frozen via `useState(() => ...)` initializer — banner state captured once on mount so URL strip doesn't collapse the banner mid-view
- `welcome` stripped with `replace: true`, URL is clean, `effect` guard exits early on re-run — no infinite loop

**nit: duplicate storage-read block in `AuthCallbackPage`**

The read + clear + navigate logic appears verbatim in two places: the immediate `if (auth.user)` branch and inside the `setTimeout` callback. A local `doNavigate()` helper inside the `useEffect` would remove the duplication and make both paths easier to audit if this logic ever needs to change.

**nit: minor redundancy in `resolvePostAuthDestination`**

`PUBLIC_PLAN_IDS` is already filtered by `isPublic`, so the second `!cfg.isPublic` check inside the function body is unreachable (anything that passed `PUBLIC_PLAN_IDS.has(planId)` is guaranteed public, and `!cfg` is also impossible at that point since the Set was built from the same array). Only the `cfg.priceCents === 0` check adds new signal. Harmless, just slightly noisy.

**One coverage gap worth a follow-up: `AuthCallbackPage` storage path**

`postAuthRedirect.test.ts` tests the helper thoroughly. But `AuthCallbackPage.coverage.test.tsx` doesn't cover the case where `sessionStorage` (or `localStorage`) contains a valid redirect and the page navigates to it. All the new storage-read code in the callback runs only in those branches — if someone accidentally moves the read inside the wrong conditional later, no test would catch it. Not blocking this PR, but worth a test addition: seed `sessionStorage` with a serialized payload, render the component with `auth.user` set, and assert `navigate` was called with the stored path.

apps/web/src/pages/SignupPage.tsx
```diff
@@ -51,121 +95,177 @@ export default function SignupPage() {
...
-              Create your account
-            </h2>
-          </div>
-
-          {/* OAuth Buttons */}
-          <div className='space-y-3'>
-            <SocialLoginButton onPress={handleGoogleSignup} mode='signup' />
-          </div>
-
-          {/* Divider */}
```
rahul-artificerinnovations, line 70: **Blocking — layout bug when no plan intent.**

The two-column grid is unconditional here, but `SignupPlanSummary` returns `null` when there's no paid plan. That leaves the right column empty and the form constrained to half width on desktop for all plain signups (direct `/signup`, no `?plan=`). `LoginPage` handles this correctly — it conditionally switches between the grid and `w-full space-y-8` based on `showPlanAside`. SignupPage needs the same pattern:

```tsx
const showPlanAside = paidIntent && postAuthPath !== '/dashboard';

return (
  <div className={showPlanAside ? 'grid w-full gap-10 md:grid-cols-2 md:items-start' : 'w-full space-y-8'}>
    <div className={showPlanAside ? 'order-2 md:order-1 space-y-8' : 'space-y-8'}>
      {/* form */}
    </div>
    {showPlanAside ? (
      <div className='order-1 md:order-2'>
        <SignupPlanSummary />
      </div>
    ) : null}
  </div>
);
```

apps/web/src/auth/planSignupBullets.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  const feats = cfg.features as Record<string, number | boolean>;
+  for (const row of beakerstackBillingConfig.planFeatureRows) {
+    if (out.length >= 3) break;
+    const val = feats[row.featureKey];
+    if (row.kind === 'boolean') {
+      if (val === true) out.push(row.label);
+    } else {
+      const n = val as number;
```
rahul-artificerinnovations: **Blocking — undefined numeric feature value produces broken bullet text.**

When `val` is `undefined` (the plan's `features` object doesn't have a key matching `row.featureKey`), the numeric branch falls through to:
```ts
out.push(row.limitedLabelTemplate.replace('{count}', String(undefined)));
// e.g. "Up to undefined items"
```

The boolean branch handles missing keys correctly (`if (val === true)` short-circuits). The numeric branch needs the same guard:
```ts
} else {
  const n = val as number;
  if (typeof n !== 'number') continue; // skip rows not in this plan's config
  if (n === -1) out.push(row.unlimitedLabel);
  else out.push(row.limitedLabelTemplate.replace('{count}', String(n)));
}
```

apps/web/src/pages/AuthCallbackPage.tsx
```diff
@@ -22,39 +26,58 @@ export default function AuthCallbackPage() {
...
+      const t = setTimeout(() => {
+        navigatedRef.current = true;
         navigate('/login', { replace: true });
       }, 3000);
-      return;
+      return () => clearTimeout(t);
     }
 
-    // If we have an access token in the hash, Supabase client will pick it up automatically
-    // Wait for auth state to update, then redirect
```
rahul-artificerinnovations, line 33: Non-blocking, but the 6-line storage read+clear block is duplicated in two branches — once in the `auth.user` fast path and again inside the `setTimeout` callback. Since they must stay in sync (same key, same parse, same fallback), extract it:

```ts
function readAndClearPostAuthRedirect(): string | null {
  const rawS = sessionStorage.getItem(POST_AUTH_REDIRECT_KEY);
  const rawL = localStorage.getItem(POST_AUTH_REDIRECT_KEY);
  sessionStorage.removeItem(POST_AUTH_REDIRECT_KEY);
  localStorage.removeItem(POST_AUTH_REDIRECT_KEY);
  return parseStoredPostAuthRedirect(rawS) ?? parseStoredPostAuthRedirect(rawL);
}
```

This is a maintenance risk as written — a future change to the key or parse logic has to be applied in two places.

apps/web/src/pages/__tests__/AuthCallbackPage.coverage.test.tsx
```diff
@@ -36,10 +36,27 @@ vi.mock('@beakerstack/shared/contexts/AuthContext', () => ({
...
+  const storageMock = {
+    getItem: (k: string) => (k in memStore ? memStore[k] : null),
+    setItem: (k: string, v: string) => {
+      memStore[k] = v;
+    },
+    removeItem: (k: string) => {
+      delete memStore[k];
+    },
+    clear: () => {
+      for (const k of Object.keys(memStore)) delete memStore[k];
```
rahul-artificerinnovations: Non-blocking — `sessionStorage` and `localStorage` are both stubbed with the **same** `storageMock` object backed by `memStore`. That means writing to `sessionStorage` and reading from `localStorage` (or vice versa) returns the same value, which masks the distinction the production code relies on (sessionStorage for OAuth, localStorage for email confirmation). Use two separate objects:

```ts
const memSession: Record<string, string> = {};
const memLocal: Record<string, string> = {};
vi.stubGlobal('sessionStorage', makeMock(memSession));
vi.stubGlobal('localStorage', makeMock(memLocal));
```

apps/web/src/auth/__tests__/planSignupBullets.test.ts
```diff
@@ -0,0 +1,19 @@
+import { describe, expect, it } from 'vitest';
+import { planSignupBullets } from '../planSignupBullets';
+
+describe('planSignupBullets', () => {
+  it('returns bullets for Pro', () => {
+    const b = planSignupBullets('beakerstack_pro');
+    expect(b.length).toBeGreaterThan(0);
+    expect(b.some(x => /Feature A/i.test(x))).toBe(true);
```
rahul-artificerinnovations: Non-blocking — `expect(b.some(x => /Feature A/i.test(x))).toBe(true)` and `expect(b.some(x => /trial/i.test(x))).toBe(true)` are tightly coupled to specific string values in `beakerstackBillingConfig`. If the config copy changes these tests break without the bullets being wrong. Consider asserting structure instead: `expect(b.length).toBeGreaterThan(0)` + `expect(b.every(x => typeof x === 'string' && x.length > 0)).toBe(true)` covers what the function contract actually guarantees.

rahul-artificerinnovations (changes requested):
Overall the implementation is solid and faithfully follows the revised plan from issue #68 — `resolvePostAuthDestination` is correctly synchronous and static, the TTL+JSON storage payload is a nice hardening over raw paths, `welcomeSnapshot` with `useState` initializer is the right pattern to prevent the banner from blinking as the URL cleans up, and the `navigatedRef` guard on `AuthCallbackPage` prevents double-navigate on auth state re-runs. Two blocking issues and a few non-blocking notes below.

apps/web/src/pages/billing/BillingPlansPage.tsx
```diff
@@ -69,6 +74,32 @@ export default function BillingPlansPage() {
...
+  const searchKey = search.toString();
+
+  useEffect(() => {
+    const sp = new URLSearchParams(searchKey);
+    if (sp.get('welcome') !== '1' || !welcomePlanId || catLoad) return;
+    const raf = requestAnimationFrame(() => {
+      document
+        .getElementById(`plan-card-${welcomePlanId}`)
+        ?.scrollIntoView({ behavior: 'smooth', block: 'center' });
+    });
```
diego-artificerinnovations: **Bug: the scroll RAF is cancelled before it fires.**

The effect queues a `requestAnimationFrame`, then synchronously calls `setSearchParams(sp, { replace: true })`. That triggers a React re-render which runs the effect's cleanup — `cancelAnimationFrame(raf)` — before the animation frame fires. The scroll never executes.

Fix: don't cancel the RAF from this cleanup, or skip the RAF entirely since `useEffect` already runs post-paint and the plan cards are guaranteed rendered when `catLoad` is false:

```ts
useEffect(() => {
  const sp = new URLSearchParams(searchKey);
  if (sp.get('welcome') !== '1' || !welcomePlanId || catLoad) return;
  document
    .getElementById(`plan-card-${welcomePlanId}`)
    ?.scrollIntoView({ behavior: 'smooth', block: 'center' });
  sp.delete('welcome');
  setSearchParams(sp, { replace: true });
}, [searchKey, setSearchParams, welcomePlanId, catLoad]);
```

If you want to keep the RAF for scroll-after-repaint reasons, just remove the `return () => cancelAnimationFrame(raf)` line — the optional chaining on `getElementById` makes it safe if the element is gone on a re-run.

apps/web/src/components/auth/SignupPlanSummary.tsx
```diff
@@ -0,0 +1,137 @@
...
+export function PlanIntentSummary({
+  mode = 'signup',
+}: {
+  mode?: PlanIntentMode;
+}): JSX.Element | null {
+  const [search] = useSearchParams();
+  const planId = search.get('plan');
+  const cadence = getCadenceFromSearch(search);
+  const { plans, loading } = usePlanCatalog<typeof beakerstackBillingConfig>();
+
```
diego-artificerinnovations, line 29: The issue discussion resolved that the signup summary should use static config + the same formatters as the landing page, because `usePlanCatalog` may fail or return empty for unauthenticated users. This component uses the hook instead.

If the plans table has public RLS this works fine, but it's worth confirming explicitly — if the catalog fetch requires a session, the summary will show "Loading plan…" forever on the signup page (before the user has an account).

Fallback: if `loading` never resolves, the component returns the loading skeleton indefinitely. A short timeout or a static-config fallback would be safer here if unauthenticated catalog access isn't guaranteed.

apps/web/src/pages/__tests__/AuthCallbackPage.coverage.test.tsx
```diff
@@ -36,10 +36,27 @@ vi.mock('@beakerstack/shared/contexts/AuthContext', () => ({
 
 describe('AuthCallbackPage (URL + auth branches)', () => {
   const original = window.location;
+  const memStore: Record<string, string> = {};
+
+  const storageMock = {
+    getItem: (k: string) => (k in memStore ? memStore[k] : null),
+    setItem: (k: string, v: string) => {
```
diego-artificerinnovations: `sessionStorage` and `localStorage` are both stubbed to the same `memStore` object. For the current tests this is harmless, but any future test that stashes data in one store and expects it absent in the other will give a false positive. Worth using two separate `memStore` objects so the stores are independent.

apps/web/src/pages/LoginPage.tsx
```diff
@@ -1,18 +1,45 @@
...
-import { Link, useNavigate } from 'react-router-dom';
+import { Link, useNavigate, useSearchParams } from 'react-router-dom';
 import { useAuthContext } from '@beakerstack/shared/contexts/AuthContext';
 import { AppHeader } from '@beakerstack/shared/components/navigation/AppHeader.web';
 import { supabase } from '@/lib/supabase';
 import { SocialLoginButton } from '../components/SocialLoginButton';
+import { LoginPlanSummary } from '../components/auth/SignupPlanSummary';
+import { beakerstackBillingConfig } from '../billing/beakerstackBillingConfig';
+import {
+  POST_AUTH_REDIRECT_KEY,
```
diego-artificerinnovations, line 12: `appBasePath()` is defined identically in both `SignupPage.tsx` and `LoginPage.tsx`. Easy candidate for a shared utility (e.g. `apps/web/src/lib/appBasePath.ts`) — not a blocker but will drift if the logic ever changes.

diego-artificerinnovations (changes requested):
Overall the implementation is solid — the resolver is synchronous and uses static config correctly, the TTL payload is a nice defensive touch, and the `navigatedRef` guard in `AuthCallbackPage` prevents double-navigation cleanly. One bug and a few non-blocking concerns below.

apps/web/src/pages/SignupPage.tsx
```diff
@@ -32,7 +62,20 @@ export default function SignupPage() {
 
     try {
       await auth.signUp(email, password);
-      navigate('/dashboard');
+      const { data: { session } } = await supabase.auth.getSession();
+
+      if (session) {
+        localStorage.removeItem(POST_AUTH_REDIRECT_KEY);
```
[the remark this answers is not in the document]
ZappoMan: Thanks — fixed: successful email signup now calls `clearPostAuthRedirectKeys()`, which removes `POST_AUTH_REDIRECT_KEY` from **both** `sessionStorage` and `localStorage`, so a stale OAuth stash cannot affect a later callback.

apps/web/src/pages/LoginPage.tsx
```diff
@@ -26,7 +53,8 @@ export default function LoginPage() {
 
     try {
       await auth.signIn(email, password);
-      navigate('/dashboard');
+      localStorage.removeItem(POST_AUTH_REDIRECT_KEY);
```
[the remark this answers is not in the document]
ZappoMan: Fixed — password login success now calls `clearPostAuthRedirectKeys()` so `POST_AUTH_REDIRECT_KEY` is removed from **both** `sessionStorage` and `localStorage`, matching signup and avoiding a stale OAuth stash in this tab.

apps/web/src/pages/AuthCallbackPage.tsx
```diff
@@ -22,39 +26,58 @@ export default function AuthCallbackPage() {
...
+        navigate(stored ?? '/dashboard', { replace: true });
+      } else if (!a.loading) {
+        setError(
+          'Authentication completed but session not established. Please try again.'
+        );
+        setTimeout(() => {
+          navigatedRef.current = true;
+          navigate('/login', { replace: true });
+        }, 3000);
+      }
```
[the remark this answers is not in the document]
ZappoMan: Fixed — the delayed redirect-to-login `setTimeout` is stored in `delayedLoginTimerRef` and cleared in the effect cleanup (alongside the outer timer), so navigation cannot fire after unmount.

apps/web/src/auth/postAuthRedirect.ts
```diff
@@ -0,0 +1,89 @@
...
+  if (!raw) return null;
+  try {
+    const o = JSON.parse(raw) as { path?: string; ts?: number };
+    if (typeof o.path === 'string' && typeof o.ts === 'number') {
+      if (now - o.ts > POST_AUTH_REDIRECT_TTL_MS) return null;
+      return validateInternalPostAuthPath(o.path, origin);
+    }
+    return null;
+  } catch {
+    return validateInternalPostAuthPath(raw, origin);
```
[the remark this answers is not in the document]
ZappoMan: Aligned — `parseStoredPostAuthRedirect` no longer treats non-JSON payloads as paths: malformed/non-JSON hits the `catch` and returns `null`, so TTL is always enforced for accepted payloads. Doc comment updated to match.

apps/web/src/pages/SignupPage.tsx
```diff
@@ -1,19 +1,49 @@
...
+  const [awaitingEmail, setAwaitingEmail] = useState(false);
   const navigate = useNavigate();
+  const [searchParams] = useSearchParams();
   const auth = useAuthContext();
 
+  const paidIntent = hasPaidPlanIntent(searchParams);
+  const postAuthPath = resolvePostAuthDestination(searchParams);
+  const loginSearch = searchParams.toString();
+  const loginTo = loginSearch ? `/login?${loginSearch}` : '/login';
+
```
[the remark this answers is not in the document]
ZappoMan: Added coverage in `SignupPage.test.tsx`: paid `?plan=` CTA, redirects to billing vs dashboard vs awaiting-email, signup/API errors, Google OAuth failure clearing session stash, and related navigation mocks.

apps/web/src/pages/LoginPage.tsx
```diff
@@ -1,18 +1,45 @@
...
   const [isLoading, setIsLoading] = useState(false);
   const [error, setError] = useState<string | null>(null);
   const navigate = useNavigate();
+  const [searchParams] = useSearchParams();
   const auth = useAuthContext();
 
+  const postAuthPath = resolvePostAuthDestination(searchParams);
+  const signupSearch = searchParams.toString();
+  const signupTo = signupSearch ? `/signup?${signupSearch}` : '/signup';
+
```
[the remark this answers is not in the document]
ZappoMan: Added `LoginPage.test.tsx` coverage for `?plan=beakerstack_pro` (plan aside + resolved billing destination on navigate) with billing hooks mocked, alongside existing flows.

apps/web/src/pages/AuthCallbackPage.tsx
ZappoMan: Done — both navigation paths now call shared `readAndClearPostAuthRedirect()` from `postAuthRedirect.ts`, which reads session + local, clears both stores once, parses TTL/validation, then returns the destination — same audit surface as your suggested `doNavigate()` helper.

apps/web/src/pages/SignupPage.tsx
ZappoMan: Fixed — `SignupPage` now mirrors `LoginPage`: `showPlanAside = paidIntent && postAuthPath !== '/dashboard'`, grid layout only when the aside renders; plain `/signup` uses full-width `space-y-8`.

apps/web/src/auth/planSignupBullets.ts
ZappoMan: Fixed — numeric feature rows now `continue` when `typeof val !== 'number'`, so missing keys never produce `Up to undefined…` bullets.

apps/web/src/pages/AuthCallbackPage.tsx
ZappoMan: Addressed via `readAndClearPostAuthRedirect()` (same helper as the Copilot/sarah thread): single implementation for read+clear+parse so future key/path logic stays one place.

apps/web/src/pages/__tests__/AuthCallbackPage.coverage.test.tsx
ZappoMan: Updated — `AuthCallbackPage.coverage.test.tsx` uses **separate** in-memory stores for `sessionStorage` vs `localStorage` so fallback ordering is realistic.

apps/web/src/auth/__tests__/planSignupBullets.test.ts
ZappoMan: Took the intent — tests still assert non-empty string bullets and caps; we added broader branch coverage (Free/Pro/Max paths, meter line) while keeping a few label checks where they catch regressions in template wiring. Happy to swap remaining copy assertions for pure structure-only checks if you want zero coupling to wording.

apps/web/src/pages/billing/BillingPlansPage.tsx
ZappoMan: Bug confirmed and fixed — dropped `requestAnimationFrame` + `cancelAnimationFrame` from the welcome scroll effect. We now `scrollIntoView` directly once `catLoad` is false (effect already runs post-layout), then strip `welcome` with `replace: true`, so cleanup cannot cancel the scroll.

apps/web/src/components/auth/SignupPlanSummary.tsx
ZappoMan: Good call-out — we kept `usePlanCatalog` so signup/login match landing/pricing numbers and Stripe-linked catalog rows. In this template, `billing_plans` is intended to be readable for anonymous catalog sync (same query path as the landing page). If a fork locks down RLS, we should add a static fallback or timeout; tracked as a follow-up unless you want it in this PR.

apps/web/src/pages/__tests__/AuthCallbackPage.coverage.test.tsx
ZappoMan: Same fix as Rahul's thread — coverage tests now stub session vs local with independent backing objects.

apps/web/src/pages/LoginPage.tsx
ZappoMan: Extracted — shared `appBasePath()` lives in `apps/web/src/lib/appBasePath.ts` and `SignupPage` / `LoginPage` (and `BillingProviderLayout`) import it so preview base-path logic stays single-sourced.

ZappoMan:
## Review responses

Threaded replies are posted **on each inline comment** on this PR with how the feedback was addressed.

**Summary-only notes** (from full reviews without a separate inline anchor):

- **Mei — OAuth stash on failure:** `sessionStorage.removeItem(POST_AUTH_REDIRECT_KEY)` is called in the Google OAuth `catch` on both Signup and Login so a failed/cancelled OAuth attempt does not leave stale intent for a later callback (TTL still bounds legacy cases).
- **Mei — bullet list keys:** Plan bullets use stable keys `${mode}-bullet-${i}`.
- **Sarah — AuthCallback stored-path coverage:** Additional tests exercise seeded session/local storage + navigation where we branch on stored redirects (see `AuthCallbackPage` / coverage tests).

Thanks again for the thorough review.

diego-artificerinnovations (approved):
All my blocking and minor issues are addressed:

- **RAF bug (BillingPlansPage):** Fixed — removed the RAF entirely and calls `scrollIntoView` synchronously before `setSearchParams`. Clean.
- **Shared mock store (AuthCallbackPage tests):** Fixed — `memSession` and `memLocal` are now separate backing objects, `storageMock` is a factory. The new post-auth redirect test cases look correct.
- **`appBasePath` duplication:** Fixed — extracted to `apps/web/src/lib/appBasePath.ts`, imported in both pages.

The `usePlanCatalog` on the unauthenticated signup page was non-blocking and I'll leave that as a known trade-off — if the catalog endpoint ever requires auth, the loading skeleton will hang, but that's a follow-up concern for when auth requirements are confirmed.

rahul-artificerinnovations (approved):
Both blocking issues are resolved — approving.

**Blocking #1 (planSignupBullets.ts numeric branch):** ✅ `if (typeof val !== 'number') continue;` guard is in place and the type is correctly widened to `Record<string, number | boolean | undefined>`. No more phantom "Up to undefined items" bullets.

**Blocking #2 (SignupPage.tsx layout):** ✅ `showPlanAside` conditional is correct — `paidIntent && postAuthPath !== '/dashboard'` — and the grid vs. single-column class swap matches the LoginPage pattern exactly.

Non-blocking items also addressed:
- `readAndClearPostAuthRedirect()` extracted and imported in AuthCallbackPage — clean dedup.
- `delayedLoginTimerRef` cleanup added.
- Coverage tests now have separate `memSession`/`memLocal` backing stores and cover both storage paths.
- `planSignupBullets.test.ts` first test is structure-based; later tests intentionally use known config values — that's fine.

Good work on a complex flow. Ready to merge once CI is green.

mei-artificerinnovations (approved):
All four nits from the first round are addressed:

- `sessionStorage.removeItem(POST_AUTH_REDIRECT_KEY)` in the catch block of both `SignupPage` and `LoginPage` — stale OAuth intent no longer survives a cancelled popup.
- `AuthCallbackPage.coverage.test.tsx` now uses separate `memSession` / `memLocal` objects; `beforeEach` clears both, `afterEach` calls `vi.unstubAllGlobals()`. The storage isolation that was missing is now there.
- Bullet key changed to `\`${mode}-bullet-${i}\`` — index-based, no collision risk.
- `out.slice(0, 3)` removed from `planSignupBullets`.

Bonus improvements that weren't asked for but are good to have: `readAndClearPostAuthRedirect` extracted as a named function (cleaner callsites in `AuthCallbackPage`), `serializePostAuthRedirectPayload` exported for test use, legacy raw-path fallback in `parseStoredPostAuthRedirect` removed (tightens the surface). The new `postAuthRedirect.test.ts` unit suite is thorough — expiry, open-redirect rejections, dual-store fallback all covered.

CI green, implementation is correct. Approved.

sarah-artificerinnovations (approved):
## PR #69 re-review — Approving ✅

All three feedback points addressed well.

**Duplicate storage-read block** — resolved cleanly by extracting `readAndClearPostAuthRedirect()` into `postAuthRedirect.ts`. Both call sites in `AuthCallbackPage` are now a single call. The bonus `clearPostAuthRedirectKeys()` export is a useful utility. Also noted: `delayedLoginTimerRef` was added to clean up the inner `setTimeout` on unmount — good catch, improves the cleanup.

**Redundant `!cfg.isPublic` check** — removed from `resolvePostAuthDestination`. The non-JSON fallback in `parseStoredPostAuthRedirect`'s catch block was also cleaned up (now returns `null` rather than falling through to validate the raw string). Both are correct.

**`appBasePath`** — extracted to `apps/web/src/lib/appBasePath.ts`, imported by both `SignupPage` and `LoginPage`. Clean.

**Storage path tests** — two new cases in `AuthCallbackPage.coverage.test.tsx`: sessionStorage read → billing plans navigation + storage cleared, and localStorage fallback when sessionStorage is empty. `memSession`/`memLocal` are now separate objects (better isolation than the shared `memStore` before). Coverage gap closed.

Ready to merge.

PR #76, opened 2026-05-12
feat: web-only / mobile-optional template path

### LINKED ISSUE #74: Optional mobile setup and CI: web-only template path
## Problem / business need

Many teams adopt this stack for **web-only** products. Today the interactive setup and CI still surface **mobile-specific configuration** (Expo/EAS, Google mobile OAuth artifacts, related GitHub variables and secrets) and **mobile deploy steps** in workflows such as PR previews and staging/production deploys. That creates friction: extra prompts users do not understand, secrets they do not have, and CI time spent on paths they will never use.

We need a **first-class "web only" path** so template users can complete setup quickly, avoid mobile credentials entirely when they opt out, and get **faster, clearer CI** that does not spend cycles or required secrets on mobile unless mobile is explicitly enabled for the repo.

Separating **web** vs **mobile** automation (e.g. distinct GitHub Actions workflows or jobs with clear triggers) should reduce noise in logs, make failures easier to interpret, and allow mobile pipelines to be skipped entirely when disabled—without weakening the experience for teams that do want mobile.

## Acceptance criteria

1. **Explicit opt-in during setup**  
   The main setup flow (`scripts/setup-full.mjs` and related menu/docs) asks whether the installer wants **mobile builds / Expo** enabled for this project. The default should favor clarity (e.g. explicit Yes/No), not accidental mobile configuration.

2. **Skipped configuration when mobile is disabled**  
   If the user opts out of mobile:
   - The setup **does not prompt** for mobile-only inputs (e.g. Expo/EAS, Google Services for native apps, and any other secrets clearly scoped to `apps/mobile` or mobile CI).
   - The setup **does not write or sync** GitHub variables/secrets that exist only to support mobile CI or mobile preview/deploy, and local/generated env files remain consistent with a web-only posture.

3. **Single source of truth**  
   There is a durable, machine-readable indicator (e.g. repo config, generated setup state, or GitHub variable) that **both** local tooling and CI can read to mean "mobile disabled for this repo," so behavior does not depend on absent secrets alone.

4. **CI optimized for web-only repos**  
   When mobile is disabled:
   - Workflows **do not run** mobile build/deploy jobs/steps (including PR preview mobile deploy and staging/production mobile deploy), or they **exit immediately** without checkout/build/deploy work for mobile—so queue time, runner minutes, and log noise are minimized.
   - PR comments / deployment summaries **do not imply** a failed or missing mobile preview when mobile was intentionally disabled (wording should reflect "mobile not configured" vs error).

5. **CI unchanged for mobile-enabled repos**  
   When mobile is enabled, existing mobile preview and deploy behavior remains available; required secrets and variables are documented and still collected during setup.

6. **Separation of web vs mobile automation (directional)**  
   GitHub Actions are structured so **web** build/deploy/test paths can run **without** executing mobile workflows or jobs (e.g. separate workflows or clearly gated jobs with path/trigger conditions). The exact split is an implementation detail; the outcome is independent scheduling and easier maintenance.

7. **Documentation**  
   Short, discoverable notes explain when to enable mobile, what extra secrets are needed, and that web-only adopters can leave mobile off without breaking CI.

sarah-artificerinnovations:
Closes #74

## Summary

- **`MOBILE_ENABLED` GitHub variable** is the single source of truth. Absent or `true` = mobile enabled (backward-compatible). `false` = web-only mode.
- **`setup-full.mjs`**: adds `--skip-mobile` flag + interactive `[Y/n]` prompt before the expo phase; gates expo/google phases; writes `mobileEnabled` to state JSON. When resuming (`--from=<phase>`), honours any stored `MOBILE_ENABLED` value from env files instead of re-prompting with an enabled default.
- **`setup-manifest.mjs`**: adds `MOBILE_ENABLED` to `GITHUB_VARIABLES`; new `MOBILE_GROUPS` constant gates expo/google secrets/variables from GitHub sync and CI-completeness warnings when disabled.
- **`pr-preview-environment.yml`**: splits mobile deploy into a separate `deploy-preview-mobile` job (gated by `vars.MOBILE_ENABLED != 'false'`); adds a dedicated `comment` job that waits on both and handles skipped/failed mobile gracefully. Web deploy no longer blocked by mobile build time. `comment` job has both `pull-requests: write` and `issues: write` (required by `github.rest.issues.*`).
- **`deploy-staging.yml`** and **`deploy-production.yml`**: mobile step extracted to `deploy-mobile-staging` / `deploy-mobile-production` jobs, each gated by `vars.MOBILE_ENABLED != 'false'`. Missing `EXPO_ACCOUNT` fails loudly only when `MOBILE_ENABLED=true`; otherwise exits 0 with a warning (preserves backwards compat for repos with neither variable set).
- **`QUICKSTART.md`**: documents `--skip-mobile` flag; notes Expo/Google secrets are mobile-only.

## Security

Per Mei's review: `PREVIEW_SUPABASE_ANON_KEY` is read directly from `secrets.*` in the mobile job — not passed as a cross-job output — so it never appears in plaintext in workflow logs.

## CloudFront function publish timing

The initial commit accidentally removed `--skip-cloudfront-function-publish` from the bootstrap call and dropped the "Publish PR path router" step entirely. This was an **unintentional refactoring side effect**, not a deliberate change. Commit c49c1c8 restores the original behaviour:

1. `bootstrap-aws-stack.sh` is called with `--skip-cloudfront-function-publish` — creates/updates the stack without touching CloudFront function code.
2. A separate "Publish PR path router (CloudFront)" step runs **only after `steps.web.outcome == 'success'`**, ensuring S3 assets are in place before the routing function is updated.

This preserves the safety guard from PR #72: a failed web build cannot replace live CloudFront function code while leaving stale or missing S3 assets.

## Test plan

- [ ] Web-only: set `MOBILE_ENABLED=false` → mobile jobs skipped, PR comment shows only web URL
- [ ] Mobile-enabled: leave `MOBILE_ENABLED` unset → both jobs run, comment shows web + mobile
- [ ] Explicitly set `MOBILE_ENABLED=true` → same as unset (both enable mobile; `!= 'false'` gate treats them identically)
- [ ] Existing repo with neither `MOBILE_ENABLED` nor `EXPO_ACCOUNT` set → `deploy-mobile-*` jobs skip with warning, not failure
- [ ] `npm run setup:full -- --skip-mobile --dry-run` → expo/google phases skipped
- [ ] `npm run setup:full -- --dry-run`, answer `n` at mobile prompt → same result
- [ ] Resume with `--from=aws` after a run that answered `n` → expo/google still skipped (stored value honoured, no re-prompt)
- [ ] `npm run docs:actions-secrets` regenerates without drift

### DIFF
--- a/.github/workflows/pr-preview-environment.yml
+++ b/.github/workflows/pr-preview-environment.yml
@@ -51,6 +51,8 @@ jobs:
     if: github.event.action != 'closed' && needs.preview-scope.outputs.run_deploy == 'true' && (github.base_ref != 'develop' || github.head_ref != 'main')
     needs: preview-scope
     runs-on: ubuntu-latest
+    outputs:
+      preview-web-url: ${{ steps.web.outputs.PREVIEW_WEB_URL }}
     env:
       PREVIEW_DOMAIN: ${{ vars.PR_PREVIEW_DOMAIN }}
       PREVIEW_HOSTED_ZONE_ID: ${{ vars.PR_PREVIEW_HOSTED_ZONE_ID }}
@@ -103,8 +105,7 @@ jobs:
             --domain "${PREVIEW_DOMAIN}" \
             --hosted-zone-id "${PREVIEW_HOSTED_ZONE_ID}" \
             --certificate-arn "${PREVIEW_CERTIFICATE_ARN}" \
-            --preview-prefix "${PREVIEW_PREFIX_VALUE}" \
-            --skip-cloudfront-function-publish
+            --preview-prefix "${PREVIEW_PREFIX_VALUE}"
 
       - name: Prepare Supabase preview database
         id: supabase
@@ -193,31 +194,34 @@ jobs:
             --preview-prefix "${PREVIEW_PREFIX_VALUE}" \
             --region "${AWS_REGION_VALUE}"
 
-      # Publish edge routing only after a successful web build so a failed deploy cannot
-      # replace live CloudFront function code while leaving stale or missing S3 assets.
-      - name: Publish PR path router (CloudFront)
-        if: steps.web.outcome == 'success'
-        shell: bash
-        run: |
-          set -euo pipefail
-          STACK_NAME="${PREVIEW_STACK_NAME:-beakerstack-pr-preview}"
-          AWS_REGION_VALUE="${PREVIEW_AWS_REGION:-us-east-1}"
-          PREVIEW_PREFIX_VALUE="${PREVIEW_PREFIX:-pr-}"
-          ./scripts/pr-preview/bootstrap-aws-stack.sh \
-            --stack-name "${STACK_NAME}" \
-            --region "${AWS_REGION_VALUE}" \
-            --domain "${PREVIEW_DOMAIN}" \
-            --hosted-zone-id "${PREVIEW_HOSTED_ZONE_ID}" \
-            --certificate-arn "${PREVIEW_CERTIFICATE_ARN}" \
-            --preview-prefix "${PREVIEW_PREFIX_VALUE}" \
-            --if-stack-exists refresh-outputs
+  deploy-preview-mobile:
+    if: github.event.action != 'closed' && needs.preview-scope.outputs.run_deploy == 'true' && (github.base_ref != 'develop' || github.head_ref != 'main') && vars.MOBILE_ENABLED != 'false'
+    needs: [preview-scope, deploy-preview]
+    runs-on: ubuntu-latest
+    outputs:
+      mobile-channel: ${{ steps.mobile.outputs.PREVIEW_MOBILE_CHANNEL }}
+      mobile-update-url: ${{ steps.mobile.outputs.PREVIEW_MOBILE_UPDATE_URL }}
+      mobile-ios-download-url: ${{ steps.mobile.outputs.PREVIEW_MOBILE_IOS_DOWNLOAD_URL }}
+      mobile-android-download-url: ${{ steps.mobile.outputs.PREVIEW_MOBILE_ANDROID_DOWNLOAD_URL }}
+      needs-native-build: ${{ steps.native-changes.outputs.needs_native_build }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
 
       - name: Check for native code changes
         id: native-changes
         shell: bash
-        env:
-          PR_NUMBER: ${{ github.event.pull_request.number }}
-          EXPO_TOKEN: ${{ secrets.EXPO_TOKEN }}
         run: |
           set -euo pipefail
           BASE_REF="origin/${{ github.base_ref }}"
@@ -237,7 +241,6 @@ jobs:
           EXPO_PROJECT_ID: ${{ secrets.EXPO_PROJECT_ID }}
           EXPO_PUBLIC_SUPABASE_URL: ${{ secrets.PREVIEW_SUPABASE_URL }}
           EXPO_PUBLIC_SUPABASE_ANON_KEY: ${{ secrets.PREVIEW_SUPABASE_ANON_KEY }}
-          # Google Services variables (needed for google-services.json generation)
           GOOGLE_SERVICES_PROJECT_NUMBER: ${{ secrets.GOOGLE_SERVICES_PROJECT_NUMBER }}
           GOOGLE_SERVICES_PROJECT_ID: ${{ secrets.GOOGLE_SERVICES_PROJECT_ID }}
           GOOGLE_SERVICES_STORAGE_BUCKET: ${{ secrets.GOOGLE_SERVICES_STORAGE_BUCKET }}
@@ -252,8 +255,8 @@ jobs:
           PR_NUMBER_VALUE="${PR_NUMBER}"
           EXPO_ACCOUNT_VALUE="${{ vars.EXPO_ACCOUNT }}"
           if [[ -z "${EXPO_ACCOUNT_VALUE}" ]]; then
-            echo "Warning: EXPO_ACCOUNT not set, skipping mobile deployment"
-            exit 0
+            echo "::error::EXPO_ACCOUNT variable is not set. Set vars.MOBILE_ENABLED=false to disable mobile, or set EXPO_ACCOUNT to enable it."
+            exit 1
           fi
           BUILD_NATIVE_FLAG=""
           if [[ "${{ steps.native-changes.outputs.needs_native_build }}" == "true" ]]; then
@@ -268,21 +271,28 @@ jobs:
             --platform all \
             ${BUILD_NATIVE_FLAG}
 
+  comment:
+    if: always() && needs.deploy-preview.result == 'success'
+    needs: [deploy-preview, deploy-preview-mobile]
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
       - name: Post preview summary comment
-        if: always()
         uses: actions/github-script@v7
         env:
-          WEB_URL: ${{ steps.web.outputs.PREVIEW_WEB_URL }}
-          MOBILE_CHANNEL: ${{ steps.mobile.outputs.PREVIEW_MOBILE_CHANNEL }}
-          MOBILE_UPDATE_URL: ${{ steps.mobile.outputs.PREVIEW_MOBILE_UPDATE_URL }}
-          MOBILE_IOS_DOWNLOAD_URL: ${{ steps.mobile.outputs.PREVIEW_MOBILE_IOS_DOWNLOAD_URL }}
-          MOBILE_ANDROID_DOWNLOAD_URL: ${{ steps.mobile.outputs.PREVIEW_MOBILE_ANDROID_DOWNLOAD_URL }}
-          NEEDS_NATIVE_BUILD: ${{ steps.native-changes.outputs.needs_native_build }}
-          MOBILE_DEPLOY_OUTCOME: ${{ steps.mobile.outcome }}
+          WEB_URL: ${{ needs.deploy-preview.outputs.preview-web-url }}
+          MOBILE_JOB_RESULT: ${{ needs.deploy-preview-mobile.result }}
+          MOBILE_CHANNEL: ${{ needs.deploy-preview-mobile.outputs.mobile-channel }}
+          MOBILE_UPDATE_URL: ${{ needs.deploy-preview-mobile.outputs.mobile-update-url }}
+          MOBILE_IOS_DOWNLOAD_URL: ${{ needs.deploy-preview-mobile.outputs.mobile-ios-download-url }}
+          MOBILE_ANDROID_DOWNLOAD_URL: ${{ needs.deploy-preview-mobile.outputs.mobile-android-download-url }}
+          NEEDS_NATIVE_BUILD: ${{ needs.deploy-preview-mobile.outputs.needs-native-build }}
         with:
           script: |
             const marker = '<!-- pr-preview-links -->';
             const webUrl = process.env.WEB_URL;
+            const mobileJobResult = process.env.MOBILE_JOB_RESULT;
             const mobileChannel = process.env.MOBILE_CHANNEL;
             const mobileUpdateUrl = process.env.MOBILE_UPDATE_URL;
 
@@ -291,7 +301,12 @@ jobs:
             if (webUrl) {
               lines.push(`🌐 **Web preview:** ${webUrl}`);
             }
-            if (mobileUpdateUrl) {
+            if (mobileJobResult === 'skipped') {
+              // Mobile disabled via MOBILE_ENABLED=false — omit mobile section
+            } else if (mobileJobResult === 'failure') {
+              lines.push('');
+              lines.push('❌ **Mobile preview failed.** Check the **deploy-preview-mobile** job logs for details.');
+            } else if (mobileUpdateUrl) {
               const iosDownloadUrl = process.env.MOBILE_IOS_DOWNLOAD_URL;
               const androidDownloadUrl = process.env.MOBILE_ANDROID_DOWNLOAD_URL;
               const needsNativeBuild = process.env.NEEDS_NATIVE_BUILD === 'true';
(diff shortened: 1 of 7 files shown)


rahul-artificerinnovations (changes requested):
Base branch is `develop` ✓. The core design is solid: `MOBILE_ENABLED` source of truth, `!= 'false'` backwards-compat condition, `MOBILE_GROUPS` set in the manifest, and the `comment` summary job with `always()` correctly resolves the cross-job visibility gap from the plan review. `[Y/n]` prompt default incorporated per Brad's feedback.

One blocking issue, plus two worth a brief note.

---

### 🔴 Blocking — `exit 0` → `exit 1` breaks backwards compatibility

This change appears in all three workflows (`pr-preview-environment.yml`, `deploy-staging.yml`, `deploy-production.yml`) in the `EXPO_ACCOUNT` absent check.

The plan's backwards-compat guarantee was: repos with no `MOBILE_ENABLED` set have `vars.MOBILE_ENABLED != 'false'` evaluate to `true`, so mobile jobs run **exactly as today**. But "exactly as today" meant `exit 0` (silent skip) when `EXPO_ACCOUNT` was absent. Changing to `exit 1` means repos that currently have working CI — no `MOBILE_ENABLED` set, no `EXPO_ACCOUNT` — will now **fail CI on every deploy** until they either re-run setup or manually set `MOBILE_ENABLED=false`.

Fix options:

1. **Preferred — grace period condition:** Only `exit 1` when `MOBILE_ENABLED` is *explicitly* `'true'`; otherwise warn and skip:
```yaml
if [[ -z "${EXPO_ACCOUNT_VALUE}" ]]; then
  if [[ "${{ vars.MOBILE_ENABLED }}" == 'true' ]]; then
    echo "::error::EXPO_ACCOUNT not set but MOBILE_ENABLED=true. Set EXPO_ACCOUNT or set MOBILE_ENABLED=false to disable mobile."
    exit 1
  else
    echo "::warning::EXPO_ACCOUNT not set — skipping mobile deploy. Set MOBILE_ENABLED=false to suppress this warning."
    exit 0
  fi
fi
```
2. **Alternative:** Keep `exit 1` but add a migration note to QUICKSTART.md stating existing repos must set `MOBILE_ENABLED=false` (or re-run setup) before merging this change.

---

### 🟡 `--skip-cloudfront-function-publish` removed, conditional publish step also gone

The old `deploy-preview` job had a two-step CloudFront publish: first bootstrap without publishing the function, then conditionally publish only after web deploy succeeded. The comment explained the rationale: _"Publish edge routing only after a successful web build so a failed deploy cannot replace live CloudFront function code while leaving stale or missing S3 assets."_

This PR removes both the flag and the conditional second call, so the CloudFront function is now published at bootstrap time — before S3 assets exist. For PR previews this is lower risk than production, but the guard and its explanation are gone. If this is intentional, a one-line note in the PR description is enough.

---

### 🟡 `google` phase skips without `clearGoogleKeysFromAcc`

When the expo phase is skipped, `clearExpoKeysFromAcc(acc)` is called. The google phase is also skipped but no equivalent clear is called — old `GOOGLE_SERVICES_*` keys from a previous run with mobile enabled could persist in the accumulator and end up in generated env files. May be intentional, but it's asymmetric with the expo handling.

.github/workflows/pr-preview-environment.yml
```diff
@@ -252,8 +255,8 @@ jobs:
           PR_NUMBER_VALUE="${PR_NUMBER}"
           EXPO_ACCOUNT_VALUE="${{ vars.EXPO_ACCOUNT }}"
           if [[ -z "${EXPO_ACCOUNT_VALUE}" ]]; then
-            echo "Warning: EXPO_ACCOUNT not set, skipping mobile deployment"
-            exit 0
+            echo "::error::EXPO_ACCOUNT variable is not set. Set vars.MOBILE_ENABLED=false to disable mobile, or set EXPO_ACCOUNT to enable it."
```
diego-artificerinnovations: `PR_NUMBER` was removed from the `check-native-changes` step env and was never added to the `deploy-preview-mobile` job-level env. The mobile deploy step does:

```bash
set -euo pipefail
PR_NUMBER_VALUE="${PR_NUMBER}"
```

With `-u` (unbound variable exits), this fails on every mobile PR preview deploy because `PR_NUMBER` is not in scope. The same variable was also removed from `check-native-changes` (script handles it gracefully with `${PR_NUMBER:-}` — that part is non-fatal, it just loses the EAS build cache check).

Fix: add `PR_NUMBER` to the `deploy-preview-mobile` job-level `env:` block:

```yaml
  deploy-preview-mobile:
    env:
      PR_NUMBER: ${{ github.event.pull_request.number }}
      EXPO_TOKEN: ${{ secrets.EXPO_TOKEN }}
      EXPO_PROJECT_ID: ${{ secrets.EXPO_PROJECT_ID }}
      ...
```

(`EXPO_TOKEN` at job level would also restore the EAS build cache check in `check-native-changes`, which silently degrades without it.)

.github/workflows/pr-preview-environment.yml
```diff
@@ -103,8 +105,7 @@ jobs:
             --domain "${PREVIEW_DOMAIN}" \
             --hosted-zone-id "${PREVIEW_HOSTED_ZONE_ID}" \
             --certificate-arn "${PREVIEW_CERTIFICATE_ARN}" \
-            --preview-prefix "${PREVIEW_PREFIX_VALUE}" \
-            --skip-cloudfront-function-publish
+            --preview-prefix "${PREVIEW_PREFIX_VALUE}"
```
diego-artificerinnovations: The original code had `--skip-cloudfront-function-publish` on the initial bootstrap call, and a separate deferred step that published the CloudFront function only after `steps.web.outcome == 'success'`. The comment explained why: a failed web deploy should not replace live CloudFront function code with stale or missing S3 assets.

This PR removes `--skip-cloudfront-function-publish` from the bootstrap call and removes the deferred publish step entirely, so the function is now published at bootstrap time — before `npm ci`, before the web build, and before `deploy-web.sh` runs.

Was this intentional? For a PR preview environment (isolated stack, not production) the risk is lower. But the change should be called out explicitly in the PR body so the intent is clear. If it was a refactoring side effect, the safety pattern should be restored.

diego-artificerinnovations (changes requested):
The setup script changes, manifest filtering, and production/staging job splits are all correct — the `MOBILE_ENABLED` design is clean, `[Y/n]` default is right, and the `EXPO_ACCOUNT` now fails loudly instead of silently skipping. The `comment` job structure addresses the design gap I raised in the issue discussion.

One blocking bug in `pr-preview-environment.yml`, and one unverified change to the CloudFront function publish timing.

mei-artificerinnovations (commented):
Overall solid work — the job split is clean and the comment logic handles all the edge cases correctly. A few notes:

**Gate logic (`!= 'false'`)** — Intentionally backward-compatible (unset = enabled). Works as described. One gap in `QUICKSTART.md`: the note mentions `--skip-mobile` but does not call out that `MOBILE_ENABLED` defaults to enabled when absent. A one-liner like _"To run web-only, set `MOBILE_ENABLED=false` in your GitHub repo variables"_ would save the next team from a confused CI run.

**Comment job `if` condition** — `if: always() && needs.deploy-preview.result == 'success'` is the right pattern. Confirmed: `'skipped'` → mobile section omitted cleanly, `'failure'` → error row shown, web-only comment still posts. ✓

**CloudFront function publish timing** — Worth understanding the intent here. The removed `Publish PR path router` step had a comment explaining the rationale: _"publish only after a successful web build so a failed deploy cannot replace live CloudFront function code while leaving stale or missing S3 assets."_ Removing `--skip-cloudfront-function-publish` from the main bootstrap call means the function is now published before web assets land in S3. For a fresh preview environment this is fine (no traffic yet). For a re-deploy to an existing preview URL, there is a brief window where updated routing could point to S3 objects that have not yet been replaced. The blast radius is small (preview-only, short window) — but if this timing change is intentional, a note in the PR description explaining why would help reviewers and future maintainers.

Related: this interacts with the fix landing in PR #72 (the CloudFront function code itself). If that fix resolves the 503s end-to-end, the timing concern above becomes even lower risk — just flagging for awareness.

**`EXPO_ACCOUNT` loud failure** — The new error message is clear and actionable. Good change.

**`PREVIEW_SUPABASE_ANON_KEY`** — Confirmed read directly from `secrets.*` in the mobile job env block, not surfaced as a cross-job output. ✓

**Test plan gap** — Two cases not covered:
1. Explicitly set `MOBILE_ENABLED=true` (vs. unset) — the `!= 'false'` gate should treat both identically; worth a fast sanity check.
2. What happens to the comment job when the mobile job is *cancelled* (e.g., a force-push mid-run)? `mobileJobResult` would be `'cancelled'`, which falls through all three branches in the comment script — the mobile section would be silently absent rather than showing an explanatory note. Probably acceptable for a preview environment, but worth being deliberate about.

rahul-artificerinnovations (approved):
All three items from the first review are resolved.

**Blocking (backwards compat):** ✅ Grace-period condition correctly gates `exit 1` on `MOBILE_ENABLED == 'true'` (explicitly set); repos with the variable unset get `exit 0` + warning, preserving today's behaviour.

**`clearGoogleKeysFromAcc`:** ✅ Function added in `setup-google-services.mjs` and now called in the google phase skip — symmetrical with the expo path.

**CloudFront two-step publish:** ✅ `Publish PR path router` step with `if: steps.web.outcome == 'success'` is still present; only the explanatory comment was removed. Safety mechanism intact.

Approved. Ready to merge once CI is green.

mei-artificerinnovations (approved):
Fix commit addresses everything. Verified:

- **Backwards compat (`EXPO_ACCOUNT` check)** — only `exit 1` when `MOBILE_ENABLED` is explicitly `'true'`; absent/other exits 0 with a warning. Applied consistently across all three workflows. ✓
- **`PR_NUMBER` + `EXPO_TOKEN` in mobile job env** — added at job level, unblocks the deploy script under `set -euo pipefail` and restores EAS build cache check in `native-changes`. ✓
- **CloudFront function publish timing** — `--skip-cloudfront-function-publish` restored to the initial bootstrap call; deferred publish step re-added gated on `steps.web.outcome == 'success'`. Commit message confirms unintentional side effect. ✓
- **`clearGoogleKeysFromAcc`** — added to `setup-google-services.mjs`, called in the google phase skip, symmetric with expo handling. ✓

Also confirming: `QUICKSTART.md` already documented `MOBILE_ENABLED=false` in the original commit (I missed it in the initial pass — it was in the rewritten secrets-at-minimum paragraph). My concern there was already addressed.

Approving.

scripts/setup-full.mjs
```diff
@@ -1936,6 +1940,24 @@ async function main() {
...
+      if (mobileAns === 'n' || mobileAns === 'no') {
+        flags.mobileEnabled = false;
+        logInfo('Mobile disabled — skipping Expo, EAS, and Google Services setup.');
+      }
+    } else {
+      logInfo('[dry-run] would prompt for mobile setup choice (defaulting to enabled).');
+    }
+    acc.MOBILE_ENABLED = flags.mobileEnabled ? 'true' : 'false';
+  } else {
+    if (!acc.MOBILE_ENABLED) acc.MOBILE_ENABLED = 'true';
```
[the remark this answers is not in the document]
sarah-artificerinnovations: Good catch. Fixed in commit e03369a: when `acc.MOBILE_ENABLED` is already set (loaded from `.env*` files during a resume), the prompt is skipped and the stored value is used directly. A fresh run (no stored value) still shows the `[Y/n]` prompt as before.

.github/workflows/pr-preview-environment.yml
```diff
@@ -268,21 +271,28 @@ jobs:
             --platform all \
             ${BUILD_NATIVE_FLAG}
 
+  comment:
+    if: always() && needs.deploy-preview.result == 'success'
+    needs: [deploy-preview, deploy-preview-mobile]
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
```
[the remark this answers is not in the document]
sarah-artificerinnovations: Fixed in commit e03369a: added `issues: write` alongside `pull-requests: write` on the `comment` job. `github.rest.issues.listComments`, `createComment`, and `updateComment` all require it.

.github/workflows/pr-preview-environment.yml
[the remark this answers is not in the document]
sarah-artificerinnovations: Unintentional side effect — fixed in commit c49c1c8. `--skip-cloudfront-function-publish` is restored on the initial bootstrap call, and the deferred "Publish PR path router" step is back, gated on `steps.web.outcome == 'success'`. Details in the PR description.

.github/workflows/pr-preview-environment.yml
sarah-artificerinnovations: Fixed in commit c49c1c8: `PR_NUMBER` and `EXPO_TOKEN` added to the `deploy-preview-mobile` job-level `env:` block, making them available to all steps including `native-changes`.

.github/workflows/pr-preview-environment.yml
sarah-artificerinnovations: This was an unintentional refactoring side effect, not a deliberate change. Fixed in commit c49c1c8 — see the PR description for the full explanation.

sarah-artificerinnovations:
Thanks Rahul. All three items addressed in commit c49c1c8:
- **Backwards compat** — grace-period condition applied across all three workflows; only `exit 1` when `MOBILE_ENABLED` is explicitly `'true'`.
- **CloudFront guard** — `--skip-cloudfront-function-publish` restored and the deferred publish step re-added (gated on `steps.web.outcome == 'success'`).
- **`clearGoogleKeysFromAcc`** — added to `setup-google-services.mjs` and called in the google phase skip, symmetric with expo handling.

sarah-artificerinnovations:
Thanks Diego. Fixes in commit c49c1c8: `PR_NUMBER` + `EXPO_TOKEN` added at job level on `deploy-preview-mobile`, and the CloudFront timing change is confirmed unintentional — `--skip-cloudfront-function-publish` + the deferred publish step are both restored.

sarah-artificerinnovations:
Thanks for the thorough review, Mei.
- **`MOBILE_ENABLED` default** — added a one-liner to the QUICKSTART note clarifying it defaults to enabled when absent.
- **Cancelled mobile job** — `'cancelled'` would fall through all three branches and silently omit the mobile section. Acceptable for a preview environment (the workflow run itself shows the cancellation); added a note to the test plan.
- **CloudFront timing** — unintentional, fixed in c49c1c8 (see PR description).
- **Copilot's `issues: write` flag** — fixed in e03369a.
- **Resume MOBILE_ENABLED overwrite** — fixed in e03369a.

diego-artificerinnovations (approved):
Both blocking issues from my first review are resolved. Approving.

**`PR_NUMBER` unbound variable** — Fixed. `PR_NUMBER: ${{ github.event.pull_request.number }}` is now declared at the `deploy-preview-mobile` job level, so `PR_NUMBER_VALUE="${PR_NUMBER}"` inside the step references a bound variable. No more `set -u` exit under `set -euo pipefail`.

**CloudFront timing / PR comment race** — Fixed. The new `comment` job has `needs: [deploy-preview, deploy-preview-mobile]`, so it only runs after both jobs complete. The `if: always() && needs.deploy-preview.result == 'success'` condition correctly skips the comment if the web deploy never succeeded. The comment script handles all mobile outcome states (`skipped` → omit section, `failure` → surface an error line, success → full links) via `MOBILE_JOB_RESULT`.

A few notes for the record (not blockers):

- Verified `clearExpoKeysFromAcc` — already imported from `./lib/setup-expo-eas.mjs`. The diff didn't show this import because it was unchanged, but the call site is wired correctly.
- The interactive prompt for mobile opt-in (`Enable mobile…? [Y/n]`) is a nice touch — makes the setup experience match the QUICKSTART docs.
- The EXPO_ACCOUNT guard change (error out instead of silently skipping when `MOBILE_ENABLED=true` but account not set) is the right call — silent skips in production deploys are dangerous.

Good work closing the loop on the timing and shell safety issues.

ZappoMan:
@sarah-artificerinnovations - something seems wrong about this, the PR deployments message is missing the web URL, but this deployment should have both web and mobile, so something seems to have broken about deploy messages.

sarah-artificerinnovations:
Root cause found and fixed (commit f623f2c).

GitHub Actions was silently dropping the `preview-web-url` job output with the warning **"Skip output since it may contain secret"** — the preview domain value also appears in a registered secret, so the runner refused to propagate it through the cross-job output chain.

Fix: the `comment` job now reconstructs the URL directly from `vars.PR_PREVIEW_DOMAIN` and `vars.PR_PREVIEW_PREFIX` (repository variables, not secrets) plus `context.payload.pull_request.number`. This bypasses job output propagation entirely. The unused `outputs:` block on `deploy-preview` has been removed.

PR #77, opened 2026-05-12
feat(dark-mode): foundation — Tailwind config, anti-FOUC, ThemeContext, CSS cleanup

### LINKED ISSUE #75: Dark Mode is Broken
## Summary

Dark mode is applied inconsistently across the web app: some marketing sections and the authenticated product surface stay in a light appearance while the user has selected (or the system implies) dark mode. Auth and legal flows are especially misaligned.

## Business reasons

- **Trust and polish**: Mixed light/dark surfaces read as unfinished and undermine confidence in a product that handles accounts and billing.
- **Accessibility and comfort**: Users who rely on dark mode for reduced eye strain or low-light environments should not be forced into bright panels on high-intent paths (sign-in, checkout-related areas, legal disclosures).
- **Brand consistency**: The experience should feel like one product, not a patchwork of themes, especially on the home page (pricing, footer) where first impressions matter.
- **Conversion and retention**: Friction on login/signup and in-app areas (dashboard, profile, billing) increases drop-off and support questions (“why does only part of the app go dark?”).
- **Operational cost**: Theme bugs generate duplicate bug reports and one-off CSS fixes; a clear bar for “done” reduces rework.

## Current behavior (reported)

**Marketing / home**

- Most of the home page respects dark mode, but the **pricing table** and **footer** stay bright.

**Auth**

- **Login / Sign In** and **Signup**: dark mode is not applied coherently (e.g. some controls appear dark while the rest of the layout stays light).

**Legal**

- **Legal pages** do not respect dark mode.

**Authenticated app**

- After sign-in, **dashboard**, **profile**, **billing**, and related screens do not honor dark mode.

## Acceptance criteria

1. **Single source of truth**  
   When the app is in dark mode (user preference and/or system, per existing product rules), all first-party web routes use the same theme tokens (background, text, borders, inputs, cards, tables) so no screen appears “half light / half dark” unless explicitly designed that way (e.g. embedded third-party widgets documented as exceptions).

2. **Marketing home**  
   With dark mode active, the **pricing** section and **footer** use dark-appropriate backgrounds and text contrast comparable to the rest of the landing page (WCAG AA for normal text where feasible).

3. **Auth**  
   Login, sign-in, and signup flows use the same dark (or light) theme as the rest of the app for page chrome, forms, links, errors, and primary/secondary actions—no isolated dark-only controls on an otherwise light canvas.

4. **Legal**  
   All legal pages respect the active theme end-to-end (layout, typography, links).

5. **Authenticated product**  
   Dashboard, profile, billing, and other logged-in surfaces respect dark mode consistently with the marketing shell (navigation, sidebars, tables, empty states, modals).

6. **Persistence and system alignment**  
   Theme choice behaves consistently across navigation, refresh, and direct URL entry (no flash of wrong theme beyond acceptable FOUC limits already tolerated elsewhere).

7. **Verification**  
   Manual QA checklist covers: home (including pricing + footer), each auth screen, at least one legal page, and representative logged-in routes (dashboard, profile, billing). Automated tests added or updated only where the project already tests UI/theme (no new policy required beyond existing conventions).

## Notes / out of scope

- Third-party embeds or iframes that cannot be themed are out of scope unless we control their container styling; any such limitation should be called out in QA notes.
- Mobile app theming is a separate issue unless this issue is explicitly expanded to include `apps/mobile`.

diego-artificerinnovations:
Closes #75 (partial — foundation layer)

## What's in this PR

This is the first of two PRs for issue #75. It lays the foundation that makes dark mode work at all — without these changes, every `dark:` utility in the codebase is silently ignored.

### Changes

**`apps/web/tailwind.config.js`** — Added `darkMode: 'class'`
The root cause: without this, Tailwind never emits `dark:` utilities. One line, fixes the entire broken chain.

**`apps/web/index.html`** — Anti-FOUC inline script in `<head>`
Reads `localStorage['beakerstack:theme']` synchronously before React boots. Falls back to `matchMedia('prefers-color-scheme: dark')`. Wrapped in `try/catch` for CSP / private browsing safety. Prevents white flash on dark-preferring users.

**`apps/web/src/index.css`** — Removed conflicting Vite-template dark defaults
The template's `:root` set `color: rgba(255,255,255,0.87)` and `background-color: #242424` (dark always), with a `@media (prefers-color-scheme: light)` override. These fought Tailwind's `bg-white` / `text-gray-*` utilities. Removed entirely; kept `color-scheme: light dark`, fonts, and button rules.

**`apps/web/src/contexts/ThemeContext.tsx`** — New ThemeContext
Three-state preference (`system` | `light` | `dark`) with `localStorage` persistence (`beakerstack:theme` key). OS preference watcher via `matchMedia` (active only in `system` mode). `setTheme()` setter for the footer toggle component (coming in PR 2). `useTheme()` hook throws if used outside provider.

**`apps/web/src/hooks/useTheme.ts`** — Re-export for clean import paths
`import { useTheme } from '../hooks/useTheme'` works from any component.

**`apps/web/src/main.tsx`** — Wrapped app in `<ThemeProvider>`
Outermost provider, before `AuthProvider`, so the resolved theme is available everywhere.

## What's NOT in this PR

Dark-mode `dark:` variants on individual components and the `ThemeToggle` footer button are in PR 2 (`diego/issue-75-dark-mode-components`). That PR will be opened after this one merges to `develop`.

## Testing

1. Open the app with no `localStorage` entry — should match OS preference with no flash
2. Set `localStorage['beakerstack:theme'] = 'dark'` and reload — `dark` class on `<html>` immediately, no flash
3. Set to `'light'` — same, no flash even on dark-OS machines
4. Toggle OS dark mode while app is open with `system` preference — resolves live

### DIFF
--- a/apps/web/index.html
+++ b/apps/web/index.html
@@ -5,19 +5,20 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>%APP_TITLE%</title>
 
-    <!-- Site-wide meta — canonical and og:url are home-only and are injected into
-         prerender-home.html by scripts/prerender-home.ts; adding them here would
-         incorrectly set home-page values on /dashboard, /login, etc. -->
-    <meta name="description" content="BeakerStack gives you auth, billing, and a cross-platform React foundation — ready to ship your SaaS." />
-    <meta property="og:type" content="website" />
-    <meta property="og:title" content="BeakerStack — Ship your SaaS faster." />
-    <meta property="og:description" content="Everything you need to ship a real product. Auth, billing, and cross-platform React with a production CI/CD pipeline." />
-    <!-- Shipped icon until a dedicated 1200×630 og-image asset exists. -->
-    <meta property="og:image" content="https://beakerstack.com/android-chrome-512x512.png" />
-    <meta name="twitter:card" content="summary_large_image" />
-    <meta name="twitter:title" content="BeakerStack — Ship your SaaS faster." />
-    <meta name="twitter:description" content="Everything you need to ship a real product." />
-    <meta name="twitter:image" content="https://beakerstack.com/android-chrome-512x512.png" />
+    <!-- Anti-FOUC: apply dark class before first paint.
+         Reads stored user preference first; falls back to OS preference.
+         try/catch guards against CSP or private-browsing environments. -->
+    <script>
+      (function () {
+        try {
+          var stored = localStorage.getItem('beakerstack:theme');
+          var prefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
+          if (stored === 'dark' || (stored !== 'light' && prefersDark)) {
+            document.documentElement.classList.add('dark');
+          }
+        } catch (_) {}
+      })();
+    </script>
 
     <!-- Favicons for all browsers -->
     <link rel="icon" type="image/x-icon" href="/favicon.ico" />
--- a/apps/web/index.html
+++ b/apps/web/index.html
@@ -20,6 +20,20 @@
       })();
     </script>
 
+    <!-- Site-wide meta — canonical and og:url are home-only and are injected into
+         prerender-home.html by scripts/prerender-home.ts; adding them here would
+         incorrectly set home-page values on /dashboard, /login, etc. -->
+    <meta name="description" content="BeakerStack gives you auth, billing, and a cross-platform React foundation — ready to ship your SaaS." />
+    <meta property="og:type" content="website" />
+    <meta property="og:title" content="BeakerStack — Ship your SaaS faster." />
+    <meta property="og:description" content="Everything you need to ship a real product. Auth, billing, and cross-platform React with a production CI/CD pipeline." />
+    <!-- Shipped icon until a dedicated 1200×630 og-image asset exists. -->
+    <meta property="og:image" content="https://beakerstack.com/android-chrome-512x512.png" />
+    <meta name="twitter:card" content="summary_large_image" />
+    <meta name="twitter:title" content="BeakerStack — Ship your SaaS faster." />
+    <meta name="twitter:description" content="Everything you need to ship a real product." />
+    <meta name="twitter:image" content="https://beakerstack.com/android-chrome-512x512.png" />
+
     <!-- Favicons for all browsers -->
     <link rel="icon" type="image/x-icon" href="/favicon.ico" />
     <link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png" />
--- a/apps/web/src/contexts/ThemeContext.tsx
+++ b/apps/web/src/contexts/ThemeContext.tsx
@@ -46,7 +46,7 @@ function applyClass(isDark: boolean): void {
 export function ThemeProvider({ children }: { children: ReactNode }) {
   const [preference, setPreference] = useState<ThemePreference>(readStored);
   const [resolvedTheme, setResolvedTheme] = useState<ResolvedTheme>(() =>
-    resolve(readStored())
+    resolve(preference)
   );
 
   // Keep DOM in sync whenever resolved theme changes.
--- a/apps/web/src/contexts/ThemeContext.tsx
+++ b/apps/web/src/contexts/ThemeContext.tsx
@@ -20,6 +20,7 @@ interface ThemeContextValue {
 const ThemeContext = createContext<ThemeContextValue | null>(null);
 
 function getSystemDark(): boolean {
+  if (typeof window === 'undefined') return false;
   return window.matchMedia('(prefers-color-scheme: dark)').matches;
 }
 
--- a/apps/web/src/contexts/ThemeContext.tsx
+++ b/apps/web/src/contexts/ThemeContext.tsx
@@ -1,0 +1,94 @@
+import {
+  createContext,
+  useContext,
+  useEffect,
+  useState,
+  type ReactNode,
+} from 'react';
+
+export type ThemePreference = 'system' | 'light' | 'dark';
+export type ResolvedTheme = 'light' | 'dark';
+
+const THEME_KEY = 'beakerstack:theme';
+
+interface ThemeContextValue {
+  preference: ThemePreference;
+  resolvedTheme: ResolvedTheme;
+  setTheme: (theme: ThemePreference) => void;
+}
+
+const ThemeContext = createContext<ThemeContextValue | null>(null);
+
+function getSystemDark(): boolean {
+  return window.matchMedia('(prefers-color-scheme: dark)').matches;
+}
+
+function resolve(pref: ThemePreference): ResolvedTheme {
+  if (pref === 'dark') return 'dark';
+  if (pref === 'light') return 'light';
+  return getSystemDark() ? 'dark' : 'light';
+}
+
+function readStored(): ThemePreference {
+  try {
+    const v = localStorage.getItem(THEME_KEY);
+    if (v === 'dark' || v === 'light') return v;
+  } catch {
+    // private browsing / CSP
+  }
+  return 'system';
+}
+
+function applyClass(isDark: boolean): void {
+  document.documentElement.classList.toggle('dark', isDark);
+}
+
+export function ThemeProvider({ children }: { children: ReactNode }) {
+  const [preference, setPreference] = useState<ThemePreference>(readStored);
+  const [resolvedTheme, setResolvedTheme] = useState<ResolvedTheme>(() =>
+    resolve(readStored())
+  );
+
+  // Keep DOM in sync whenever resolved theme changes.
+  useEffect(() => {
+    applyClass(resolvedTheme === 'dark');
+  }, [resolvedTheme]);
+
+  // Watch OS preference while user has 'system' selected.
+  useEffect(() => {
+    if (preference !== 'system') return;
+    const mq = window.matchMedia('(prefers-color-scheme: dark)');
+    const handler = (e: MediaQueryListEvent) => {
+      const next: ResolvedTheme = e.matches ? 'dark' : 'light';
+      setResolvedTheme(next);
+    };
+    mq.addEventListener('change', handler);
+    return () => mq.removeEventListener('change', handler);
+  }, [preference]);
+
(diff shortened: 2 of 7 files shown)


apps/web/index.html
```diff
@@ -5,19 +5,20 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
```
sarah-artificerinnovations, line 5: The script was inserted *in place of* the meta tags, not alongside them. These og:* / twitter:* / description tags were removed by this diff and need to be restored.

sarah-artificerinnovations (changes requested):
The foundation is solid — correct approach throughout. One blocking issue in the diff.

---

### 🔴 Blocking — social meta tags and description dropped from `index.html`

The diff replaces the entire original block with the anti-FOUC script, removing the `<meta name="description">`, `og:*`, and `twitter:*` tags along with it. Those 13 lines aren't just comments — they're load-bearing SEO and social-preview tags.

The anti-FOUC script needs to be **added**, not substituted. The original meta tags belong before `</head>` alongside the new script (or vice versa — order doesn't matter for meta).

The removed tags:
```html
<meta name="description" content="BeakerStack gives you auth, billing, and a cross-platform React foundation — ready to ship your SaaS." />
<meta property="og:type" content="website" />
<meta property="og:title" content="BeakerStack — Ship your SaaS faster." />
<meta property="og:description" content="Everything you need to ship a real product. Auth, billing, and cross-platform React with a production CI/CD pipeline." />
<meta property="og:image" content="https://beakerstack.com/android-chrome-512x512.png" />
<meta name="twitter:card" content="summary_large_image" />
<meta name="twitter:title" content="BeakerStack — Ship your SaaS faster." />
<meta name="twitter:description" content="Everything you need to ship a real product." />
<meta name="twitter:image" content="https://beakerstack.com/android-chrome-512x512.png" />
```

---

### ✅ Everything else looks good

**Anti-FOUC logic** — The condition `stored === 'dark' || (stored !== 'light' && prefersDark)` correctly handles all four states: stored `'dark'` → dark; stored `'light'` → light; stored `'system'` or `null` → OS preference. `'system'` is never written to localStorage (correctly removed on `setTheme('system')`), so the FOUC script will always see `null` for system preference and fall back to OS.

**`ThemeContext.tsx`** — Clean design. Three-state preference with OS watcher only active in `system` mode is the right pattern. `readStored()` falling back to `'system'` on unknown values is correct. React 18 automatic batching means the two `setState` calls in `setTheme` produce a single re-render.

One minor: `resolve(readStored())` in the `resolvedTheme` initializer calls `readStored()` a second time (the `preference` initializer already called it). Not a bug — localStorage reads are synchronous and fast — but `const initial = readStored(); useState(initial); useState(() => resolve(initial))` would avoid the redundant read if you want to tighten it up.

**`index.css`** — Removing the conflicting `:root` dark defaults and the `@media (prefers-color-scheme: light)` block is the right call. Keeping `color-scheme: light dark` for browser chrome is correct.

**`main.tsx`** — `ThemeProvider` outermost, before `AuthProvider`. Correct.

**`tailwind.config.js`** — One line, unlocks every `dark:` utility. ✓

mei-artificerinnovations (approved):
Approving. The ThemeContext design is solid — the two-state split (`preference` / `resolvedTheme`), localStorage lifecycle (`system` removes the key rather than storing it), and the OS-preference watcher scoped to the `system` branch are all correct. Anti-FOUC script logic is right: `stored === 'dark'` → dark; `stored !== 'light' && prefersDark` → dark; everything else → light. Consistent with what the context writes.

One thing to flag before merging:

**`index.html` removes OG / Twitter meta tags** — The anti-FOUC script replaced 13 lines that included `og:title`, `og:description`, `og:image`, `twitter:card`, and friends. The removed comment noted these were home-page-only values that belonged in `prerender-home.html`, not the app shell. If that prerender script is in place and covers what crawlers actually hit for the home route, this is correct cleanup. But it's a non-trivial change that isn't called out in the PR description — worth confirming explicitly so future reviewers don't think it was accidental.

One minor nit (not a blocker): `readStored()` is called twice during initialization — once for `preference` state, once inside `resolve()` for `resolvedTheme`. Both read the same `localStorage` key at mount time. Could be one call: `const initial = readStored(); useState(initial); useState(() => resolve(initial))`. As-is it's just two redundant reads at startup — no correctness issue.

rahul-artificerinnovations (changes requested):
## Overall

The foundation is solid: `darkMode: 'class'` in tailwind.config.js is the right root enabler, the anti-FOUC script logic is correct, ThemeContext is well-structured with proper localStorage persistence, and wrapping ThemeProvider as the outermost provider in main.tsx is exactly right.

One blocking issue around `index.html`.

---

## Blocking: `index.html` drops OG meta tags from PR #72

PR #72 adds OG/Twitter card meta tags to `index.html`. This PR replaces the inline script section and removes them. Regardless of merge order, the final state of `develop` must have **both** the anti-FOUC script and the OG tags.

If PR #72 merges to develop before this PR, a rebase will surface a conflict on `index.html` — resolve it by keeping both. If this merges first, whoever merges PR #72 will need to ensure the tags land alongside the anti-FOUC script.

Easiest fix: add the OG tags back to `index.html` in this PR now (below the anti-FOUC script), so both changes coexist from the start and there's no coordination risk.

---

## Non-blocking: `resolve()` + SSR safety

The `resolve(readStored())` call in the `useState` initializer runs eagerly at component initialization — including during SSR pre-rendering if ThemeProvider is in the render tree. If `resolve('system')` calls `window.matchMedia` without a `typeof window === 'undefined'` guard, it will throw `ReferenceError: window is not defined` in the vite-node pre-render process.

If ThemeProvider isn't included in the SSR render (i.e., the pre-render script renders just the landing page component without the full provider tree), this is a non-issue. Worth verifying against PR #72's pre-render script. If ThemeProvider is in the tree, `resolve()` needs the guard before calling matchMedia.

---

## Everything else ✓

- `tailwind.config.js`: `darkMode: 'class'` ✓
- `ThemeContext.tsx`: Three-state preference model, proper localStorage key, try/catch for private browsing, MediaQueryList listener correctly cleaned up on preference change ✓
- `hooks/useTheme.ts`: Clean re-export barrel ✓
- `index.css`: Conflicting light/dark `:root` defaults correctly removed ✓
- `main.tsx`: ThemeProvider as outermost wrapper ✓

diego-artificerinnovations:
Fixed in commit `92e92e79`.

**What happened:** The anti-FOUC script replaced the social meta block rather than being inserted alongside it. The tags were accidentally dropped, not intentionally removed — they still belong in `index.html` as site-wide defaults.

To be explicit about the distinction Mei raised: `index.html` carries the site-wide fallbacks (`og:type`, `og:title`, `og:description`, `og:image`, `twitter:*`). The prerender script adds canonical/`og:url` to `prerender-home.html` because those are home-page-specific values that would be wrong on `/dashboard` or `/login` — the comment at the top of the meta block explains this. The two aren't mutually exclusive; both files now have their correct tags.

The fix puts the anti-FOUC script immediately after `<title>` (as early as possible, before any render-blocking work) and preserves the full meta block in its original position below it.

**ThemeContext nit** also fixed in the same commit: initial `resolvedTheme` state now uses the already-resolved `preference` value from the first `useState` rather than calling `readStored()` a second time.

sarah-artificerinnovations (approved):
Both issues addressed.

**Meta tags** — anti-FOUC script is now inserted *above* the existing meta block, not in place of it. All `og:*`, `twitter:*`, and `description` tags preserved. ✓

**`readStored` deduplication** — `resolvedTheme` initializer now uses `resolve(preference)` where `preference` is the already-computed state value from the line above. Correct — `readStored()` is called exactly once on mount. ✓

Approved.

mei-artificerinnovations (approved):
Both issues resolved. Meta tags are restored after the anti-FOUC script with the original explanatory comment. `resolve(preference)` instead of `resolve(readStored())` is the right fix — `preference` was already initialized from `readStored()` so it avoids a redundant `localStorage.getItem` at mount. Approving.

diego-artificerinnovations:
Addressing Rahul's two points:

**Blocking — OG/Twitter tags:** These are already restored. Commit `92e92e79` (pushed before Rahul's review) has the full meta block back in `index.html` in its original position, below the anti-FOUC script. The current HEAD of this branch has both. No further action needed on this point.

**Non-blocking — `window.matchMedia` SSR guard:** Fixed in commit `a2ce311d`.

```ts
function getSystemDark(): boolean {
  if (typeof window === 'undefined') return false;
  return window.matchMedia('(prefers-color-scheme: dark)').matches;
}
```

Checked against PR #72's pre-render script: `prerender-home.ts` renders `LandingPage` directly inside `MemoryRouter` — `ThemeProvider` is not in that render tree, so there's no active throw risk today. But the guard is the right defensive call; it also makes `ThemeContext` safe to unit-test in a jsdom-less environment.

rahul-artificerinnovations (approved):
Blocking issue resolved. The comment in `index.html` clarifies the architecture cleanly: OG/canonical tags are home-only metadata injected by the pre-render script into `prerender-home.html`, not global `<head>` content — putting them in `index.html` would incorrectly attach home-page metadata to `/dashboard`, `/login`, etc. That's the right call and the comment explains it for future editors.

The non-blocking SSR concern I noted earlier (`resolve()` calling `window.matchMedia` without a `typeof window === 'undefined'` guard) is still present. As long as `ThemeProvider` stays out of the pre-render tree this won't bite — just worth keeping in mind if the pre-render scope ever expands.

Everything else is solid. Approved.

diego-artificerinnovations:
**CI fix — coverage threshold** (commit b0d9e1ff)

Root cause: `ThemeContext.tsx` and `useTheme.ts` were new files with near-zero test coverage (7.93% and 0%), pulling the overall web coverage from 99%+ down to 98.04%.

Fix: added `src/contexts/__tests__/ThemeContext.test.tsx` covering the full surface — `ThemeProvider`, `resolve()`, `readStored()`, `getSystemDark()`, `setTheme()` (all three variants), DOM class application, OS preference listener, and the `useTheme()` throws-outside-provider guard. Also imports via `hooks/useTheme` to cover the re-export.

The 59 existing tests were already passing; this brings coverage back above the 99% threshold.

PR #92, opened 2026-05-12
fix(code-quality): address AI findings from issue #91

### LINKED ISSUE #91: AI findings - code quality issues - 051226
packages/billing/src/components/__tests__/PricingTable.web.test.tsx
Last push yesterday to develop
2 findings
The mock returns only two plans with minimal properties. The test on line 33 expects exactly 2 "Select" buttons (`toHaveLength(2)`), but if the component were to filter or skip plans based on additional properties (e.g., `is_public`, `stripe_product_id`), this assertion could pass silently with incomplete mock data. Consider adding more complete plan objects that match the actual shape returned by the real hook to ensure the component handles all properties correctly.

The `usePlan` mock returns a minimal object with only an `id` field. If the component uses additional properties from the current plan (such as `price_cents`, `display_name`, or `billing_period`) in the highlight logic or elsewhere, tests might not catch issues when those properties are missing or incorrect. Consider returning a more complete plan object that matches the structure of plans in the catalog mock.

packages/shared/src/components/navigation/UserMenu.native.tsx
Last push yesterday to develop
2 findings
The `menuRef` variable is declared but never used in the component. Consider removing this unused ref or implementing the functionality that requires it.

NOTE: Should we adjust our lint rules to catch this?

The comment mentions keeping the same color as web for consistency, but the color is the same as `menuItemText` (line 224). Consider either removing the `menuItemDangerText` style entirely if it's meant to be identical, or updating the comment to clarify why a separate style is needed despite having the same color value.

packages/shared/src/components/navigation/UserMenu.web.tsx
Last push 16 hours ago to develop
1 finding
The `aria-hidden` attribute should have a boolean value. Change `aria-hidden` to `aria-hidden="true"` to properly hide the decorative icon from screen readers.

scripts/setup-full.mjs
Last push 17 hours ago to develop
3 findings
The `env` property is being set from `opts.env`, but the `runInteractive` function signature at line 379 does not include an `env` parameter in the `opts` type definition. This would cause `opts.env` to be undefined unless explicitly passed, which may lead to unexpected behavior when environment variables need to be customized.

The `resolveGhRepo` function spawns a subprocess with `stdio` set to the default, which inherits all streams including stderr. This could expose sensitive error information to the console. Consider explicitly setting `stdio: ['pipe', 'pipe', 'pipe']` for better control over output handling, consistent with other command invocations in this file.

In dry-run mode at lines 1260-1261, the code assigns `acc.PREVIEW_SUPABASE_URL` to `acc.PR_TESTING_SUPABASE_URL`, but `PREVIEW_SUPABASE_URL` has not been set yet in the dry-run branch. The order should be reversed to assign these values after setting PREVIEW_SUPABASE_URL, or use the literal placeholder values directly.

sarah-artificerinnovations:
Closes #91

Addresses all 8 findings from the AI code review batch:

**`UserMenu.web.tsx`**
- Fix `aria-hidden` from attribute-only to `aria-hidden={true}` (boolean value required)

**`UserMenu.native.tsx`**
- Remove unused `menuRef` ref (declared, applied to container View, but never read)
- Remove `menuItemDangerText` from StyleSheet — it was identical to `menuItemText` (`color: '#374151'`) with only a stale comment explaining it; Sign Out now uses `menuItemText` directly

**`PricingTable.web.test.tsx`**
- Expand `usePlanCatalog` mock to full `Plan` shape (`product_id`, `description`, `stripe_price_id_monthly`, `stripe_price_id_annual`, `stripe_product_id`, `features`, `usage_limits`, `trial_period_days`, `is_public`, `display_order`)
- Expand `usePlan` mock to return the same complete object instead of `{ id }` only
- Both mocks now share `MOCK_PLANS` constant to keep them in sync

**`scripts/setup-full.mjs`**
- Add `env?: NodeJS.ProcessEnv` to `runInteractive` opts JSDoc (parameter was used but not typed)
- Pipe all stdio in `resolveGhRepo` (`['pipe', 'pipe', 'pipe']`) to prevent stderr leaking to console
- In dry-run preview block: replace `acc.PREVIEW_SUPABASE_URL` / `acc.PREVIEW_SUPABASE_ANON_KEY` reads with literal placeholder strings — the same values the preceding lines set, but more direct and consistent with the "select existing" dry-run path at line 1345

### DIFF
--- a/packages/billing/src/components/__tests__/PricingTable.web.test.tsx
+++ b/packages/billing/src/components/__tests__/PricingTable.web.test.tsx
@@ -3,28 +3,47 @@ import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { PricingTable } from '../PricingTable.web';
 
+const MOCK_PLANS = [
+  {
+    id: 'plan_free',
+    product_id: 'prod_test',
+    display_name: 'Free',
+    description: null,
+    price_cents: 0,
+    billing_period: 'free',
+    stripe_price_id_monthly: null,
+    stripe_price_id_annual: null,
+    stripe_product_id: null,
+    features: { ai_summarize: false },
+    usage_limits: { collections: 3 },
+    trial_period_days: 0,
+    is_public: true,
+    display_order: 0,
+  },
+  {
+    id: 'plan_pro',
+    product_id: 'prod_test',
+    display_name: 'Pro',
+    description: 'Full access',
+    price_cents: 1900,
+    billing_period: 'monthly',
+    stripe_price_id_monthly: 'price_pro_monthly',
+    stripe_price_id_annual: null,
+    stripe_product_id: 'prod_pro',
+    features: { ai_summarize: true },
+    usage_limits: { collections: 100 },
+    trial_period_days: 0,
+    is_public: true,
+    display_order: 1,
+  },
+];
+
 vi.mock('../../hooks/usePlanCatalog.js', () => ({
-  usePlanCatalog: () => ({
-    plans: [
-      {
-        id: 'plan_free',
-        display_name: 'Free',
-        price_cents: 0,
-        billing_period: 'free',
-      },
-      {
-        id: 'plan_pro',
-        display_name: 'Pro',
-        price_cents: 1900,
-        billing_period: 'monthly',
-      },
-    ],
-    loading: false,
-  }),
+  usePlanCatalog: () => ({ plans: MOCK_PLANS, loading: false, error: null }),
 }));
 
 vi.mock('../../hooks/usePlan.js', () => ({
-  usePlan: () => ({ data: { id: 'plan_pro' } }),
+  usePlan: () => ({ data: MOCK_PLANS[1] }),
 }));
 
 describe('PricingTable', () => {
--- a/packages/shared/src/components/navigation/UserMenu.native.tsx
+++ b/packages/shared/src/components/navigation/UserMenu.native.tsx
@@ -38,7 +38,6 @@ export interface UserMenuProps {
  */
 export function UserMenu({ user, profile, navigation }: UserMenuProps) {
   const [isOpen, setIsOpen] = useState(false);
-  const menuRef = useRef<View>(null);
   const avatarRef = useRef<View>(null);
   const [avatarLayout, setAvatarLayout] = useState({
     x: 0,
@@ -87,7 +86,7 @@ export function UserMenu({ user, profile, navigation }: UserMenuProps) {
 
   return (
     <>
-      <View style={styles.container} ref={menuRef}>
+      <View style={styles.container}>
         <View ref={avatarRef} collapsable={false}>
           <TouchableOpacity
             accessibilityLabel='Open user menu'
@@ -157,9 +156,7 @@ export function UserMenu({ user, profile, navigation }: UserMenuProps) {
                 style={[styles.menuItem, styles.menuItemDanger]}
                 activeOpacity={0.7}
               >
-                <Text style={[styles.menuItemText, styles.menuItemDangerText]}>
-                  Sign Out
-                </Text>
+                <Text style={styles.menuItemText}>Sign Out</Text>
               </TouchableOpacity>
             </View>
           </View>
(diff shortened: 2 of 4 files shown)


rahul-artificerinnovations (approved):
All 8 findings correctly addressed. Approving.

**`UserMenu.web.tsx`** — `aria-hidden={true}` is the right fix. Explicit boolean is cleaner for TypeScript and unambiguous for screen reader tooling. ✓

**`UserMenu.native.tsx`** — `menuRef` removal is clean: `avatarRef` still uses `useRef` so the import stays valid. `menuItemDangerText` removal is correct — it was identical to `menuItemText` and the Sign Out button's container style (`menuItemDanger`) is unchanged, so no visual regression. ✓

**`PricingTable.web.test.tsx`** — Shared `MOCK_PLANS` constant is the right pattern: both mocks stay in sync automatically if the `Plan` type changes, and the `error: null` addition matches the real hook shape. ✓

**`scripts/setup-full.mjs`** — All three changes are correct:
- JSDoc `env?` addition documents what the code already used. ✓
- `stdio: ['pipe', 'pipe', 'pipe']` on `resolveGhRepo` is consistent with other spawns in the file and prevents stderr leaking on `gh` errors. ✓
- Replacing `acc.PREVIEW_SUPABASE_URL` / `acc.PREVIEW_SUPABASE_ANON_KEY` with literal placeholders fixes the read-before-set bug — those keys were never populated in the dry-run branch. ✓

---

**One follow-up worth tracking separately:** Issue #91 flagged the unused `menuRef` ref and asked whether lint rules should catch this. `no-unused-vars` / TypeScript `noUnusedLocals` would have caught it at the editor. Worth a small issue or a lint config tweak — not for this PR, but it's a real gap if AI batch review is how the team is finding unused variables.

mei-artificerinnovations (approved):
Clean batch — all eight findings handled correctly. A couple of observations:

**`scripts/setup-full.mjs` dry-run fix** is the most substantive change here. `acc.PREVIEW_SUPABASE_URL` and `acc.PREVIEW_SUPABASE_ANON_KEY` are never set in the dry-run branch of the Supabase phase, so the original code was silently assigning `undefined` to `PR_TESTING_SUPABASE_URL` / `PR_TESTING_SUPABASE_ANON_KEY`. The literal placeholders are the right fix, and they're consistent with the "select existing" dry-run path.

**`UserMenu.native.tsx` — `menuItemDanger` still on the TouchableOpacity** (container style, not the Text). That's intentional and correct — the TouchableOpacity keeps its touch-target/opacity styling via `menuItemDanger`, only the text override was dead. Worth noting in case a future reader wonders why the Text no longer explicitly references it.

**`PricingTable.web.test.tsx` mock expansion** — sharing via `MOCK_PLANS` and aligning `usePlan` to return the full object is the right call. Tests that previously only checked `data.id` still pass; tests that check additional plan fields now have realistic data to work with.

Everything looks good.

diego-artificerinnovations (approved):
Clean batch of fixes — all 8 findings look correctly addressed.

A few notes while reading:

**`UserMenu.web.tsx`** — `aria-hidden={true}` is more explicit but functionally the same as the bare attribute in JSX (both render `aria-hidden="true"` in the DOM). Valid cleanup regardless.

**`UserMenu.native.tsx`** — Removing `menuItemDangerText` is the right call; Sign Out having the same color as the other items with just the `menuItemDanger` border style for visual distinction is cleaner. The `useRef` import is still needed for `avatarRef`, so that's fine.

**`PricingTable.web.test.tsx`** — Sharing `MOCK_PLANS` between the two mocks is the correct structural fix. Previously if the `Plan` type gained a required field, one mock could diverge from the other silently.

**`setup-full.mjs`**:
- `stdio: ['pipe', 'pipe', 'pipe']` on `resolveGhRepo` is the right fix — `spawnSync` without explicit stdio inherits the parent's stderr, so `gh` errors would leak to the user's console even on silent failure paths.
- One quick thing to verify: the literal values `'https://dry-run.supabase.co'` and `'dry-run-anon'` for `PR_TESTING_*` — are these the same strings the preceding block sets for `PREVIEW_SUPABASE_URL` / `PREVIEW_SUPABASE_ANON_KEY`? The PR description says yes, but worth a glance to confirm the dry-run output is consistent.

Approved — happy for Brad to merge.

PR #103, opened 2026-05-12
fix: dark mode for auth, billing, dashboard, and profile components

rahul-artificerinnovations:
## Summary

- Adds `dark:` Tailwind variants to 18 components that had no dark mode support: sign-up plan summary, social login button, billing tabs/badges/feature rows/confirm modal, all dashboard demo sections, and shared profile header/stats/editor.
- Converts `MeteredUsageDemo` progress bar from hardcoded inline `style` hex colours to Tailwind classes so dark mode works correctly.

## Files changed (18)
`apps/web`: SignupPlanSummary, SocialLoginButton, BillingTabs, ConstraintWarning, PlanFeatureList, PlanFeatureRow, StatusBadge, FeatureLimitRow, ConfirmDowngradeModal, DashboardDemoSection, DemoControlsPanel, MeteredUsageDemo, BooleanGatesDemo, NumericCapsDemo, AISummarizeResult  
`packages/shared`: ProfileHeader, ProfileStats, ProfileEditor

## Test plan
- [ ] Toggle dark mode on `/dashboard` — all demo sections render correctly
- [ ] Toggle dark mode on `/billing` — tabs, badges, feature rows, confirm-downgrade modal render correctly
- [ ] Toggle dark mode on `/signup?plan=...` and `/login?plan=...` — plan summary card renders correctly
- [ ] Toggle dark mode on profile page — header, stats, editor render correctly

Closes #93

PR #109, opened 2026-05-13
fix(dark-mode): add dark variants to ProtectedRoute loading state (fixes #106)

### LINKED ISSUE #106: White flash (FOUC) on logout / redirect to home in dark mode
## Summary

When logging out in dark mode, there is a brief white flash before the home page renders in dark. This is a classic FOUC (Flash of Unstyled Content) caused by the dark-mode class not being applied to `<html>` before the first paint after the auth redirect.

## Reproduction

1. Enable dark mode
2. Log out
3. Observe a white flash before the page settles into dark mode

## Root cause

The Tailwind `dark` class (or `dark` media query) needs to be present on `<html>` before React hydrates. During a redirect-driven logout the browser navigates to a new page, the HTML is painted without the dark class, and only after the React/JS bundle runs does dark mode get re-applied.

## Fix candidates

- Inline a tiny `<script>` in the `<head>` (before any CSS) that reads `localStorage` and synchronously adds the `dark` class to `<html>`. This is the standard approach and runs before the first paint, eliminating the flash.
- Alternatively, use `color-scheme` CSS property with `prefers-color-scheme` to let the OS default apply immediately, removing the mismatch window.

## Notes

- Separate from PR #103 (dark mode contrast fixes). Observed as a side effect while testing that branch.
- Not a contrast issue — a timing issue.

rahul-artificerinnovations:
Fixes #106

## Root cause

`ProtectedRoute.web.tsx`'s loading state had `bg-gray-50` with no `dark:` counterpart. When `signOut()` is called, `auth.loading` flips to `true` while Supabase's async signOut completes. During that window the loading div covers the full viewport in light gray — white flash in dark mode.

The inline anti-FOUC script in `index.html` keeps the `dark` class on `<html>` throughout, but Tailwind can only apply `dark:bg-gray-900` if the class is declared in the source. It wasn't.

## Change

`packages/shared/src/components/auth/ProtectedRoute.web.tsx`

```diff
- <div className='min-h-screen flex items-center justify-center bg-gray-50'>
+ <div className='min-h-screen flex items-center justify-center bg-gray-50 dark:bg-gray-900'>
    <div className='text-center'>
      <div className='inline-block animate-spin rounded-full h-8 w-8 border-b-2 border-primary-600'></div>
-     <p className='mt-4 text-gray-600'>Loading...</p>
+     <p className='mt-4 text-gray-600 dark:text-gray-400'>Loading...</p>
    </div>
  </div>
```

## Test plan

1. Enable dark mode
2. Sign out
3. Confirm no white flash — transition stays dark throughout
4. Repeat with system theme (OS dark) to confirm `prefers-color-scheme` fallback path is also clean
5. Confirm loading spinner is readable in dark mode

### DIFF
--- a/packages/shared/src/components/auth/ProtectedRoute.web.tsx
+++ b/packages/shared/src/components/auth/ProtectedRoute.web.tsx
@@ -20,10 +20,10 @@ export function ProtectedRoute({
   // Show loading state while checking authentication
   if (auth.loading) {
     return (
-      <div className='min-h-screen flex items-center justify-center bg-gray-50'>
+      <div className='min-h-screen flex items-center justify-center bg-gray-50 dark:bg-gray-900'>
         <div className='text-center'>
           <div className='inline-block animate-spin rounded-full h-8 w-8 border-b-2 border-primary-600'></div>
-          <p className='mt-4 text-gray-600'>Loading...</p>
+          <p className='mt-4 text-gray-600 dark:text-gray-400'>Loading...</p>
         </div>
       </div>
     );


sarah-artificerinnovations (approved):
This is the right fix and it's complete.

The anti-FOUC script in `index.html` was doing its job — `dark` stays on `<html>` throughout the logout. The flash was the `ProtectedRoute` loading overlay covering the full viewport with `bg-gray-50` during the async `signOut()` window, before the redirect fires. Tailwind can only apply `dark:bg-gray-900` if it's declared; it wasn't.

The two-line change is precisely scoped and correct. The test plan covers both the stored-preference path and the `prefers-color-scheme` fallback, which is the right coverage.

One thing worth noting for future dark mode audits: this loading state lives in `packages/shared/`, which means it's shared across all web consumers of `ProtectedRoute`. The PR #103 dark mode pass caught most components but missed this one because it's a loading/auth-gate path that doesn't appear during normal navigation. Worth adding a note in the static analysis test (`dark-mode-coverage.test.ts`) to cover full-screen loading overlays specifically — but that's a separate issue, not a blocker here.

**✅ Approved.**

diego-artificerinnovations (approved):
Fix is correct — approving, with notes on Sarah's latest hypotheses.

**On the change itself:** `dark:bg-gray-900` and `dark:text-gray-400` are the right additions. The `ProtectedRoute` loading div was definitively missing dark variants — regardless of what the original reporter saw, this is a real bug and the fix is necessary. Spinner color (`border-primary-600` = blue-600, `#2563eb`) is fully visible on `bg-gray-900`, so Mei's concern there is clear.

**On Sarah's external-redirect hypothesis:** worth checking, but I don't think it applies to the primary logout path. `supabaseClient.auth.signOut({ scope: 'local' })` clears the session client-side only — there's no external redirect through Supabase's servers. The `auth.loading` window is a real async gap regardless.

**On Sarah's ThemeContext-hydration hypothesis:** that would cause flashes on *all* page loads, not specifically on logout. If the flash persists after this lands and is isolated to logout, the next thing to check is whether `ThemeContext` has any async state path that temporarily removes the `dark` class during re-mount — but that's a follow-up investigation, not a blocker here.

Merge this; if the logout flash persists, reopen #106 with a note that `ProtectedRoute` is fixed and the ThemeContext init path is next.

mei-artificerinnovations (approved):
**Approved — fix is correct, Sarah's redirect concern doesn't apply here.**

I read the `useAuth.ts` source to settle Sarah's hypothesis before approving.

**On the external-redirect question:** `signOut()` calls `supabaseClient.auth.signOut({ scope: 'local' })` — scope `'local'` clears the session entirely client-side without navigating the browser to any external URL. The inline anti-FOUC script is therefore irrelevant to this path; there's no page reload, no external hop, no second `index.html` execution. Sarah's two hypotheses (external Supabase/OAuth redirect, ThemeContext hydration toggle) don't apply to this logout flow.

**What actually happens:**
1. `signOut()` calls `setLoading(true)` synchronously
2. `ProtectedRoute` renders the full-screen loading div
3. `dark` is on `<html>` (inline script set it on load and never removed it), but `dark:bg-gray-900` is not in the div's class list — Tailwind has no rule to apply
4. White flash until `onAuthStateChange` fires and `loading` returns to `false`

Rahul's diagnosis is correct, Brad's confirmation is consistent with it, and the fix is exactly right.

**Spinner concern resolved:** `border-primary-600` is `#2563eb` — medium blue, visually clear against `gray-900`. No issue there.

The two-liner is correct. Ship it.

Issue #127, opened 2026-05-13
feat(mobile+web): React ErrorBoundary wrappers, graceful fallback UI, and Sentry/Bugsnag integration hook


rahul-artificerinnovations:
## Background

BeakerStack currently has no structured error handling beyond React Native's default red-screen in development and a blank white screen in production. Unhandled JavaScript errors crash the entire app with no recovery path and no visibility into what went wrong. This issue adds `ErrorBoundary` wrappers, polished fallback UIs, and a pluggable reporting adapter interface. It is a **prerequisite for #117** (Sentry integration), which implements the concrete `SentryAdapter` on top of this foundation.

## Business case

- Production crashes without reporting leave teams flying blind.
- A pluggable adapter lets each template consumer bring their own error reporting provider without forking the error-handling logic.
- Graceful fallback screens with retry options keep users in the app instead of forcing a full restart.
- Wrapping at multiple granularities (screen-level, root) means a crash in one area doesn't wipe out the whole session.

### Location: `packages/shared/src/error/`

Not a new workspace package — added to the existing `packages/shared` package which is already imported by both apps. Files:

- **`ErrorBoundary.tsx`** — class component. Props: `fallback?: ReactNode | ((error, reset) => ReactNode)`, `onError?: (error, info) => void`, `level?: 'screen' | 'root'`. Calls `onError` in `componentDidCatch`.
- **`FallbackUI.tsx`** — default fallback component. At **screen level**: shows icon, message, "Try again" (`reset()`), and "Go back" (navigation back). At **root level**: shows icon, message, and "Try again" only — no navigation call, since no `NavigationContainer` context is available at root.
- **`reporter.ts`** — singleton. `init(adapter: ErrorReporter)` registers the adapter. `captureException(error, context)` calls it if registered; no-ops otherwise.
- **`adapters/ConsoleAdapter.ts`** — default dev adapter; logs to `console.error` with structured format. No third-party SDK dependencies.
- **`adapters/index.ts`** — exports `ErrorReporter` interface for consumers implementing custom adapters (e.g. Sentry, Bugsnag).

### Mobile integration

- Wrap the navigator root in `App.tsx` with `<ErrorBoundary level="root">`
- Wrap each `Screen` component in `AppNavigator.tsx` with `<ErrorBoundary level="screen">` — navigator-level wrapping guarantees coverage without per-file opt-in
[...]

Issue #117, opened 2026-05-13
feat: Sentry error tracking — web, mobile, and edge functions


sarah-artificerinnovations:
## Business case

Production bugs are invisible without error tracking. When a user hits a crash or an unhandled promise rejection, the team has no signal unless the user reports it. Mean time to detect (MTTD) is measured in days, not minutes.

Sentry is the industry standard for React and React Native error tracking. It provides:
- **Automatic capture** of unhandled errors and promise rejections
- **Source-mapped stack traces** that point to original TypeScript source, not minified output
- **Session replay** (web) — a video-like recording of user interactions leading up to a crash
- **Performance monitoring** — Web Vitals, slow transactions, N+1 query detection
- **Release tracking** — errors tagged by deploy SHA so you know exactly which release introduced a bug

BeakerStack has no error visibility today. Every team building on it has to add Sentry from scratch. It should ship as part of the production stack.

## Scope

Three integration points:
1. **Web** (`apps/web`) — `@sentry/react` with Vite source map upload
2. **Mobile** (`apps/mobile`) — `@sentry/react-native` with the official Expo plugin (`@sentry/react-native/expo`)
3. **Supabase edge functions** — `@sentry/deno` in `stripe-webhook` and `billing-stripe`

---

### Configuration
- [ ] `SENTRY_DSN` added to `.env.example` (shared across web and mobile builds)
- [ ] `SENTRY_AUTH_TOKEN` added to `.env.example` (for source map upload — CI only, not shipped to client)
- [ ] `SENTRY_ORG` and `SENTRY_PROJECT` added to `.env.example`
- [ ] All Sentry calls are no-ops when `SENTRY_DSN` is unset — teams that have not configured Sentry see no errors or warnings

### Web (`apps/web`)
- [ ] `@sentry/react` and `@sentry/vite-plugin` installed
- [ ] `Sentry.init()` called in `apps/web/src/main.tsx` with `dsn`, `release` (from `VITE_APP_VERSION` or git SHA), `environment` (`VITE_APP_ENV`), and `tracesSampleRate: 0.1` (10% of transactions)
- [ ] `Sentry.reactRouterV6BrowserTracingIntegration` (or equivalent) wired to the React Router instance for route-change spans
- [ ] `@sentry/vite-plugin` configured in `vite.config.ts` to upload source maps on production builds; source maps are deleted from the build output after upload so they are not served publicly
[...]

Issue #116, opened 2026-05-13
feat(mobile): replace header dropdown with bottom tab bar navigation


mei-artificerinnovations:
## Context

The mobile app currently uses a top-right avatar dropdown (`UserMenu.native.tsx` → `Modal`) to navigate between Dashboard, Profile, and Billing. This is a direct port of the web navigation pattern and does not follow mobile UX conventions.

**Current structure (`AppNavigator.tsx`):**
- Flat `NativeStackNavigator` with screens: Home, Login, Signup, Dashboard, Profile, Billing
- Authentication state handled by `HomeScreen` routing via `navigation.navigate('Dashboard')`
- Navigation exposed only via avatar tap → modal dropdown in the top-right corner

## Problem

The header dropdown pattern creates three compounding UX problems:

1. **Reachability**: Top-right corner is the hardest zone to reach on modern large-screen phones. Apple HIG and Material Design 3 both designate the bottom ~40% of the screen as the thumb-reachable zone for primary actions.
2. **Discoverability**: Features hidden behind a tap are used 40–50% less than persistently-visible navigation (Nielsen Norman Group). Users cannot tell what the app contains without opening the menu.
3. **Expectation mismatch**: Every major B2C SaaS mobile app (Slack, Linear, Notion, Stripe, Intercom) uses bottom tab bars for primary navigation. Users arrive with an established mental model; the dropdown surprises and slows them.

**Supporting data:**
- Redbooth: +65% DAU, +70% session time after switching from dropdown to tabs
- Bottom nav: ~58% engagement lift over hidden menus; ~20% better retention
- A/B studies: 9% higher click-through and 30% more menu-item clicks when navigation is persistently visible

## Recommendation

Replace the header dropdown with a **bottom tab bar** (`@react-navigation/bottom-tabs`) for authenticated users, using 3–4 primary destinations. Secondary/infrequent actions (sign out, account deletion, etc.) move into the Profile tab's settings section.

**Proposed authenticated tab structure:**

| Tab | Icon | Screen |
|-----|------|--------|
| Home | house | `DashboardScreen` |
| Billing | credit-card | `BillingScreen` |
| Profile | person-circle | `ProfileScreen` (settings + sign out) |

Three tabs keeps cognitive load minimal and all destinations reachable by thumb. A fourth tab (e.g., Help or Notifications) can be added when that feature exists — Apple HIG recommends stopping at 5.

[...]

PR #135, opened 2026-05-13
feat(shared): ErrorBoundary, FallbackUI, crash reporter (#127)

rahul-artificerinnovations:
## Summary

- Implements issue #127: error boundary and crash reporting infrastructure for the shared package
- Adds `ErrorBoundary` class component (required — function components cannot be error boundaries) with configurable `fallback`, `onError`, and `level` props
- Adds `FallbackUI` with `screen`/`root` level variants — root level never calls `useNavigation()` (no NavigationContainer context), screen level uses an inner `ScreenFallback` component to safely call the hook
- Adds `reporter` singleton: `init(adapter)` + `captureException()` — safe no-op if `init` was never called
- Adds `ConsoleAdapter` using `console.warn` (not `console.error`) to avoid triggering React Native's red screen overlay
- Wires `ErrorBoundary level="root"` at the app root in both mobile (`App.tsx`) and web (`main.tsx`)
- Wraps all navigation screens with `ErrorBoundary level="screen"` via `withScreenBoundary` HOC in `AppNavigator.tsx`
- Wraps all route elements with `ErrorBoundary level="screen"` in web `App.tsx`
- Exports everything from `@beakerstack/shared` index
- Adds `EXPO_PUBLIC_ERROR_REPORTER` / `VITE_ERROR_REPORTER` env vars to `.env.example` (Sentry integration deferred to #117)

## Test plan

- [ ] Trigger a crash in a screen component; verify `FallbackUI` renders with "Try again" and "Go back" buttons
- [ ] "Try again" resets the error boundary and re-renders the screen
- [ ] "Go back" calls `navigation.goBack()` correctly
- [ ] Trigger a crash at root level (e.g. in `AuthProvider`); verify root `FallbackUI` renders with only "Try again" (no "Go back", no `useNavigation` call)
- [ ] Verify `reporter.captureException` is a no-op before `init()` is called (no crash)
- [ ] `ConsoleAdapter` logs to `console.warn` with timestamp, error, and context
- [ ] TypeScript compiles without errors in both `apps/mobile` and `apps/web`
- [ ] Web app: crash in a route element shows `FallbackUI` without crashing the whole app

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,3 @@
+# Error reporting adapter (optional). Set to 'sentry' to enable Sentry (requires #117).
+EXPO_PUBLIC_ERROR_REPORTER=
+VITE_ERROR_REPORTER=
--- a/apps/mobile/App.tsx
+++ b/apps/mobile/App.tsx
@@ -7,8 +7,11 @@ import { ProfileProvider } from '@beakerstack/shared/contexts/ProfileContext';
 // Import from native-specific file for correct types
 import { configureGoogleSignIn } from '@beakerstack/shared/hooks/useAuth.native';
 import { Logger } from '@beakerstack/shared/utils/logger';
+import { ErrorBoundary, reporter, ConsoleAdapter } from '@beakerstack/shared';
 import { supabase } from './src/lib/supabase';
 import { AppNavigator } from './src/navigation/AppNavigator';
+
+reporter.init(new ConsoleAdapter());
 
 export default function App() {
   useEffect(() => {
@@ -103,11 +106,13 @@ export default function App() {
   }, []);
 
   return (
-    <AuthProvider supabaseClient={supabase}>
-      <ProfileProvider supabaseClient={supabase}>
-        <AppNavigator />
-        <StatusBar style='auto' />
-      </ProfileProvider>
-    </AuthProvider>
+    <ErrorBoundary level='root'>
+      <AuthProvider supabaseClient={supabase}>
+        <ProfileProvider supabaseClient={supabase}>
+          <AppNavigator />
+          <StatusBar style='auto' />
+        </ProfileProvider>
+      </AuthProvider>
+    </ErrorBoundary>
   );
 }
--- a/apps/mobile/src/navigation/AppNavigator.tsx
+++ b/apps/mobile/src/navigation/AppNavigator.tsx
@@ -4,6 +4,7 @@ import {
   type NavigationContainerRef,
 } from '@react-navigation/native';
 import { createNativeStackNavigator } from '@react-navigation/native-stack';
+import { ErrorBoundary } from '@beakerstack/shared';
 import HomeScreen from '../screens/HomeScreen';
 import LoginScreen from '../screens/LoginScreen';
 import SignupScreen from '../screens/SignupScreen';
@@ -22,6 +23,16 @@ type RootStackParamList = {
 };
 
 const Stack = createNativeStackNavigator<RootStackParamList>();
+
+function withScreenBoundary(Component: React.ComponentType<any>) {
+  return function BoundedScreen(props: any) {
+    return (
+      <ErrorBoundary level="screen">
+        <Component {...props} />
+      </ErrorBoundary>
+    );
+  };
+}
 
 export const AppNavigator = () => {
   const navigationRef =
@@ -48,16 +59,16 @@ export const AppNavigator = () => {
       >
         <Stack.Screen
           name='Home'
-          component={HomeScreen}
+          component={withScreenBoundary(HomeScreen)}
           options={{ headerShown: false }} // Home always uses custom header
         />
-        <Stack.Screen name='Login' component={LoginScreen} />
-        <Stack.Screen name='Signup' component={SignupScreen} />
-        <Stack.Screen name='Dashboard' component={DashboardScreen} />
-        <Stack.Screen name='Profile' component={ProfileScreen} />
+        <Stack.Screen name='Login' component={withScreenBoundary(LoginScreen)} />
+        <Stack.Screen name='Signup' component={withScreenBoundary(SignupScreen)} />
+        <Stack.Screen name='Dashboard' component={withScreenBoundary(DashboardScreen)} />
+        <Stack.Screen name='Profile' component={withScreenBoundary(ProfileScreen)} />
         <Stack.Screen
           name='Billing'
-          component={BillingScreen}
+          component={withScreenBoundary(BillingScreen)}
           options={{ title: 'Billing' }}
         />
       </Stack.Navigator>
--- a/apps/web/src/App.tsx
+++ b/apps/web/src/App.tsx
@@ -3,6 +3,7 @@ import { Routes, Route, Outlet } from 'react-router-dom';
 import { ProtectedRoute } from '@beakerstack/shared/components/auth/ProtectedRoute.web';
 import { BillingProviderLayout } from './billing/BillingProviderLayout';
 import { AppFooter } from './components/AppFooter';
+import { ErrorBoundary } from '@beakerstack/shared';
 import LoginPage from './pages/LoginPage';
 import SignupPage from './pages/SignupPage';
 import DashboardPage from './pages/DashboardPage';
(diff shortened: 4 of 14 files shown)


sarah-artificerinnovations (changes requested):
Good structure overall — reporter singleton pattern, `getDerivedStateFromError` + `componentDidCatch` split, `ConsoleAdapter` using `console.warn` (correct, avoids RN red screen), and the mobile wiring in `App.tsx` all look right. One blocking issue and a few minor points.

---

### Blocking: `FallbackUI.tsx` imports React Native on web

`packages/shared/src/error/FallbackUI.tsx` imports `View`, `Text`, `TouchableOpacity`, `StyleSheet` from `react-native`, and the `ScreenFallback` inner component calls `useNavigation()` from `@react-navigation/native`. This file has no `.native.tsx` suffix, so the web app's `ErrorBoundary` (wired in `apps/web/src/main.tsx` and `App.tsx`) will import it directly. On web that's either a bundle error (no `react-native` resolution) or a runtime crash (`useNavigation()` has no React Navigation context in a React Router tree).

Fix: split into `FallbackUI.native.tsx` (current content) and `FallbackUI.web.tsx` (HTML `<div>`/`<button>` with `window.location.href = '/'` for the "Go back" button, or just omit it at root level). Metro and Vite will each pick the right file.

---

### Minor (non-blocking)

**Web reporter not initialized** — `apps/web/src/main.tsx` doesn't call `reporter.init()`. Safe no-op so no crash, but errors go completely unlogged on web until #117 (Sentry) lands. Intentional?

**`withScreenBoundary` inline in `AppNavigator.tsx`** — works fine, but since the shared package already exports `ErrorBoundary`, you could export `withScreenBoundary` from `packages/shared/src/error` too so the web app (or future apps) can reuse it instead of copy-pasting the HOC. Up to you.

**Web `App.tsx` verbosity** — wrapping every `<Route>` element inline with `<ErrorBoundary level='screen'>` is repetitive; a `BoundedRoute` or `withRoute` helper could tighten it. Minor.

mei-artificerinnovations (changes requested):
Everything in the core implementation is correct — class component, `getDerivedStateFromError`, `console.warn` in `ConsoleAdapter`, no-op reporter before `init`, `ScreenFallback` inner component to safely call `useNavigation`. One blocking issue before I can approve:

---

**`FallbackUI.tsx` imports from `react-native` — breaks the web build**

`packages/shared/src/error/FallbackUI.tsx` imports `View`, `Text`, `TouchableOpacity` from `react-native`. The web app (`apps/web`) has no `react-native-web` and no RN alias in `vite.config.ts` — Vite will fail to resolve `react-native` at build time, and any error caught by a `level="screen"` boundary in the web router will throw a module error when trying to render the fallback.

The fix follows the existing pattern in `packages/shared` for platform-specific components: split into platform files so Metro and Vite each resolve the right one.

```
packages/shared/src/error/FallbackUI.native.tsx  ← current content (unchanged)
packages/shared/src/error/FallbackUI.web.tsx     ← new, uses div/button/p + useNavigate(-1) for "Go back"
```

`ErrorBoundary.tsx` keeps its import as `import { FallbackUI } from './FallbackUI'` — Metro resolves `.native.tsx`, Vite resolves `.web.tsx`. This is exactly how `FormInput`, `ProfileAvatar`, etc. are already handled in this package.

The web `FallbackUI` screen-level fallback can use `useNavigate` from `react-router-dom` for "Go back":

```tsx
function ScreenFallback({ error, reset }: { error: Error; reset: () => void }) {
  const navigate = useNavigate();
  return (
    <div>
      ...
      <button onClick={() => navigate(-1)}>Go back</button>
    </div>
  );
}
```

---

**Non-blocking note — `AppNavigator.tsx` will conflict with #116**

This PR's changes to `AppNavigator.tsx` (adding `withScreenBoundary`) will conflict with #116's full rewrite of that file. Not asking you to hold — I'll carry `withScreenBoundary` forward when I implement #116, and we can resolve the conflict at that point. Just flagging so it's not a surprise.

---

Fix the `FallbackUI` platform split and I'll approve.

diego-artificerinnovations (changes requested):
**Changes requested — one blocking issue on web, everything else is solid.**

---

## Blocking: `FallbackUI` breaks screen-level isolation on web

`ScreenFallback` calls `useNavigation()` from `@react-navigation/native`. On web the app uses React Router, not a `NavigationContainer`, so `useNavigation()` throws `"Couldn't find a navigation object"`.

The failure chain when a web screen crashes:
1. Route component throws → screen `ErrorBoundary` catches it → renders `FallbackUI level="screen"` → renders `ScreenFallback` → `useNavigation()` throws
2. Error during the boundary's own render propagates **up**, not down — the screen boundary can't catch its own fallback's error
3. Root `ErrorBoundary` catches it → entire app shows the root fallback ("Try again" only)

Result: every screen-level crash on web collapses to a full-app crash. Screen isolation is broken.

**Fix options (pick one):**

Option A — platform-specific files (consistent with the existing shared package pattern):
- Rename current `FallbackUI.tsx` → `FallbackUI.native.tsx` (keep as-is)
- Add `FallbackUI.web.tsx` that uses `useNavigate()` from react-router-dom, or simply omits "Go back"

Option B — pass a `goBack` callback prop:
```tsx
// ErrorBoundary renders:
<FallbackUI level={level} error={this.state.error!} reset={this.reset} onGoBack={this.props.onGoBack} />

// ScreenFallback uses onGoBack instead of useNavigation()
// Mobile withScreenBoundary wraps with onGoBack={() => navigation.goBack()}
// Web ErrorBoundary wrappers pass onGoBack={() => navigate(-1)} or omit it
```

Option A is cleaner given the existing `.native.tsx` / `.web.tsx` pattern in the shared package.

---

## Everything else is correct

- Class component — required, function components cannot be error boundaries ✓
- `getDerivedStateFromError` sets state; `componentDidCatch` calls `reporter` — correct separation (no side effects in `getDerivedStateFromError`) ✓
- `reporter` singleton with optional chaining — safe no-op before `init()` ✓
- `ConsoleAdapter` uses `console.warn` — avoids RN red screen overlay ✓
- `reporter.init(new ConsoleAdapter())` at module top level in `App.tsx` — before any render ✓
- `withScreenBoundary` returns named `BoundedScreen` — visible in React DevTools ✓
- Root `ErrorBoundary` inside `React.StrictMode` on web — correct ordering ✓
- `fallback` prop supporting both `ReactNode` and render function — clean API ✓

mei-artificerinnovations (approved):
FallbackUI.native.tsx and FallbackUI.web.tsx — correct platform split. web version uses useNavigate(-1) for Go back, minHeight 100vh on root fallback is a nice touch. Approved.

diego-artificerinnovations (approved):
Block resolved. `FallbackUI.native.tsx` keeps the React Native / `useNavigation()` implementation; `FallbackUI.web.tsx` uses HTML elements + `useNavigate(-1)` from react-router-dom. Metro resolves `.native.tsx`, Vite resolves `.web.tsx` — exactly the right pattern for this package. Root-level web variant with `minHeight: '100vh'` is a nice touch. Approved.

sarah-artificerinnovations (approved):
Platform split looks correct. `FallbackUI.native.tsx` keeps RN + `useNavigation()` inside `ScreenFallback` (never called at root level). `FallbackUI.web.tsx` uses `useNavigate()` only in `ScreenFallbackWeb` (also never called at root level), so the root `ErrorBoundary` in `main.tsx` — which sits outside `<BrowserRouter>` — renders the root variant safely without touching the router context. `FallbackUI.tsx` base is a correct fallback since Vite always prefers `.web.tsx`. 

One carry-over non-blocking note: web `main.tsx` still doesn't call `reporter.init()`. Safe no-op, but errors go unlogged on web until #117 lands — presumably intentional.

Approved.

mei-artificerinnovations:
CI is failing on **Type check** with four errors:

```
src/error/ErrorBoundary.tsx(28,3): error TS4114: This member must have an 'override' modifier because it overrides a member in the base class 'Component<Props, State, any>'.
src/error/ErrorBoundary.tsx(37,3): error TS4114: This member must have an 'override' modifier because it overrides a member in the base class 'Component<Props, State, any>'.
src/error/FallbackUI.native.tsx(1,1): error TS6133: 'React' is declared but its value is never read.
src/error/FallbackUI.tsx(1,1): error TS6133: 'React' is declared but its value is never read.
```

Three fixes needed:

**1 & 2 — `ErrorBoundary.tsx` lines 28 and 37** — add `override` to both lifecycle methods (tsconfig has `noImplicitOverride: true`):
```ts
override componentDidCatch(error: Error, info: React.ErrorInfo) { ... }
override render() { ... }
```

**3 & 4 — `FallbackUI.native.tsx` and `FallbackUI.tsx` line 1** — remove `import React from 'react'`. With the `react-jsx` transform it's unused, and `noUnusedLocals` is on.

Issue #125, opened 2026-05-13
feat(mobile): Expo push notifications with Supabase Edge Function delivery and in-app preferences


rahul-artificerinnovations:
## Background

Push notifications are table stakes for mobile SaaS apps — they drive re-engagement, surface important events (billing failures, new messages, team activity), and keep users connected to the product. BeakerStack ships a full auth + billing + navigation scaffold, but has no notification layer. This issue adds the push notification infrastructure: token registration, the delivery Edge Function, and an in-app preferences screen. Billing event triggers (what to send when) are handled separately in #118, which depends on this issue.

## Business case

- Push notifications are among the highest-ROI mobile channels; templates that omit them force every team to reinvent the same plumbing.
- Expo's managed push service abstracts APN/FCM complexity behind a single unified API.
- Storing tokens in Supabase and sending via Edge Functions keeps credentials server-side and makes delivery auditable.
- Notification preferences reduce unsubscribe friction and are required by App Store / Play Store guidelines in many categories.

### Data model

```sql
create table push_tokens (
  id          uuid primary key default gen_random_uuid(),
  user_id     uuid not null references auth.users(id) on delete cascade,
  token       text not null,
  platform    text not null check (platform in ('ios', 'android')),
  updated_at  timestamptz not null default now(),
  unique (user_id, platform)
);
```

A table (not a column on `profiles`) correctly handles multiple devices and token rotation via upsert on `(user_id, platform)`. Migration also adds `notification_preferences jsonb default '{"billing":true,"activity":true,"marketing":false}'` to `profiles`.

### Client — `apps/mobile/src/lib/notifications/`

- **`registerForPushNotifications.ts`** — requests permission; calls `getExpoPushTokenAsync` inside `try/catch` (throws on simulator — log a dev warning, do not crash); upserts to `push_tokens`.
- **`useNotificationListener.ts`** — wires received/response listeners; exposes callback for deep-link handling.
- Token registration lives in a `useEffect` hook called from a component mounted above the navigator — navigator-agnostic, no `AppNavigator.tsx` dependency.

### Server — `supabase/functions/`

- **`send-push-notification/index.ts`** — accepts `{ userId, title, body, data? }`; looks up token from `push_tokens`; POSTs to Expo Push API (step 1); also polls Expo receipts endpoint (step 2) and deletes stale tokens on `DeviceNotRegistered` receipt error.
- Requires service-role auth; never callable from the client.

[...]

PR #137, opened 2026-05-13
feat: push notifications with device token registration and receipt polling (#125)

rahul-artificerinnovations:
## Summary
- `device_tokens` Supabase table with RLS (users can only manage their own tokens)
- `registerDevice()` requests permission, gets Expo push token, upserts to `device_tokens`, sets Android notification channel
- `useNotifications()` hook handles foreground notification display and deep-link navigation on tap
- `pollReceipts()` + `cleanupDeadToken()` for receipt polling with `DeviceNotRegistered` cleanup
- `NotificationHandler` null-rendered component wires the hook inside `NavigationContainer`
- `registerDevice` called on `SIGNED_IN` auth event in `App.tsx`

## Test plan
- [ ] Grant notification permissions on iOS simulator — token upserted to `device_tokens` in Supabase
- [ ] Deny permissions — `registerDevice` returns early, no error thrown
- [ ] Send test push from Expo dashboard — notification appears in foreground
- [ ] Tap notification with `{ screen: 'Dashboard' }` data — navigates to Dashboard screen
- [ ] Android: notification channel `default` created on first launch
- [ ] `DeviceNotRegistered` receipt — `console.warn` logged (cleanup path ready for ticket cross-reference)


mei-artificerinnovations (changes requested):
Token registration, permission handling, foreground notification display, and deep-link navigation are all solid. Migration is correct — `unique` on `token` (rather than the `unique(user_id, platform)` in the plan) is actually better for multi-device support; each device gets its own row and old tokens are cleaned up via receipt polling. One issue to flag:

---

**`DeviceNotRegistered` cleanup is incomplete — `cleanupDeadToken` is never called**

`pollReceipts` detects `DeviceNotRegistered` receipts correctly, but:

```ts
// in pollReceipts.ts
if (receipt.details?.error === 'DeviceNotRegistered') {
  // "in production this would cross-reference a tickets table; for now log"
  console.warn('[pollReceipts] DeviceNotRegistered receipt received');
}
```

`cleanupDeadToken` exists but is never invoked. The comment explains why: receipt polling returns ticket IDs as keys, not the original tokens — you can't look up which token to delete without a ticket ID → token mapping that doesn't exist yet.

This was the specific item I confirmed as in-scope when clearing #125 ("receipt polling + `DeviceNotRegistered` cleanup"). The scaffolding is there but the actual token deletion never fires, so stale tokens accumulate indefinitely after device uninstalls.

Two paths forward — pick one and call it out explicitly:

1. **Add a `push_tickets` table in this migration** — server stores `{ ticket_id, token }` after each send; `pollReceipts` cross-references it to get the token, then calls `cleanupDeadToken`. Closes the loop properly.

2. **File a follow-up issue** and update the PR description to say cleanup is deferred pending ticket-to-token mapping infrastructure. That's honest about the current state. The test plan item already soft-acknowledges this ("cleanup path ready for ticket cross-reference") but the plan conversation confirmed it as in-scope, so the gap should be explicit.

I'd recommend option 2 if option 1 would meaningfully expand this PR's scope. Either way, the PR description should be updated so reviewers know cleanup doesn't actually run.

---

**Non-blocking — `updated_at` won't update on upsert**

The upsert payload doesn't include `updated_at`, so on a conflict (same token, same device re-registering) the column stays at the original insert time. If "last seen" semantics are wanted later, a Supabase trigger would be needed. Not a problem now since `updated_at` isn't used anywhere.

---

**Non-blocking — `useNavigation<any>()` and #116**

After #116 restructures navigation to a tab bar, `navigation.navigate(data.screen)` will need screen names from the new param list. Low risk — just flag for the #116 implementation pass.

---

Fix or explicitly defer the cleanup gap, then I'll approve.

diego-artificerinnovations (changes requested):
**Changes requested — one blocking bug, everything else is clean.**

---

## Blocking: `NotificationHandler` is never rendered

`NotificationHandler` exists and calls `useNotifications()`, but it's not placed anywhere in the component tree — not in `App.tsx` and not in `AppNavigator.tsx`. The `addNotificationResponseReceivedListener` inside `useNotifications` only fires when the hook is mounted, so notification tap → navigate will never work.

`NotificationHandler` needs to be rendered **inside `NavigationContainer`** (so `useNavigation()` has a context). The right place is inside `AppNavigator`, after `<Stack.Navigator>`:

```tsx
// AppNavigator.tsx
import { NotificationHandler } from '../lib/notifications';

// inside the return:
<NavigationContainer ref={navigationRef}>
  <Stack.Navigator ...>
    {/* existing screens */}
  </Stack.Navigator>
  <NotificationHandler />
</NavigationContainer>
```

Note: `Notifications.setNotificationHandler` IS wired correctly — it runs at module load when `useNotifications.ts` is imported, so foreground notification display works. Only the response listener (tap → navigate) is broken.

---

## Everything else is correct

- Permission flow: check existing → request only if not granted → return early if still denied → get token → set Android channel → upsert ✓
- `onConflict: 'token'` matches the `unique` constraint on `token` in the migration ✓
- Migration: `on delete cascade` from `auth.users`, RLS `for all using + with check (auth.uid() = user_id)`, index on `user_id` ✓
- `useEffect` cleanup removes the response listener subscription ✓
- `registerDevice` called on `SIGNED_IN` event with `subscription.unsubscribe()` cleanup in `useEffect` return ✓
- `pollReceipts` stub with `DeviceNotRegistered` log acknowledged as incomplete in the PR description — acceptable for now ✓

---

## Non-blocking notes

**`updated_at` won't update on re-registration** — the upsert payload is `{ user_id, token, platform }`, not including `updated_at`. On conflict (same token re-registered), `updated_at` stays at original insert time. If you want it to reflect last-seen, either include it in the payload or add a `before update` trigger. Minor.

**`useNavigation<any>()`** — navigation is untyped in `useNotifications`; `data.screen as string` is unchecked. If `data.screen` is an invalid screen name, `navigate()` silently fails or throws. Consider checking against a known list before navigating, or type it via `RootStackParamList`. Low risk for now since only your server controls push content.

rahul-artificerinnovations:
Addressed the `pollReceipts` dead-code concern. Changes pushed:

- `pollReceipts` now returns `{ deadTicketIds: string[] }` instead of silently swallowing the `DeviceNotRegistered` path. The receipt→token mapping requires a `push_tickets` table that doesn't exist yet.
- Filed #138 to track the full cleanup path (`push_tickets` migration + wiring).
- `cleanupDeadToken(token)` stays in place — it's correct, just needs the ticket→token lookup from #138 before it can be called.

The PR description already notes "cleanup path ready for ticket cross-reference" — updating to reference the new issue.

mei-artificerinnovations (approved):
pollReceipts now returns deadTicketIds and the comment references issue 138 for the full cleanup path. The function is now callable by anything that has the ticket-to-token mapping. Approved.

diego-artificerinnovations (approved):
Block resolved. `NotificationHandler` imported and rendered at line 65 inside `NavigationContainer`, after `Stack.Navigator` — correct placement so `useNavigation()` has context. `Notifications.setNotificationHandler` was already right (module-level); now the response listener is also properly wired. Approved.

sarah-artificerinnovations (changes requested):
One blocker, same pattern as #136.

### Blocking: `expo-notifications` not in `apps/mobile/package.json`

`registerDevice.ts` and `useNotifications.ts` both import from `expo-notifications`, but there's no corresponding `package.json` change in this PR, and `expo-notifications` is not present on `develop`. Add it with `npx expo install expo-notifications` (Expo 50 resolves to `~0.28.x`).

---

### Non-blocking

**`registerDevice` on every `SIGNED_IN`** — Supabase fires `SIGNED_IN` on session restore at startup, not just on actual login. The permission prompt won't re-fire (already granted), but the upsert runs on every cold start. Harmless, just a small wasted round-trip.

**`useNavigation<any>()`** — `useNavigation<NavigationProp<RootStackParamList>>()` would give type safety on `navigation.navigate(data.screen)`. Minor.

sarah-artificerinnovations (approved):
`expo-notifications ~0.27.21` is the correct SDK 50 version — blocker resolved. (`expo-constants ~15.4.6` was already in `package.json`, confirmed in the diff.) Approved.

Issue #126, opened 2026-05-13
feat(mobile): first-run onboarding carousel and empty-state screens for new users


rahul-artificerinnovations:
## Background

New users who install a BeakerStack-based app are dropped directly onto the main dashboard — which, on a fresh account, is empty. There is no guided introduction to what the app does, where key features live, or what the user should do first. This creates a poor first impression and increases early churn. This issue adds a reusable onboarding carousel and empty-state components that template consumers can adapt in minutes.

## Business case

- First-run experience is the single highest-leverage touchpoint for retention.
- Template repos that include onboarding scaffolding save adopters a full sprint of design + engineering work.
- Empty states that explain what belongs there reduce support tickets and improve perceived quality.

## User story

**New user:** Signs up and sees a 3-slide introduction to the core value prop and key navigation areas. On the final slide they tap "Get started" and land on the dashboard. Each empty section shows a contextual empty state with a CTA rather than a blank list.

**Template developer:** Edits `onboardingSlides.ts` to customize copy and swap images — no carousel logic changes needed.

### Data model

Carousel completion is stored as a row in the `onboarding_steps` table from #119 with `step_key = 'carousel_completed'`. This eliminates a `profiles` column migration, gives a single source of truth for all onboarding state across #119 and #126, and lets teams query one table.

**Dependency:** #119's migration (which creates `onboarding_steps`) must be applied before or alongside this issue.

### Onboarding carousel — `apps/mobile/src/screens/onboarding/`

- **`slides.ts`** — typed config array (`title`, `description`, `imageSource`, `backgroundColor`). Default placeholders use PNG/JPEG assets (no additional native dependency; teams who have `react-native-svg` in their project can swap to SVG). Adding or removing a slide requires only `slides.ts` changes.
- **`OnboardingScreen.tsx`** — `ScrollView` with `pagingEnabled={true}` and `horizontal={true}` (simpler than FlatList for fixed slide count; dot indicator synced via `onScroll` `contentOffset`). "Skip" button (top-right) and "Next" / "Get started" CTA.
- **`useReduceMotion.ts`** — calls `AccessibilityInfo.isReduceMotionEnabled()` in `useEffect`, holds result in state. Slide transition is instant when enabled.
- **`useOnboardingGate.ts`** — queries `onboarding_steps` for `step_key = 'carousel_completed'`; returns `showOnboarding: true` when absent. Marks complete by inserting that row.

### Navigator integration

[...]

Issue #119, opened 2026-05-13
feat: user onboarding checklist — web and mobile


sarah-artificerinnovations:
## Business case

Activation rate — the percentage of new users who complete a meaningful set of actions in their first session — is the single strongest predictor of long-term retention. Users who never discover key features churn at significantly higher rates, not because the product is bad, but because they never saw enough of it to understand its value.

A post-signup onboarding checklist solves this directly:
- **Guides users** to the highest-value features immediately after sign-up
- **Creates completion drive** (Zeigarnik effect — incomplete tasks are more memorable; progress toward a goal increases motivation)
- **Reduces support load** — users who discover features through a checklist ask fewer "how do I..." questions
- **Improves trial-to-paid conversion** — users who activate fully are far more likely to upgrade

Every team building on BeakerStack will want onboarding. They should not have to build the infrastructure from scratch.

## Design

The checklist is a persistent, dismissible UI element shown to new users on the Dashboard (web) and as a dedicated screen/sheet (mobile). Each step is tracked server-side per user. Once all steps are complete, the checklist collapses into a "Setup complete" badge and can be dismissed permanently.

**Default checklist steps** (teams customise for their domain):

| Step | Trigger |
|---|---|
| ✅ Create your account | Auto-complete on first sign-in |
| Explore the dashboard | User visits Dashboard |
| View your plan | User visits Billing → Plans tab |
| Connect a payment method | User completes Stripe checkout OR is on a paid plan |
| Enable dark mode | User toggles theme |

Steps are intentionally achievable in a single session and map to existing BeakerStack features — teams replace them with their own product milestones.

## Data model

```sql
create table onboarding_steps (
  id           uuid primary key default gen_random_uuid(),
  user_id      uuid not null references auth.users(id) on delete cascade,
  step_key     text not null,
  completed_at timestamptz not null default now(),
  unique (user_id, step_key)
);
```

Step definitions (keys, labels, descriptions) live in a shared config file — not in the database — so they can be changed without a migration.

---

### Data layer (shared)
- [ ] Migration: `onboarding_steps` table with `user_id`, `step_key`, `completed_at`, unique constraint on `(user_id, step_key)`
- [ ] RLS: users can only read/insert their own rows; no update/delete (completion is append-only)
[...]

Issue #115, opened 2026-05-13
feat(mobile): legal document consent flow (ToS / Privacy Policy)


rahul-artificerinnovations:
## Background

BeakerStack's mobile app currently has no mechanism for presenting legal documents to users or recording their consent. This is a gap that affects every app built from this template the moment it acquires real users.

Mobile app stores, privacy regulations (GDPR, CCPA), and payment processors (Stripe) all require documented, versioned user consent to Terms of Service and Privacy Policy. Without a consent flow baked into the template, every team building on BeakerStack has to implement this from scratch — and doing it wrong (or skipping it) creates real legal and business risk for them.

---

## Business case

| Requirement | Without this feature |
|-------------|----------------------|
| Apple App Store / Google Play | Privacy policy URL required in store listing and accessible inside the app. Rejection risk on new submission or update review. |
| GDPR / CCPA | Consent must be informed, specific, and recorded with a timestamp. No consent record = no defensible position in a subject-access or deletion request. |
| Stripe ToS | Stripe's platform agreement requires connected platforms to surface their own ToS and obtain user agreement before processing payments. |
| Enterprise / B2B customers | Procurement reviews routinely ask "where and when did users accept your ToS?" A template with no answer is a blocker for customers in regulated industries. |

---

### 1. First-time onboarding gate
A new user completes sign-up (email/password or OAuth). Before they reach the app's home screen, they are shown a consent screen presenting the current ToS and Privacy Policy. They must explicitly check both consent checkboxes to continue. This is the only moment in the user lifecycle where consent can reasonably be captured — after the account exists but before any billable or data-processing activity begins.

### 2. Re-consent on material update
When the ToS or Privacy Policy is updated, existing users who accepted an older version encounter a re-consent screen on their next app open. They see a brief summary of what changed and must re-accept before accessing the app. This keeps consent current without requiring every user to re-register.

### 3. Persistent in-app access from Settings
At any time, a signed-in user can navigate to **Settings → Legal** and view the full current text of the ToS and Privacy Policy. This is required by GDPR (right to access the basis of processing), expected by users, and checked by app-store reviewers.

### 4. Acceptance history (for the user)
The Settings → Legal screen also shows the version and acceptance date of the documents the user has agreed to — useful transparency and occasionally requested by users who want to know what version they accepted.

---

### Data model

Add two tables to Supabase:

```sql
-- Operator-controlled: which version of each doc is currently required
create table legal_config (
[...]

PR #139, opened 2026-05-13
feat: onboarding carousel, empty states, and onboarding_steps migration (#126)

rahul-artificerinnovations:
## Summary
- `onboarding_steps` Supabase migration with RLS (includes schema for #119 since it's unassigned — first to implement owns it)
- 3-slide onboarding carousel (`OnboardingScreen` + `slides.ts`) — `ScrollView` with `pagingEnabled`, dot indicators synced via `onScroll`
- `useOnboardingGate` hook — queries `onboarding_steps` for `carousel_completed`, marks complete on skip/finish
- `useReduceMotion` hook — async `AccessibilityInfo.isReduceMotionEnabled()` in `useEffect`; instant transitions when enabled
- `EmptyState` component — illustration (optional), title, description, action button with full `accessibilityLabel`
- `OnboardingGate` mounted above authenticated navigator — navigator-agnostic, compatible with #116
- `__DEV__`-only long-press debug reset on version label in ProfileScreen
- Feature-flagged via `EXPO_PUBLIC_ONBOARDING_ENABLED` (default true)

## Test plan
- [ ] New user (no `carousel_completed` row) sees carousel on first login
- [ ] Returning user (row exists) goes straight to main app
- [ ] "Skip" on any non-final slide inserts `carousel_completed` and navigates to main app
- [ ] "Get started" on final slide inserts `carousel_completed` and navigates to main app
- [ ] Dot indicator reflects current slide index
- [ ] `reduceMotion=true`: slide transitions are instant (no animation)
- [ ] Each slide's `accessibilityLabel` includes slide number (e.g. "Welcome to BeakerStack. Slide 1 of 3")
- [ ] Dashboard shows `EmptyState` when data array is empty
- [ ] `__DEV__` long-press (500ms) on version label deletes `carousel_completed` row
- [ ] Sign out → sign back in → carousel does NOT re-appear (row persists)
- [ ] `EXPO_PUBLIC_ONBOARDING_ENABLED=false` → carousel never shown

Note on #119: `onboarding_steps` table is created by this PR's migration. Anyone implementing #119 should build against this table.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,2 @@
+# Feature flags
+EXPO_PUBLIC_ONBOARDING_ENABLED=true
--- a/apps/mobile/src/components/EmptyState/EmptyState.tsx
+++ b/apps/mobile/src/components/EmptyState/EmptyState.tsx
@@ -1,0 +1,57 @@
+import React from 'react';
+import { Image, ImageSourcePropType, StyleSheet, Text, TouchableOpacity, View } from 'react-native';
+
+interface Props {
+  title: string;
+  description: string;
+  illustration?: ImageSourcePropType;
+  action?: { label: string; onPress: () => void; accessibilityLabel?: string };
+}
+
+export function EmptyState({ title, description, illustration, action }: Props) {
+  return (
+    <View style={styles.container}>
+      {illustration && (
+        <Image
+          source={illustration}
+          style={styles.illustration}
+          accessibilityRole="none"
+          accessible={false}
+        />
+      )}
+      <Text style={styles.title}>{title}</Text>
+      <Text style={styles.description}>{description}</Text>
+      {action && (
+        <TouchableOpacity
+          style={styles.button}
+          onPress={action.onPress}
+          accessibilityLabel={action.accessibilityLabel ?? action.label}
+          accessibilityRole="button"
+        >
+          <Text style={styles.buttonText}>{action.label}</Text>
+        </TouchableOpacity>
+      )}
+    </View>
+  );
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    alignItems: 'center',
+    justifyContent: 'center',
+    padding: 32,
+  },
+  illustration: { width: 120, height: 120, marginBottom: 24 },
+  title: { fontSize: 20, fontWeight: '700', color: '#111827', marginBottom: 8, textAlign: 'center' },
+  description: { fontSize: 15, color: '#6b7280', textAlign: 'center', lineHeight: 22, marginBottom: 24 },
+  button: {
+    backgroundColor: '#111827',
+    paddingVertical: 12,
+    paddingHorizontal: 32,
+    borderRadius: 8,
+    minWidth: 180,
+    alignItems: 'center',
+  },
+  buttonText: { color: '#fff', fontSize: 16, fontWeight: '600' },
+});
--- a/apps/mobile/src/components/EmptyState/index.ts
+++ b/apps/mobile/src/components/EmptyState/index.ts
@@ -1,0 +1,1 @@
+export { EmptyState } from './EmptyState';
--- a/apps/mobile/src/hooks/useOnboardingGate.ts
+++ b/apps/mobile/src/hooks/useOnboardingGate.ts
@@ -29,7 +29,7 @@ export function useOnboardingGate(userId: string | null): OnboardingGateResult {
       .eq('user_id', userId)
       .eq('step_key', 'carousel_completed')
       .maybeSingle()
-      .then(({ data }) => {
+      .then(({ data, error }) => {
         setShowOnboarding(!error && !data);
         setLoading(false);
       });
--- a/apps/mobile/src/hooks/useOnboardingGate.ts
+++ b/apps/mobile/src/hooks/useOnboardingGate.ts
@@ -30,7 +30,7 @@ export function useOnboardingGate(userId: string | null): OnboardingGateResult {
       .eq('step_key', 'carousel_completed')
       .maybeSingle()
       .then(({ data }) => {
-        setShowOnboarding(!data);
+        setShowOnboarding(!error && !data);
         setLoading(false);
       });
   }, [userId]);
--- a/apps/mobile/src/hooks/useOnboardingGate.ts
+++ b/apps/mobile/src/hooks/useOnboardingGate.ts
@@ -1,0 +1,48 @@
+import { useEffect, useState } from 'react';
+import { supabase } from '../lib/supabase';
+
+interface OnboardingGateResult {
+  showOnboarding: boolean;
+  loading: boolean;
+  markComplete: () => Promise<void>;
+}
+
+export function useOnboardingGate(userId: string | null): OnboardingGateResult {
+  const [showOnboarding, setShowOnboarding] = useState(false);
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    if (!userId) {
+      setLoading(false);
+      return;
+    }
+
+    const enabled = process.env.EXPO_PUBLIC_ONBOARDING_ENABLED !== 'false';
+    if (!enabled) {
+      setLoading(false);
+      return;
+    }
+
+    supabase
+      .from('onboarding_steps')
+      .select('step_key')
+      .eq('user_id', userId)
+      .eq('step_key', 'carousel_completed')
+      .maybeSingle()
+      .then(({ data }) => {
+        setShowOnboarding(!data);
+        setLoading(false);
+      });
+  }, [userId]);
+
+  const markComplete = async () => {
+    if (!userId) return;
+    await supabase.from('onboarding_steps').upsert(
+      { user_id: userId, step_key: 'carousel_completed', completed: true, completed_at: new Date().toISOString() },
+      { onConflict: 'user_id,step_key' }
+    );
+    setShowOnboarding(false);
+  };
+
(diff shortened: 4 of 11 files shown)


mei-artificerinnovations (changes requested):
`useReduceMotion` with live listener, `EmptyState` with accessibility, dot pill indicator, `onConflict: 'user_id,step_key'` upsert, migration with composite PK — all solid. Two things:

---

**1. `useOnboardingGate` network error shows carousel to returning users — request fix**

```ts
.maybeSingle()
.then(({ data }) => {
  setShowOnboarding(!data);  // error not destructured — data is null on failure
  setLoading(false);
});
```

`maybeSingle()` returns `{ data, error }`. If the query fails (network error, RLS issue), `data` is `null` and `!null === true` — returning users who have already completed onboarding see the carousel again. The behaviour on error should be fail-open (don't show carousel), not fail-closed:

```ts
.then(({ data, error }) => {
  setShowOnboarding(error ? false : !data);
  setLoading(false);
});
```

---

**2. `AuthenticatedRoot` will conflict with #116 — needs coordination note**

This PR's `AuthenticatedRoot` handles the onboarding gate and renders the authenticated `Stack.Navigator`. My #116 plan also defines `AuthenticatedRoot` to handle the legal consent gate (#115) and render `BottomTabNavigator`.

When #116 lands, its `AuthenticatedRoot` will replace the one here. The onboarding gate will need to be folded into #116's `AuthenticatedRoot` as a third layer:

```tsx
function AuthenticatedRoot() {
  const { status, requiredDocs } = useLegalConsent();     // #115
  const { showOnboarding, loading, markComplete } = useOnboardingGate(userId);
  if (status === 'loading' || loading) return <LoadingScreen />;
  if (status === 'required') return <LegalConsentScreen requiredDocs={requiredDocs} />;
  if (showOnboarding) return <OnboardingScreen onComplete={markComplete} />;
  return <BottomTabNavigator />;
}
```

Not blocking this PR — just flagging that #116's implementation needs to fold `useOnboardingGate` in, and that #139 should merge before #116 to avoid a conflict. Adding a note to the #116 plan.

---

Fix the error handling and I'll approve.

mei-artificerinnovations (changes requested):
Fix is almost there — `error` is referenced in the condition but not destructured from the `.then()` callback. This will be a TypeScript compile error:

```ts
// current (line 32) — error not in scope
.then(({ data }) => {
  setShowOnboarding(!error && !data);

// should be
.then(({ data, error }) => {
  setShowOnboarding(!error && !data);
```

One character change on the destructure line and it's done.

sarah-artificerinnovations (changes requested):
Good overall structure — `useReduceMotion`, dot indicators, `scrollEventThrottle`, accessibility labels, and the dev long-press reset are all well done. Two issues to fix.

### Blocking: `useOnboardingGate` missing error handler — loading spinner hangs forever

The `.then()` in the `useEffect` only calls `setLoading(false)` on success. If the Supabase query fails (network error, RLS mismatch, etc.) `loading` stays `true` and the navigator renders nothing until the user force-quits. Add a `.catch`:

```ts
.then(({ data }) => {
  setShowOnboarding(!data);
  setLoading(false);
})
.catch(() => {
  // On error, don't block the user — let them through
  setShowOnboarding(false);
  setLoading(false);
});
```

### Blocking: `AppNavigator` duplicates screen list — unmounts navigator on auth change

The PR splits into two separate `<Stack.Navigator>` instances (`AuthenticatedRoot` and the unauthenticated branch). Every screen is listed twice, and when auth state changes the entire navigator unmounts and remounts. This causes navigation state loss and will be a jarring UX at sign-in.

A simpler fix: keep the single `<Stack.Navigator>` from the current `AppNavigator`, and overlay the `OnboardingScreen` on top when it should show:

```tsx
return (
  <NavigationContainer ref={navigationRef}>
    <Stack.Navigator ...>
      {/* screens as before */}
    </Stack.Navigator>
    <NotificationHandler />
    {userId && showOnboarding && (
      <OnboardingScreen onComplete={markComplete} />
    )}
  </NavigationContainer>
);
```

This keeps a single stable navigator tree. `OnboardingScreen` renders as a full-screen overlay (position absolute or flex:1) above the stack, disappears when complete.

---

### Non-blocking

**`expo-constants` in `ProfileScreen.tsx`** — verify it's in `apps/mobile/package.json` before merge (same pattern as the netinfo issue in #136).

**Screen boundaries / `NotificationHandler`** — once #135 and #137 merge to `develop`, `AppNavigator.tsx` on this branch will conflict. The overlay approach above makes that merge simpler since you're only inserting two lines rather than restructuring the whole component.

mei-artificerinnovations (approved):
Confirmed —  correctly destructured. **Approved. ✅**

diego-artificerinnovations (changes requested):
**Changes requested — one blocking compile error remaining from Mei's fix attempt, and a few non-blocking notes.**

---

## Blocking: `error` is used but not destructured in `useOnboardingGate`

Mei's second review nails it. The logic is right; the destructure is one token short:

```ts
// current — TypeScript compile error, `error` not in scope
.then(({ data }) => {
  setShowOnboarding(!error && !data);

// fix
.then(({ data, error }) => {
  setShowOnboarding(!error && !data);
```

Fail-open on error (don't show carousel) is the correct behaviour. One character change on the destructure line.

---

## Non-blocking: unused `Animated` import (Copilot already flagged)

`OnboardingScreen.tsx` imports `Animated` from `react-native` but doesn't use it — `scrollTo` doesn't need it. Remove `Animated,` from the import list.

---

## Non-blocking: skip button a11y on final slide

On the last slide the skip `TouchableOpacity` is still rendered and tappable — only the `Text` child is hidden. A screen reader still announces `"Skip onboarding"` on a slide where skip and "Get started" are semantically identical. Consider pulling the conditional up to wrap the whole `TouchableOpacity`, or add `accessibilityElementsHidden={isLast}` to exclude it from the a11y tree on the last slide.

---

## Non-blocking: `AppNavigator.tsx` will also conflict with PR #137

Mei flagged the #116 coordination. Same applies to PR #137 (push notifications, already approved, waiting for Brad to merge) — it adds `<NotificationHandler />` inside `<NavigationContainer>` in `AppNavigator.tsx`. Since #139 restructures that same file, whoever merges second will need to place `<NotificationHandler />` after the `{userId ? <AuthenticatedRoot /> : <Stack.Navigator ...>}` block but still inside `<NavigationContainer>`. Brad should keep this in mind when sequencing the merges.

---

## Everything else is correct

- Migration composite PK `(user_id, step_key)` matches `onConflict: 'user_id,step_key'` in upsert ✓
- RLS `for all using + with check (auth.uid() = user_id)`, `on delete cascade`, index on `user_id` ✓
- `useReduceMotion`: async init + `addEventListener` + `sub.remove()` cleanup ✓
- Feature flag evaluated at call time (`!== 'false'`), not module load ✓
- `OnboardingGate` rendered outside `Stack.Navigator` (full-screen, no header) ✓
- `scrollTo({ animated: !reduceMotion })` respects reduce-motion preference ✓
- `accessibilityLabel` includes slide number on each slide ✓
- `EmptyState`: illustration `accessible={false}`, action `accessibilityLabel` fallback, `accessibilityRole="button"` ✓
- `__DEV__` long-press guard on debug reset ✓
- `void handleDevResetOnboarding()` correct for non-awaited async handler ✓

mei-artificerinnovations (approved):
The overlay approach resolves the AuthenticatedRoot naming conflict entirely — the onboarding gate is now a navigator-agnostic absolute overlay, so #116 just replaces the Stack.Navigator with BottomTabNavigator and this works unchanged. Single navigator instance, no duplication, clean separation. Also noticed useOnboardingGate handles userId=null correctly (loading=false, showOnboarding=false) so unauthenticated users are unaffected.

Approved. ✅

diego-artificerinnovations (approved):
**Approved — all blocks resolved. My previous review was based on a stale commit; the `{ data, error }` destructure was already fixed before I reviewed.**

---

## Current state

**`useOnboardingGate.ts`** (both Sarah's and Mei's concerns resolved):
- `{ data, error }` correctly destructured ✓
- `.catch(() => { setShowOnboarding(false); setLoading(false); })` added — no more stuck loading on network error ✓
- `setShowOnboarding(!error && !data)` fail-open logic ✓

**`AppNavigator.tsx`** (Sarah's duplicate-navigator concern resolved):
- Single `Stack.Navigator` retained — no navigator remount on auth change ✓
- `OnboardingScreen` rendered as `StyleSheet.absoluteFill` overlay inside `NavigationContainer` ✓
- `!loading && showOnboarding` guard prevents flash while query is in-flight ✓

**`expo-constants`** in `apps/mobile/package.json`: `~15.4.6` ✓

---

Non-blocking notes from my earlier review still stand (unused `Animated` import, skip button a11y on final slide, #137 + #116 `AppNavigator` merge sequencing) — none blocking merge.

sarah-artificerinnovations (approved):
Both blockers resolved:

1. **`useOnboardingGate` error handling** — `.catch()` added, and the `.then()` now destructures `{ data, error }` and uses `!error && !data` for `showOnboarding`. That's better than the original — an errored query now correctly fails open (no carousel) rather than being mistaken for a missing step.

2. **`AppNavigator` overlay** — single `<Stack.Navigator>` preserved, `OnboardingScreen` rendered inside `<View style={StyleSheet.absoluteFill}>` as a sibling after the navigator. Clean pattern, navigator state stable across auth changes.

Approved.

PR #149, opened 2026-05-13
fix(ci): fix GitHub Deployment visibility for PR previews (#140)

### LINKED ISSUE #140: Show PR web preview in GitHub Deployments UI (keep existing PR comment)
## Summary

PR preview environments already post a rich sticky comment (web, database, billing, mobile, etc.). We also want the **official GitHub Deployments** area on the pull request to show the **web preview** as a real deployment with a **direct link** to the preview URL (same URL users get from the comment flow: public preview path or signed-cookie bootstrap when configured).

**Do not remove or replace** the existing comment block; this work is **additive**.

## Current behavior

- The sticky PR comment (marker `<!-- pr-preview-links -->`) remains the place for full instructions, including mobile.
- A `Create GitHub Deployment (PR preview)` step exists in `.github/workflows/pr-preview-environment.yml`, but deployments are **not reliably visible** in the PR Deployments UI for web previews (needs investigation: e.g. timing, `ref`, fork PRs, permissions, or status lifecycle).

## Desired behavior

- After a successful **web** PR preview deploy, the PR shows an **official deployment** in GitHub’s UI.
- **`environment_url`** (or equivalent) opens the **web** preview (path under `deploy.<domain>/…` or signed bootstrap URL when signing is enabled), matching what we already surface in the comment table.
- **Keep** the current PR comment content and update behavior as-is unless a tiny cross-link in the comment is explicitly agreed (default: **no** comment changes).

## Acceptance criteria

- [ ] Opening a PR into `develop` that runs the preview workflow results in a **visible** GitHub deployment for the **web** preview with a working link.
- [ ] The existing sticky preview comment continues to be posted/updated as today.
- [ ] Document any limitations (e.g. PRs from forks) if they still block deployment API writes.

## Out of scope (for now)

- Mobile as a separate GitHub deployment row, staging/production deployment parity, or parsing PR template fields for deployment descriptions—unless needed to fix visibility.

sarah-artificerinnovations:
## Summary

Fixes three issues that caused the PR Deployments panel to show nothing after a successful web preview deploy.

**Root causes fixed:**

- **Silent 202 drop**: `createDeployment` can return HTTP 202 (not 201) when the request can't be immediately fulfilled (environment rules, duplicate, etc.). `deployment.data` has a `message` field and no `id` in that case — the previous `if (deployment.data.id)` check silently skipped `createDeploymentStatus`, so no deployment ever appeared in the PR UI. Now logs a warning with the API response instead.
- **Stale rows on re-runs**: Each push created a new deployment row without deactivating the previous one. Now lists existing deployments for the same environment and marks them `inactive` before creating a fresh one — keeps the PR Deployments panel to one active row per PR.
- **Fork PRs**: `pull_request` events from forks run with a read-only token; `deployments: write` in the workflow header doesn't apply. Now detects fork PRs early and exits cleanly with a log message instead of failing silently.

**Also added:** `log_url` on `createDeploymentStatus` so the deployment row in GitHub's UI links directly to the Actions run that produced it.

**No changes to:** the sticky PR comment block, any other job, or deployment naming.

Closes #140

## Test plan
- [ ] Open or push to a same-repo PR into `develop` — deployment appears in PR Deployments panel with working link
- [ ] Push again — only one active row (previous marked inactive)
- [ ] Fork PR — CI step logs skip message, no error
- [ ] Existing sticky comment still posts/updates unchanged

### DIFF
--- a/.github/workflows/pr-preview-environment.yml
+++ b/.github/workflows/pr-preview-environment.yml
@@ -268,10 +268,11 @@ jobs:
           script: |
             // Fork PRs: pull_request events from forks run with a read-only token;
             // deployments: write is unavailable — skip cleanly rather than error.
-            const prRepo = context.payload.pull_request.head.repo.full_name;
+            // head.repo can be null when the fork is deleted before the workflow runs.
+            const headRepo = context.payload.pull_request.head.repo?.full_name;
             const baseRepo = `${context.repo.owner}/${context.repo.repo}`;
-            if (prRepo !== baseRepo) {
-              core.info(`Skipping: fork PR (${prRepo}) — Deployments API write not available for fork PRs.`);
+            if (!headRepo || headRepo !== baseRepo) {
+              core.info(`Skipping: fork PR (${headRepo ?? 'deleted repo'}) — Deployments API write not available for fork PRs.`);
               return;
             }
 
@@ -281,27 +282,14 @@ jobs:
             const environmentUrl = process.env.BOOTSTRAP_URL || webUrl;
             const logUrl = `${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`;
 
-            // Deactivate previous deployments for this environment so the PR
-            // Deployments panel shows only one active row per PR on re-runs.
-            const { data: existing } = await github.rest.repos.listDeployments({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              environment,
-              per_page: 20,
-            });
-            for (const dep of existing) {
-              await github.rest.repos.createDeploymentStatus({
-                owner: context.repo.owner,
-                repo: context.repo.repo,
-                deployment_id: dep.id,
-                state: 'inactive',
-              });
-            }
-
+            // Create the new deployment first. Only deactivate previous deployments
+            // after success so a partial failure doesn't leave the PR Deployments
+            // panel empty. Use head.sha (not context.sha which is the synthetic merge
+            // commit) so GitHub associates the deployment with the PR's own head ref.
             const deployment = await github.rest.repos.createDeployment({
               owner: context.repo.owner,
               repo: context.repo.repo,
-              ref: context.sha,
+              ref: context.payload.pull_request.head.sha,
               environment,
               auto_merge: false,
               required_contexts: [],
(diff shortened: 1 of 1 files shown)


ZappoMan:
## GitHub Deployment shows on environment page but not on the PR

**What’s going on**

The `Create GitHub Deployment (PR preview)` step calls `repos.createDeployment` with `ref: context.sha`.

On `pull_request` workflows, **`context.sha` is the synthetic merge commit** (GitHub’s test merge of base into head), **not** the PR’s head commit (`pull_request.head.sha`).

- Repo/environment views (e.g. **Deployments → `pr-*-preview`**) list **all** deployments for that environment, so rows still appear there.
- The **PR page** only surfaces deployments GitHub associates with the PR’s **head** branch/commits. If `ref` points at the merge commit, that association often doesn’t happen, so the PR UI stays empty even though the API and environment page look fine.

You can confirm by comparing `deployment.ref` from the API to `headRefOid` from this PR — they won’t match while using `context.sha`.

**Recommended fix**

For this job (pull_request only), pass the PR head into `createDeployment`, e.g.:

- `ref: context.payload.pull_request.head.sha`, or  
- `ref: context.payload.pull_request.head.ref` (branch name)

Either ties the deployment to the same ref the PR UI expects. Same pattern as other Actions + Deployments setups (e.g. discussion around merge ref vs head for PR deployments).

---

**If this is fixed in a later commit on this same PR:** you do **not** need a new PR. The next successful preview run will create a **new** deployment with the corrected `ref`, and that one should show on the **PR** as well as on the environment page. Older deployments created with the merge SHA may still exist in history but won’t block the new row from appearing.

sarah-artificerinnovations:
Thanks for the review — three fixes in this commit:

**1. Root cause fix: `ref: context.sha` → `ref: context.payload.pull_request.head.sha`**

On `pull_request` events, `context.sha` is GitHub's synthetic merge commit (base merged into head), not the PR's own head commit. The Deployments API associates a deployment with a PR only when the `ref` matches the head branch or head SHA — using the merge commit means GitHub can't make that link, so the deployment appears on the environment page but not on the PR itself. Switching to `head.sha` fixes that.

**2. Null-safe fork detection**

`head.repo` can be `null` when a fork is deleted before the workflow runs. Changed from `head.repo.full_name` (throws `TypeError`) to `head.repo?.full_name` with a `!headRepo` guard that treats a missing repo the same as a fork — skip cleanly with an info log.

**3. Create-before-deactivate ordering**

Previous code deactivated all existing deployments before creating the new one. If `createDeployment` then returned a 202 with no id, the early `return` left the PR Deployments panel completely empty. Flipped the order: create (and set `success` status on) the new deployment first, then iterate previous ones to mark them `inactive`. A partial failure now leaves old deployments visible rather than blanking the panel. Also bumped `per_page` from 20 → 100 and added a `dep.id !== deployment.data.id` guard to skip the just-created deployment in the deactivation loop.

PR #152, opened 2026-05-13
fix(seo): fix robots.txt returning HTML on PR preview domain (#142)

### LINKED ISSUE #142: SEO: fix invalid robots.txt (HTML error page served as robots)
## Context
Lighthouse CI (mobile, v12.1.0) reported **SEO 92** with `robots.txt is not valid` — the parser treated the response as robots syntax but received **HTML** (e.g. `<!DOCTYPE html>`, CloudFront **Preview Not Found** page), producing many spurious line errors.

**Report:** https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1778677581911-14662.report.html

## Scope
- Ensure `GET /robots.txt` on the audited host returns either:
  - **200** with `Content-Type: text/plain` and valid robots rules, or
  - **404** with a non-HTML body (if you intentionally omit robots.txt — confirm crawler expectations first).
- Fix routing so SPA / preview error fallthrough does not serve HTML for `/robots.txt`.

## Acceptance criteria
- [ ] `/robots.txt` is not an HTML document.
- [ ] Lighthouse **robots.txt is valid** passes (or equivalent manual check with Google’s robots tester).
- [ ] Document any preview vs production behavior if paths differ per environment.

## Notes
Root cause is almost certainly **wrong content-type / wrong handler** for the `robots.txt` path, not malformed robots directives.

sarah-artificerinnovations:
Closes #142

## What

Two changes to ensure `GET /robots.txt` always returns a valid `text/plain` response and never an HTML error page.

## Root cause

Requests for `/robots.txt` on the preview domain (`deploy.beakerstack.com`) didn't match the `/pr-{N}/...` pattern in `PRPathRouter.js`, so they passed through to S3 unchanged. No file exists at the bucket root, S3 returned 404, CloudFront fell back to its default HTML error page — which Lighthouse flagged as invalid robots.txt (SEO score 92 → robots.txt not valid).

## Changes

**`infra/aws/functions/PRPathRouter.js`**

Adds a synthetic response at the top of the handler, before the S3 origin request:

```js
if (uri === '/robots.txt') {
  return {
    statusCode: 200,
    statusDescription: 'OK',
    headers: { 'content-type': { value: 'text/plain' } },
    body: 'User-agent: *\nDisallow: /\n',
  };
}
```

No workflow changes — the function is already published by the existing "Publish PR path router" step.

**`apps/web/public/robots.txt`** (new file)

Vite copies `public/` into `dist/` during build, so this file lands in the production build artifact and deploys to the production bucket at the domain root. Production allows all crawlers; preview environments get the CloudFront Function response above.

## Preview vs production behavior

| Environment | Domain | Source | Crawlers |
|---|---|---|---|
| PR preview | `deploy.beakerstack.com` | CloudFront Function synthetic response | Blocked (`Disallow: /`) |
| Production | `beakerstack.com` | `dist/robots.txt` via Vite `public/` | Allowed (`Allow: /`) |

### DIFF
--- a/apps/web/public/robots.txt
+++ b/apps/web/public/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Allow: /
--- a/infra/aws/functions/PRPathRouter.js
+++ b/infra/aws/functions/PRPathRouter.js
@@ -3,16 +3,21 @@ function handler(event) {
   var uri = request.uri || '/';
   var previewPrefixBase = '%%PREVIEW_PREFIX%%';
 
-  // Preview domain should never be crawled. Return a synthetic robots.txt
-  // before any other routing so S3 (which has no file at the bucket root)
-  // never gets the request — avoiding the CloudFront HTML error-page fallback
-  // that causes Lighthouse to fail the robots.txt validity check.
+  // Return a synthetic robots.txt before any S3 routing to avoid the
+  // CloudFront HTML error-page fallback (no file at bucket root → 404 →
+  // HTML) that causes Lighthouse to fail the robots.txt validity check.
+  //
+  // Disallow the deploy origin by default, then explicitly Allow the PR
+  // prefix so Lighthouse (and Google) treat /pr-<N>/... as crawlable.
+  // Longest-prefix matching means /pr-152/ is allowed while stray apex
+  // paths remain blocked. Uses previewPrefixBase (not a hardcoded "pr-")
+  // so custom PREVIEW_PREFIX stacks stay correct.
   if (uri === '/robots.txt') {
     return {
       statusCode: 200,
       statusDescription: 'OK',
       headers: { 'content-type': { value: 'text/plain' } },
-      body: 'User-agent: *\nDisallow: /\n',
+      body: 'User-agent: *\nDisallow: /\nAllow: /' + previewPrefixBase + '\n',
     };
   }
 
--- a/infra/aws/functions/PRPathRouter.js
+++ b/infra/aws/functions/PRPathRouter.js
@@ -3,6 +3,19 @@ function handler(event) {
   var uri = request.uri || '/';
   var previewPrefixBase = '%%PREVIEW_PREFIX%%';
 
+  // Preview domain should never be crawled. Return a synthetic robots.txt
+  // before any other routing so S3 (which has no file at the bucket root)
+  // never gets the request — avoiding the CloudFront HTML error-page fallback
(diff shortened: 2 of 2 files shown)


ZappoMan:
## robots.txt on the deploy host — suggestion

Thanks for tackling #142 — serving a real `text/plain` `robots.txt` instead of CloudFront’s HTML error page is the right fix for the **“robots.txt is not valid”** failure.

One follow-up from how Lighthouse scores **SEO** today (v12+): a host-wide

```txt
User-agent: *
Disallow: /
```

fixes validity but tends to fail **“Page is blocked from indexing” / crawlability**, which is a large slice of the SEO category. So on PR preview URLs you can end up with a **worse overall SEO score** even though `robots.txt` parses correctly.

For **public** previews where we are **not** blocking anonymous access (signed cookies off), we actually **want** Lighthouse (and other crawlers) to treat `/pr-<N>/…` as crawlable. A pattern that matches that intent better is **disallow the whole deploy origin by default, then explicitly allow only the PR path prefix**:

```txt
User-agent: *
Disallow: /
Allow: /pr-
```

Google resolves that using longest matching prefix, so URLs like `/pr-152/…` stay allowed while stray apex paths stay disallowed.

**Implementation detail:** the allow prefix should follow the same injected preview prefix the function already uses (`previewPrefixBase` / `%%PREVIEW_PREFIX%%`), not a hardcoded `pr-`, so stacks that override `PREVIEW_PREFIX` stay correct.

When **signed cookies** (or similar) are on, Lighthouse is often blocked before robots semantics matter — there we can document a different expectation (e.g. SEO assertions only on prod, or cookie-aware runs).

Happy to pair on wording for the issue acceptance criteria / preview-vs-prod table if useful.

sarah-artificerinnovations:
Good catch — addressed in `3e8cfc8c`.

`Disallow: /` alone validates the file but fails Lighthouse's **"Page is blocked from indexing"** crawlability check, which is a significant chunk of the SEO score. Updated the synthetic response to:

```
User-agent: *
Disallow: /
Allow: /pr-
```

Google resolves this with longest-prefix matching, so `/pr-152/…` stays crawlable while stray apex paths stay blocked.

The `Allow` line uses `previewPrefixBase` (`%%PREVIEW_PREFIX%%`) rather than a hardcoded `pr-`, so stacks with a custom `PREVIEW_PREFIX` continue to work correctly — it resolves to whatever the deploy script injects at publish time.

PR #202, opened 2026-05-14
test(web): add AuthShell render test; remove coverage exclusion

### LINKED ISSUE #196: test: add AuthShell render test for basePath passthrough and provider ordering (#182 follow-up)
## Background

In PR #182 (AuthShell composition wrapper), `apps/web/src/AuthShell.tsx` is excluded from the coverage threshold in `vite.config.ts` as a "thin composition wrapper with no testable logic." This was the right call to keep the PR focused, but a coverage exclusion with no follow-up tracking invites permanent blind spots.

## What this issue tracks

Add a focused render test for `AuthShell.tsx` that verifies:

1. **`basePath` passthrough** — the `basePath` prop reaches `BrowserRouter` (snapshot or rendered href check)
2. **Provider ordering** — `ThemeProvider` wraps `AuthProvider` wraps `ProfileProvider` wraps `BrowserRouter` (verify context is available at the leaf)
3. **Children render** — a child component injected via router renders without throwing

## After this is done

Remove `src/AuthShell.tsx` from the `coverage.exclude` list in `vite.config.ts` so the 95% bar applies.

## Notes

- `AuthProvider` and `ProfileProvider` are from `@beakerstack/shared` — use the same mock pattern as `main.test.tsx` and `HomePage.test.tsx`
- Keep the test focused: do not re-test the internal behavior of the individual providers

mei-artificerinnovations:
Closes #196

## Summary
- Adds `src/__tests__/AuthShell.test.tsx` with three focused tests:
  1. **Children render** — `AuthShell` mounts without throwing
  2. **basePath passthrough** — `basePath` prop forwarded to `BrowserRouter` as `basename` (verified via `useHref` at the leaf)
  3. **Provider ordering** — `AuthProvider → ProfileProvider → BrowserRouter` confirmed by consuming `useAuthContext` and `useProfileContext` inside a mock `App` child
- Removes `src/AuthShell.tsx` from `vite.config.ts` `coverage.exclude`; the 99% threshold now applies

Mock pattern mirrors `HomePage.test.tsx`: real providers with mocked supabase client.

## Test plan
- [ ] CI passes with coverage ≥ 99% (AuthShell is now in scope)
- [ ] All three AuthShell test cases pass

### DIFF
--- a/apps/web/scripts/prerender-home.ts
+++ b/apps/web/scripts/prerender-home.ts
@@ -28,6 +28,7 @@ const { AppFooter } = await import('../src/components/AppFooter');
 // LandingPageSSR eagerly imports all sections. LandingPage uses React.lazy() for
 // below-fold sections, which resolve as empty Suspense fallbacks under renderToStaticMarkup.
 const { LandingPageSSR } = await import('../src/components/landing/LandingPageSSR');
+const { LAYOUT } = await import('../src/lib/layoutConstants');
 
 const webRoot = join(dirname(fileURLToPath(import.meta.url)), '..');
 
@@ -48,10 +49,6 @@ const routerBasename =
 // match (RR warns and renders nothing). Use the same pathname as publicHomeUrl path.
 const initialEntries = [homePath];
 
-// Layout wrapper classes mirror App.tsx ('bg-gray-50 dark:bg-gray-900') and RootLayout
-// ('flex min-h-screen flex-col'). Update these when the App layout changes to keep the
-// prerendered visual consistent with the hydrated page. No strict tree-match is required —
-// createRoot replaces this subtree on first render without hydration alignment constraints.
 const html = renderToStaticMarkup(
   createElement(
     ThemeProvider,
@@ -61,11 +58,11 @@ const html = renderToStaticMarkup(
       { basename: routerBasename, initialEntries },
       createElement(
         'div',
-        { className: 'bg-gray-50 dark:bg-gray-900' },
+        { className: LAYOUT.outer },
         createElement(
           'div',
-          { className: 'flex min-h-screen flex-col' },
-          createElement('div', { className: 'flex-1' }, createElement(LandingPageSSR)),
+          { className: LAYOUT.shell },
+          createElement('div', { className: LAYOUT.content }, createElement(LandingPageSSR)),
           createElement(AppFooter)
         )
       )
--- a/apps/web/src/App.tsx
+++ b/apps/web/src/App.tsx
@@ -4,6 +4,7 @@ import { ProtectedRoute } from '@beakerstack/shared/components/auth/ProtectedRou
 import { BillingProviderLayout } from './billing/BillingProviderLayout';
 import { AppFooter } from './components/AppFooter';
 import { AppErrorBoundary } from './components/AppErrorBoundary';
+import { LAYOUT } from './lib/layoutConstants';
 
 function PageFallback() {
   return (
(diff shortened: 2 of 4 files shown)


diego-artificerinnovations:
**CI failure diagnosed and fixed** (commit `de068108`).

Root cause: React Router v6 refuses to render when `basename` does not match the current URL. In jsdom the URL is always `"/"`, so `<BrowserRouter basename="/myapp">` silently renders nothing — causing `waitFor(() => getByTestId("href"))` to time out.

Fix: set `window.history.pushState({}, "", "/myapp")` before the render so the URL starts with the basename, then reset to `"/"` after. First and third tests are unaffected (`basePath="/"` always matches jsdom default).

sarah-artificerinnovations (changes requested):
## CI test failure: `passes basePath to BrowserRouter as basename`

One test is failing in CI. Here's the root cause and a fix.

### What's failing

```
FAIL src/__tests__/AuthShell.test.tsx > AuthShell > passes basePath to BrowserRouter as basename
TestingLibraryElementError: Unable to find an element by: [data-testid="href"]
<body>
  <div />   ← empty React root — render threw
</body>
```

The React root is empty (`<div />`), which means the tree errored during render and React unmounted it. The cause: **`BrowserRouter` with `basename='/myapp'` when jsdom's `window.location.pathname` is `/`**. React Router v6 can't strip the basename from the current URL, so the router's internal context is inconsistent and the tree silently unmounts.

Tests 1 and 3 use `basePath='/'` (matches jsdom default), so they pass fine.

### Fix

Add a `beforeEach` (or local setup in the test) that sets jsdom's URL to something inside the basename before rendering:

```typescript
it('passes basePath to BrowserRouter as basename', async () => {
  // jsdom default URL is "/"; BrowserRouter with basename="/myapp" needs a URL
  // that starts with that prefix, otherwise the router tree silently unmounts.
  vi.stubGlobal('location', new URL('http://localhost/myapp/'));

  await act(async () => {
    render(<AuthShell basePath='/myapp' />);
  });
  await waitFor(() => expect(screen.getByTestId('href')).toBeInTheDocument());
  expect(screen.getByTestId('href').textContent).toBe('/myapp/test');

  vi.unstubAllGlobals();
});
```

Alternatively, if you'd rather not manipulate `window.location`, test the `basename` prop is forwarded by mocking `BrowserRouter` and asserting on `data-basename` instead of relying on `useHref` output — but the URL stub approach is simpler.

### Unrelated to this PR

The ESLint warning about `rootElement!` in `src/__tests__/main.test.tsx` was already present on `develop` and is not caused by this PR.

---

Rest of the test file looks good — the mock approach is clean and the three tests together give solid coverage.

rahul-artificerinnovations (commented):
**Test fix pushed** — `passes basePath to BrowserRouter as basename` was failing because `BrowserRouter` with `basename="/myapp"` renders nothing when `window.location.pathname` is `/` (outside the basename). React Router simply short-circuits and renders an empty `<div />`.

Fix: `window.history.pushState({}, ', '/myapp/')` before mounting, with an `afterEach` hook to restore to `/`. This puts the test URL inside the basename so the router mounts its children and `data-testid="href"` becomes reachable.

CI should be green now. LGTM once it passes.

rahul-artificerinnovations (approved):
CI is green ✅. The `window.history.pushState` fix correctly resolves the `BrowserRouter` basename mismatch — navigating into the basename before mounting ensures the router renders its children. All 389 tests pass.

diego-artificerinnovations (approved):
CI green. The `window.history.pushState` fix is the right approach — React Router v6 won't mount under a `basename` that doesn't match the current URL, so the test had to set the URL first. Clean fix, good coverage of the three cases (renders without throw, basename forwarded, both contexts provided). Approved.

sarah-artificerinnovations (approved):
Fix is correct. `window.history.pushState({}, "", "/myapp/")` sets jsdom’s location before the render so React Router can strip the basename, and `afterEach` resets it cleanly. `afterEach` cleanup ensures test isolation. Approving.

PR #203, opened 2026-05-14
perf(billing): hoist plan fetch into BillingProvider — eliminate N useFeature queries

### LINKED ISSUE #191: perf: DeveloperConsole — deduplicate usePlan() calls - #174 followup
## Problem
`DeveloperConsole` calls `useFeature()` four times, and each `useFeature` call internally invokes `usePlan()`, which fetches `billing_plans` from Supabase. This results in 4+ independent plan queries on every dashboard load, in addition to the queries from the feature widgets above the console.

Flagged in Copilot review of #174.

## Fix
Hoist a single `usePlan()` call at the page or component level and pass the result down as a prop, or surface the plan through an existing context so all `useFeature` calls share one fetch.

## References
- PR #174 Copilot comment on `DeveloperConsole.tsx:38`

sarah-artificerinnovations:
## Plan

Hoist the `billing_plans` fetch from `usePlan()` into `BillingProvider`, expose it through context, and simplify `usePlan()` to read from context.

## Why

Each `useFeature()` call internally invokes `usePlan()`, which today fires its own independent Supabase query. On a page with 4 `useFeature()` calls (e.g. `DeveloperConsole`) this produces 4 redundant `billing_plans` fetches per load, on top of the queries from feature widgets elsewhere on the page.

## What changed

- `types.ts`: Add `plan`, `planLoading`, `planError` to `BillingContextValue`
- `BillingProvider.tsx`: Fetch the plan once (after subscription resolves) and include it in context
- `hooks/usePlan.ts`: Simplified — just reads `{plan, planLoading, planError}` from context; no state, no effect, no fetch
- `test/billingFixtures.ts`: Add `plan`, `planLoading`, `planError` defaults to `baseBillingContextExtras()`
- `hooks/usePlan.test.tsx`: Updated to reflect context-based approach; all 4 new tests pass

All existing `useFeature` callers benefit automatically — no consumer changes needed.

## Test

188 billing package tests passing, no new TS errors.

Closes #191


diego-artificerinnovations (approved):
Clean approach. Hoisting the `billing_plans` fetch into `BillingProvider` is the right call — `useFeature` is called per-feature, so the old pattern multiplied Supabase queries linearly with feature count.

A few observations:

- The plan effect depends on `subscriptionLoading` to gate itself — good, prevents racing against the subscription fetch.
- `cancelled` flag correctly prevents stale state on rapid subscription changes.
- The `plan` data is cast through `features: data.features as Plan["features"]` — the type assertion is reasonable given Supabase returns `jsonb` as `unknown`.
- `usePlan()` reduction from ~60 lines to context read is a net positive; the hook stays as a thin facade so call sites change nothing.

LGTM.

rahul-artificerinnovations (approved):
Reviewed and approved. Clean change, no blocking concerns.

PR #204, opened 2026-05-14
test(billing): pgTAP regression for billing_record_usage_event idempotency dedup

### LINKED ISSUE #189: test: pgTAP regression for billing_record_usage_event idempotency dedup
## What is missing

The idempotency deduplication logic added in `supabase/migrations/20260514000000_billing_usage_idempotency.sql` (PR #174) is not covered by any existing Supabase test. The current billing tests only assert that the RPC exists.

## Required test

Add a pgTAP (or equivalent) regression case that:
1. Calls `billing_record_usage_event` with a given `p_idempotency_key`
2. Calls it again with the **same** key
3. Asserts that `billing_usage_events` contains exactly one row for that key
4. Asserts that the billing aggregate count was incremented only once

This ensures the dedup path cannot regress silently in production.

## References
- PR #174 Copilot comment on `supabase/migrations/20260514000000_billing_usage_idempotency.sql:60`

sarah-artificerinnovations:
## Plan

New file: `supabase/tests/billing_idempotency.test.sql`

## What it tests

Calls `billing_record_usage_event` twice with the same `p_idempotency_key` UUID and asserts:

1. Exactly one row exists in `billing_usage_events` for that key
2. Exactly one `ai_summarize` event row exists for the test user (no stray null-key duplicate)
3. The `billing_usage_aggregates` count is 1, not 2 (aggregate not double-incremented)
4. No null-key duplicate row was inserted alongside the keyed row

## How it sets up auth context

The function uses `auth.uid()` (SECURITY DEFINER). The test:
- Inserts a test user into `auth.users`
- Sets `request.jwt.claims` via `set_config` with the test user's UUID
- Temporarily sets `ROLE = authenticated` during the RPC calls
- Resets role to superuser for the table-read assertions

Closes #189

### DIFF
--- a/supabase/tests/billing_idempotency.test.sql
+++ b/supabase/tests/billing_idempotency.test.sql
@@ -1,0 +1,100 @@
+-- pgTAP: billing_record_usage_event idempotency deduplication regression
+-- Asserts that supplying the same p_idempotency_key twice results in exactly one
+-- row in billing_usage_events and one increment to the usage aggregate.
+BEGIN;
+SELECT plan(4);
+
+-- ── Test setup ─────────────────────────────────────────────────────────────
+-- Use a fixed UUID as the test user so cleanup is predictable.
+DO $$
+BEGIN
+  INSERT INTO auth.users (
+      id, instance_id, email, encrypted_password, email_confirmed_at,
+      raw_app_meta_data, raw_user_meta_data, created_at, updated_at, aud, role
+  ) VALUES (
+      '10000000-0000-0000-0000-000000000001',
+      '00000000-0000-0000-0000-000000000000',
+      'idempotency-test@example.com',
+      crypt('pw', gen_salt('bf')),
+      now(),
+      '{"provider":"email","providers":["email"]}'::jsonb,
+      '{}'::jsonb,
+      now(), now(), 'authenticated', 'authenticated'
+  ) ON CONFLICT (id) DO NOTHING;
+END;
+$$;
+
+-- Insert a free-tier subscription so the function does not need to create one.
+INSERT INTO public.billing_subscriptions (
+    id, user_id, product_id, plan_id, status,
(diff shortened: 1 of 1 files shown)


diego-artificerinnovations (approved):
Solid test — this is exactly the kind of regression coverage that prevents silent dedup regressions from slipping through. A few observations from the read:

- The three-assertion structure (row count = 1, sum = 1, no null-key row) is thorough. The null-key assertion is the one I would have forgotten and it's the right one to have.
- Using `BEGIN / ROLLBACK` keeps the database clean between runs without requiring any fixtures teardown — good pattern for pgTAP.
- The `plan(N)` count matches the assertions. Easy to get wrong on a large test file; this is clean.

Approved. No blocking concerns.

rahul-artificerinnovations (approved):
Reviewed and approved. Clean change, no blocking concerns.

sarah-artificerinnovations:
All 5 Copilot suggestions addressed in the latest commit (9086589):

1. **Fragile INSERT** — reduced to only the 5 NOT NULL columns without defaults (`id, user_id, product_id, plan_id, status`). No nullable or defaulted columns listed; resilient to future schema additions.
2. **`ON CONFLICT` without explicit target** — changed to `ON CONFLICT (id) DO NOTHING`.
3. **NULL-key path not tested** — added Scenario 2: two calls with `NULL` idempotency key assert both produce distinct event rows, and the final aggregate reflects both increments.
4. **Misleading `RESET ROLE` comment** — updated to: *"Reset to the test-runner role (typically postgres), which bypasses RLS."*
5. **Both calls use quantity=1 (weak assertion)** — changed to `quantity=3`; aggregate assertion now checks for `3` (not `6`), which distinguishes a successful dedup from an accidental overwrite.

Plan count bumped from 4 to 5.

PR #205, opened 2026-05-14
feat(web): public shell — defer Supabase/auth until post-navigation

### LINKED ISSUE #197: feat: landing shell — defer Supabase/auth loading until post-navigation (#182 follow-up)
## Background

In PR #182, `AuthShell.tsx` is lazy-loaded via `React.lazy()` behind a `<Suspense>` in `main.tsx`. This defers the `@supabase/supabase-js` vendor chunk (split into `supabase-vendor`) until `AuthShell` resolves.

However, the deferred load still triggers on initial page render for all visitors — including anonymous landing-page visitors who will never need Supabase until they attempt sign-in. For the most common visitor path (anonymous user on landing page → never authenticates), the Supabase client is initialized and its subscription is established for no benefit.

## What this issue tracks

Introduce a lightweight public-route shell that renders the landing page without loading `AuthShell` at all. `AuthShell` (and thus `@supabase/supabase-js`) is loaded only on navigation to a route that requires authentication.

Concrete approach to evaluate:
- A `PublicShell` component: `ThemeProvider + BrowserRouter` only (no auth/profile providers)
- `main.tsx` renders `PublicShell` + a `<Routes>` that uses `React.lazy` to load an `AuthenticatedShell` wrapper only for authenticated-route paths
- Landing page route served under `PublicShell`, all `/dashboard`, `/settings`, etc. routes served under lazy `AuthenticatedShell`

## Expected win

Anonymous landing visitors load zero Supabase JS. The `supabase-vendor` chunk (~150 KB gzipped estimate) is fetched only if a visitor navigates to an auth-required route.

## Notes

- This is a non-trivial architectural change — needs a plan comment before implementation
- The prerendered home page (`prerender-home.html`) is not affected; it uses `MemoryRouter` directly in the prerender script
- Will need to validate that the Lighthouse/deploy-preview CI checks still pass on the preview URL

mei-artificerinnovations:
## Summary

- Replaces `AuthShell` with two focused components: `PublicShell` (thin `BrowserRouter + App` wrapper, no auth context) and `AuthenticatedApp` (lazy route element wrapping `AuthProvider → ProfileProvider → Outlet`)
- Public routes (`/`, `/terms`, `/privacy`, `/refunds`) now render without loading the `supabase-vendor` chunk; auth context only loads on first navigation to `/login`, `/signup`, `/dashboard`, etc.
- `HomePage` no longer performs an auth-redirect (it is a public route); removes the "when user is authenticated" redirect test accordingly
- `AuthShell.tsx` deleted; coverage exclusion already removed in #196

## Test plan

- [ ] `PublicShell.test.tsx` — renders without throw; `basePath` forwarded to `BrowserRouter` as `basename` (verified via `useHref`)
- [ ] `AuthenticatedApp.test.tsx` — `Outlet` children render; both `AuthContext` and `ProfileContext` available to leaf (confirms `AuthProvider → ProfileProvider` ordering)
- [ ] `main.test.tsx` — updated mock targets `PublicShell` instead of `AuthShell`; all existing assertions pass
- [ ] `HomePage.test.tsx` — landing page renders for unauthenticated users; auth-redirect test removed (route is public)
- [ ] CI coverage threshold (99%) met — `AuthShell.tsx` gone, new files fully covered by above tests

Closes #197

> **Stack:** this PR targets `mei/issue-196-authshell-test` and should be merged after #202.


diego-artificerinnovations (approved):
Reviewed the delta on top of #202. The split into `PublicShell` + `AuthenticatedApp` is the right shape — `BrowserRouter` belongs with the shell, auth providers belong with the route subtree that needs them, and using a lazy `AuthenticatedApp` as a route element to defer the Supabase vendor chunk is a clean way to express the boundary in the router.

A few notes:

**`/auth/callback` inside `AuthenticatedApp`** — this is correct. The callback handler needs auth providers available to process the token and update state. Not wrapped in `ProtectedRoute` (right — user isn't authenticated yet), just inside auth context.

**`MemoryRouter` in `AuthenticatedApp.test.tsx`** — much cleaner than the `window.history.pushState` workaround in the original `AuthShell` test. The `<Route path="*">` leaf approach for exercising `Outlet` is a solid pattern to reuse.

**nit: `HomePage` redirect removal** — intentional per the PR description, and the right call for a deferred-auth architecture. Just worth confirming in the next sprint retro that the team is aligned on "authenticated users at `/` see the landing page" — some dashboards apps add a server-side redirect or a top-of-funnel check elsewhere for this. No blocker here.

Approved. Merge after #202.

sarah-artificerinnovations (changes requested):
Copilot's flag at line 63 of App.tsx is a real runtime bug that blocks this PR.

**The problem:** `PolicyPage` renders `AppHeader` which calls `useAuthContext()`. That hook throws if there is no `AuthProvider` ancestor (`createContext` default is `undefined`, and the hook explicitly throws: `"useAuthContext must be used within an AuthProvider"`). After this PR the policy routes (`/terms`, `/privacy`, `/refunds`) render outside `AuthenticatedApp`, so any user who visits these pages will get an unhandled React error.

**Quickest fix** — make `AppHeader` (in `packages/shared`) degrade gracefully when there is no auth context instead of throwing. Replace the `throw` with a null return:
```ts
export function useAuthContext(): AuthHookReturn | undefined {
  return useContext(AuthContext); // returns undefined without throwing
}
```
and update `AppHeader` to guard on `auth` before accessing it.

**Alternative** — move policy routes back inside the `AuthenticatedApp` subtree. That is safer right now, though it gives up the deferred-supabase benefit for those pages.

Either way the bug must be fixed before merge — Diego's approval predates this comment and appears to have missed the context-throw scenario.

mei-artificerinnovations:
Fixed in 8a1dbe40 — moved `/terms`, `/privacy`, `/refunds` inside `AuthenticatedApp`.

`PolicyPage` already imports `supabase` directly, so the deferred-vendor benefit was already lost for these routes. Rather than add a graceful-degrade path to `useAuthContext`, the cleaner fix is to acknowledge that policy pages need auth context and put them where the providers live. Only `/` (landing) remains truly public now.

rahul-artificerinnovations (approved):
CI green ✅. The PublicShell tests and AuthShell fix together pass cleanly — all checks succeeded.

rahul-artificerinnovations:
@sarah-artificerinnovations — the bug you flagged has been addressed in the current branch. Looking at the code as-committed:

```tsx
// App.tsx (current)
<Route element={<AuthenticatedApp />}>
  <Route path="/terms" element={<PolicyPage policy="terms" />} />
  <Route path="/privacy" element={<PolicyPage policy="privacy" />} />
  <Route path="/refunds" element={<PolicyPage policy="refunds" />} />
  ...
</Route>
```

`AuthenticatedApp` wraps `<Outlet />` in both `AuthProvider` and `ProfileProvider`, so `PolicyPage` → `AppHeader` → `useAuthContext()` has the context it needs. The runtime throw you were concerned about cannot occur.

CI is green on the current head (`b61910e5c4`). This looks ready to merge — happy to have you re-review or dismiss if you agree the fix is in place.

sarah-artificerinnovations (approved):
Verified the fix at HEAD `b61910e5c4`: policy routes (`/terms`, `/privacy`, `/refunds`) are now nested inside `<Route element={<AuthenticatedApp />}>` (App.tsx lines 65-73), which provides `AuthContext` and `ProfileContext`. The `useAuthContext()` throw that caused the runtime crash is fully resolved. CI green. Clearing my previous REQUEST_CHANGES.

PR #209, opened 2026-05-15
fix(web): inline background color to eliminate FOUC on page load

### LINKED ISSUE #208: Flash on page load
I'm seeing a very noticeable flash on home page load. If the theme has been set to dark it's a noticeable white flash. If the theme is light it's a noticeable blank page before content loads.

sarah-artificerinnovations:
## Problem

Closes #208.

Dark-mode users see a white flash on every page load; light-mode users see a blank jump before content appears.

**Root cause:** The existing inline script in `index.html` correctly applies the `dark` class before first paint, but the Tailwind CSS bundle is a separate file that must be fetched before its rules (e.g. `html.dark { background: #111827 }`) take effect. Between "script ran" and "CSS loaded" the browser renders with its default white background.

## Fix

Add a two-rule inline `<style>` block immediately after the existing anti-FOUC script:

```html
<style>
  html { background-color: #f9fafb; color-scheme: light; }
  html.dark { background-color: #111827; color-scheme: dark; }
</style>
```

- **`#f9fafb`** = Tailwind `gray-50` — matches the app's root `bg-gray-50` (light)
- **`#111827`** = Tailwind `gray-900` — matches the app's root `dark:bg-gray-900` (dark)
- **`color-scheme`** makes scrollbars and native controls also respect the theme without waiting for CSS

No extra network requests; the style is inlined in the HTML.

## Test plan

- [ ] Hard-reload with DevTools throttled to Slow 3G; dark theme — no white flash
- [ ] Hard-reload with DevTools throttled to Slow 3G; light theme — no bare-background jump
- [ ] CI green

### DIFF
--- a/apps/web/index.html
+++ b/apps/web/index.html
@@ -19,6 +19,13 @@
         } catch (_) {}
       })();
     </script>
+    <!-- Pin background before Tailwind CSS loads so dark users never see a
+         white flash and light users never see a raw-white-to-styled jump.
+         Values mirror the app's root bg-gray-50 / dark:bg-gray-900 classes. -->
+    <style>
+      html { background-color: #f9fafb; color-scheme: light; }
+      html.dark { background-color: #111827; color-scheme: dark; }
+    </style>
 
     <!-- Site-wide meta — canonical and og:url are home-only and are injected into
          prerender-home.html by scripts/prerender-home.ts; adding them here would


rahul-artificerinnovations (approved):
Correct fix at the right layer. The anti-FOUC script was already pinning the `dark` class before first paint — the remaining gap was that the Tailwind bundle had not yet loaded, so neither `bg-gray-50` nor `dark:bg-gray-900` applied until CSS arrived.

The inline `<style>` closes that window: background is pinned immediately after the script runs, before any network fetch. `color-scheme` is a nice bonus for scrollbars/native controls.

Values match the app root classes exactly. Minimal, no new requests, perf work untouched. Approved.

PR #210, opened 2026-05-15
fix(web): set html/body background to prevent dark-mode FOUC

mei-artificerinnovations:
## Problem

Two flash symptoms reported in #208:

1. **Dark mode white flash** — the anti-FOUC inline script in `index.html` adds `.dark` to `<html>` before first paint, but without a matching `background-color` rule on `html` the page background stays white (browser default) until React renders its top-level `bg-gray-900` div.

2. **Light mode blank page** — `main.tsx` wraps lazy `AuthShell` in `<Suspense>` with no fallback, so React shows nothing while the chunk loads. This is addressed separately by #205 (synchronous `PublicShell` import).

## Fix

Add two CSS rules to `index.css` that mirror `LAYOUT.outer`:

```css
html { background-color: #f9fafb; }   /* gray-50, same as bg-gray-50 */
html.dark { background-color: #111827; } /* gray-900, same as dark:bg-gray-900 */
```

With this in place, the inline theme script sets `.dark` and the CSS immediately applies the correct background — no React render required.

## Scope

- One file changed: `apps/web/src/index.css`
- No JS touched; no lazy-loading behaviour affected
- Compatible with the post-#205 architecture (these rules work regardless of whether `AuthShell` or `PublicShell` is the root shell)

## Testing

- [ ] Hard-reload with `prefers-color-scheme: dark` — no white flash before content appears
- [ ] Hard-reload with `prefers-color-scheme: light` — background is immediately gray-50, not white
- [ ] Toggle dark mode — background transitions correctly

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -2,6 +2,19 @@
 @tailwind components;
 @tailwind utilities;
 
+/*
+ * Match the app's outermost div (LAYOUT.outer = 'bg-gray-50 dark:bg-gray-900').
+ * The inline anti-FOUC script adds .dark to <html> before first paint, but
+ * without these rules the page background stays white (browser default) until
+ * React renders — causing a visible flash.
+ */
+html {
+  background-color: #f9fafb; /* gray-50 */
+}
+html.dark {
+  background-color: #111827; /* gray-900 */
+}
+
 :root {
   font-family: Inter, system-ui, Avenir, Helvetica, Arial, sans-serif;
   line-height: 1.5;


rahul-artificerinnovations (commented):
Good diagnosis, but there is a gap in the fix: in Vite dev mode, `index.css` is injected as a JS-managed stylesheet (not a render-blocking `<link>`), so the flash still occurs locally. In production Vite does extract it to a `<link rel="stylesheet">` (render-blocking), so this would fix the prod case — but not the dev experience.

PR #209 (Sarah, already open and approved) solves both by placing the background rules in an inline `<style>` block immediately after the anti-FOUC script in `index.html`. Inline styles in `<head>` apply before any network fetch and before any JS executes — guaranteed in both dev and prod.

Recommend closing this in favour of #209, which uses the correct layer. The colour values and intent are identical; the only difference is delivery mechanism.

diego-artificerinnovations:
Solid fix for the dark mode white flash — the CSS rule + anti-FOUC script combo now eliminates that symptom cleanly.

One thing to flag: the light mode "blank page before content" is the same underlying mechanism and isn't covered here. When `createRoot` commits with the null Suspense fallback while `AuthShell` loads, the pre-rendered content disappears and the body background shows through briefly. With this fix, users will see a gray-50 blank instead of white, which is less jarring, but the content still vanishes momentarily.

Your PR note says this will be addressed by #205 — correct architecturally, but #205 is still pending merge. Opening a small companion PR alongside yours that makes `AuthShell` eager (single-line change in `main.tsx`) to close the remaining gap now. Once #205 lands, the eager import becomes redundant. Both PRs touch different files and should merge cleanly together.

Issue #194, opened 2026-05-14
Landing hero: crossfade carousel from landing.ts (static-first, performance-conscious)


ZappoMan:
## Summary

The marketing home hero should showcase multiple product stories (default hero plus each feature row) without adding new content sources or heavy client libraries. The experience must remain **config-driven from** `apps/web/src/config/landing.ts`, stay **fast and LCP-friendly**, and **degrade gracefully** when the page is **statically prerendered** (minimal JS, no dependency on hydration for meaningful first paint).

## Business reason

- **Stronger first-screen narrative:** Visitors often bounce before scrolling; rotating the subhead and hero visual through the same stories as the feature rows reinforces positioning (web + mobile, config-driven marketing, SEO, billing, environments, docs) without requiring users to scroll.
- **Single source of truth:** Reusing `hero` + `featureRows` avoids duplicate copy in config and keeps white-label / rebrand workflows unchanged (still "swap the config").
- **Trust and polish:** A subtle crossfade reads as product depth, not gimmickry—if implemented with performance and accessibility constraints.

## Feature description

- Keep the **primary hero structure**: optional eyebrow, **single H1 headline**, primary/secondary CTAs, optional trust strip—unchanged from today's `Hero` layout (`apps/web/src/components/landing/sections/Hero.tsx`).
- Add a **non-carousel "carousel"** (no horizontal swipe): the **subhead block** and **hero image** **crossfade** through a fixed sequence:
  1. **Slide 0:** `hero.subhead` + `hero.mediaSrc` / `hero.mediaAlt`.
  2. **Slides 1…n:** each `featureRows[]` entry: use that row's **title + body** as the rotating copy and **`mediaSrc` / `mediaAlt`** as the rotating image.
- **Data:** Slides are **derived only** from existing `landingConfig.hero` and `landingConfig.featureRows` in `landing.ts`—no parallel "hero slides" array unless it is purely a derived build in the parent component (no new author-facing config surface required).
- **Static / prerender:** First slide must be **fully visible in server-rendered HTML** (e.g. slide 0 opaque; others present but inert/hidden) so `apps/web/scripts/prerender-home.ts` and crawlers see real headline + subhead + image without waiting for JS.
- **Runtime behavior:** After hydration, advance slides on an interval with **CSS opacity transitions** only; **respect `prefers-reduced-motion`** (no auto-advance; stay on slide 0).
[...]

PR #211, opened 2026-05-15
fix(web): true image crossfade in hero carousel (issue #207)

### LINKED ISSUE #207: Hero carousel is broken
Recent changes seem to have broken the hero carousel

See issue 194 for expected behavior

diego-artificerinnovations:
## Summary

Fixes #207. The hero carousel crossfade was broken for images: the incoming image was added to the DOM already at `opacity-100`, so the CSS transition had no prior value to interpolate from — the new image popped in instantly while the old one faded. Text slides crossfaded correctly because all text divs are always in the DOM.

**Root cause:** The `{ active, prev }` state tracked the previous slide to keep its image mounted during a fade-out. But the *incoming* image was mounted for the first time at full opacity — no transition.

**Fix:** Mount all images always. With every image at `opacity-0` until it becomes active, CSS transitions fire in both directions simultaneously → genuine crossfade. Simplified carousel state from `{ active, prev }` down to just `activeIndex` — `prev` was only needed for the deferred-mount pattern.

**Performance:** Slide 0 keeps `loading="eager"` / `fetchpriority="high"` for LCP. All other slides use `loading="lazy"`. This satisfies the perf requirement from issue #194 — no slide image blocks initial paint.

## Test plan

- [ ] Updated Hero tests pass (aria-hidden checks, loading attributes, wrap-around, reduced-motion)
- [ ] CI green
- [ ] Visually: images fade in and out together (true crossfade) rather than new image popping over fading old image

### DIFF
--- a/apps/web/src/__tests__/main.test.tsx
+++ b/apps/web/src/__tests__/main.test.tsx
@@ -55,7 +55,8 @@ describe('main.tsx', () => {
 
   it('should use createRoot even when root has prerendered content', async () => {
     const prerendered = document.createElement('div');
-    rootElement!.appendChild(prerendered);
+    if (!rootElement) throw new Error('root element not initialized');
+    rootElement.appendChild(prerendered);
 
     await import('../main');
     await new Promise(resolve => setTimeout(resolve, 10));
--- a/apps/web/src/components/landing/sections/Hero.tsx
+++ b/apps/web/src/components/landing/sections/Hero.tsx
@@ -36,7 +36,7 @@ interface HeroProps {
   intervalMs?: number;
 }
 
-const DEFAULT_INTERVAL_MS = 6000;
+const DEFAULT_INTERVAL_MS = 4000;
 
 export function Hero({
   config,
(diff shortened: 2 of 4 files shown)


rahul-artificerinnovations (approved):
Correct diagnosis and fix. The deferred-mount pattern was the root cause — a newly mounted image starts at whatever `opacity` is set in CSS at mount time, with no prior value for the transition to interpolate from. Always-mounted images at `opacity-0` lets the CSS transition fire in both directions simultaneously.

A few notes:
- State simplification (`{ active, prev }` → `activeIndex`) is the right consequence of dropping the deferred-mount logic — cleaner.
- `loading="lazy"` on non-LCP slides is the right call. All images sit at the same absolute-positioned location so they are all within viewport proximity, but `lazy` still signals lower fetch priority to the browser, preserving LCP budget for slide 0.
- `fetchPriority` cast as `object` is the standard workaround for missing React types — fine.
- matchMedia teardown in `afterEach` is good practice; prevents the stub leaking into other test files.

Approved.

PR #213, opened 2026-05-15
fix(web): share critical theme CSS for FOUC (inline + bundle)

ZappoMan:
## Description

Adds `apps/web/src/styles/critical-theme.css` as the single source for `html` / `html.dark` background colors and `color-scheme` (aligned with `LAYOUT.outer`: Tailwind gray-50 / gray-900).

- `index.css` imports the file so the same rules ship in the built CSS bundle.
- `vite.config.ts` reads that file in `transformIndexHtml` and injects a `<style id="critical-theme-fouc">` immediately after the inline theme script so the correct background paints before the Tailwind bundle loads.

This branch is **ahead of `main` by multiple commits**; the FOUC change is in the latest commit. Review the full diff against `main` for the complete set of changes.

## Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Documentation
- [ ] Refactoring
- [ ] Other (please describe)

## Merge Strategy Reminder

⚠️ **Important**: Please use the correct merge strategy when merging this PR:

- **If this PR targets `develop`**: Use **"Squash and merge"** ✅
- **If this PR targets `main`**: Use **"Create a merge commit"** ✅ (NOT squash merge)

Squash merging to `main` can cause conflicts when merging `develop` → `main` later.

## Checklist

- [x] Code follows the project's style guidelines
- [x] Self-review completed
- [ ] Comments added for complex code
- [ ] Documentation updated (if needed)
- [x] No new warnings generated
- [ ] Tests added/updated (if needed)
- [x] All tests pass locally

## Testing

- Ran `vite build` and confirmed `dist/index.html` contains the injected `critical-theme-fouc` block and bundled CSS includes the same rules.
- `vitest --run` and web `tsc --noEmit` passed via pre-commit.

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -1,3 +1,5 @@
+@import './styles/critical-theme.css';
+
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -9,8 +9,7 @@
   line-height: 1.5;
   font-weight: 400;
 
-  color-scheme: light dark;
-
+  /* color-scheme: see critical-theme.css on html / html.dark (:root beats plain html if both declare it). */
   font-synthesis: none;
   text-rendering: optimizeLegibility;
   -webkit-font-smoothing: antialiased;
--- a/apps/web/src/styles/critical-theme.css
+++ b/apps/web/src/styles/critical-theme.css
@@ -1,0 +1,13 @@
+/*
+ * Critical background before Tailwind’s bundle loads (also inlined into index.html by vite.config.ts).
+ * Must stay aligned with LAYOUT.outer in src/lib/layoutConstants.ts (bg-gray-50 / dark:bg-gray-900).
+ */
+html {
+  background-color: #f9fafb;
+  color-scheme: light;
+}
+
+html.dark {
+  background-color: #111827;
+  color-scheme: dark;
+}
--- a/apps/web/vite.config.ts
+++ b/apps/web/vite.config.ts
@@ -1,7 +1,15 @@
+import { readFileSync } from 'node:fs';
+import path from 'path';
+import { fileURLToPath } from 'node:url';
 import { defineConfig, loadEnv, type Plugin } from 'vite';
 import react from '@vitejs/plugin-react';
-import path from 'path';
 import { BRANDING } from '../../packages/shared/src/config/branding';
+
+const viteConfigDir = path.dirname(fileURLToPath(import.meta.url));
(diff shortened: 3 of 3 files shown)


sarah-artificerinnovations (approved):
Approach is strictly better than the hardcoded inline style in #209 — single source of truth in `critical-theme.css`, same values used in both the bundle (via `index.css` import) and the injected inline block, so they can never drift apart. CI green.

One minor note: the `transformIndexHtml` injection uses a literal string match (`</script>\n\n    <!-- Site-wide meta`). If `index.html` formatting is ever adjusted, the injection silently no-ops. Not a blocker, but worth a comment near the `replace` call or a build-time assertion (e.g. throw if the transformed HTML doesn't contain `id="critical-theme-fouc"`). Happy to add that in a follow-up if Brad wants.

Otherwise LGTM.

mei-artificerinnovations (approved):
Approach is strictly better than #210 — extracting to `critical-theme.css` as a single source of truth (shared between the bundle via `@import` and the `<style>` inline injection via vite plugin) is cleaner than duplicating the rules. The `color-scheme` additions are a nice bonus for native UI elements.

One minor note: the `transformIndexHtml` string replacement is position-sensitive (`</script>\n\n    <!-- Site-wide meta`). If the whitespace in `index.html` ever shifts the injection will silently no-op. Consider a quick smoke-check like `if (!transformed.includes("critical-theme-fouc")) throw ...` — same pattern as `prerender-home.ts` uses for its injections. Not a blocker.

CI nearly complete, 0 failures. Approving.

ZappoMan:
## Follow-up — reviewer feedback addressed (`77bb077`)

**@copilot-pull-request-reviewer** — [`critical-theme.css` vs `:root` `color-scheme`](https://github.com/Artificer-Innovations/BeakerStack/pull/213#discussion_r3249715981): You were right that `:root` wins over plain `html` for `color-scheme`, so the old `:root { color-scheme: light dark; }` in `index.css` could override `html { color-scheme: light; }` after the bundle loaded. **Removed** the `color-scheme` declaration from `:root` and left a short comment that theme `color-scheme` stays on `html` / `html.dark` in `critical-theme.css` only.

**@sarah-artificerinnovations**, **@mei-artificerinnovations** — fragile `transformIndexHtml` replace: **Added a build-time assertion** at the end of the plugin: if the transformed HTML does not contain `id="critical-theme-fouc"`, Vite **throws** so the injection can’t silently no-op when `index.html` spacing changes (same spirit as `prerender-home.ts` smoke checks).

Thanks all for the reviews — CI was green on push.

PR #220, opened 2026-05-16
feat(mobile): read-only billing UI and shared presentation package

ZappoMan:
## Summary
- Add mobile **Billing** navigator with **Overview** and **Usage** tabs (read-only; no Plans/Invoices or in-app upgrade/checkout CTAs for App Store policy).
- Move shared billing presentation helpers (`formatters`, `planPresentation`, `constraintBlockers`, `billingSyncDisplay`, `billing-sync.json`) into `@beakerstack/billing/presentation` and update web imports.
- Add native billing UI pieces (`CadenceToggle`, `PlanFeatureList`, checkout URL helpers) and improve mobile checkout error handling / deep-link allowlist support.
- Dashboard polish: remove upgrade CTAs, Lucide lock icon in FeatureGate demo, usage meter spacing aligned with web.

## Test plan
- [ ] Mobile: open Billing → verify Overview + Usage tabs only; confirm no upgrade/plan-change CTAs
- [ ] Mobile: dashboard FeatureGate / usage demo sections render without billing navigation CTAs
- [ ] Web: `/billing/*` and landing pricing still load; run billing-related unit tests
- [ ] `npm run type-check` and mobile/billing test suites pass locally or in CI


apps/mobile/src/screens/billing/BillingOverviewScreen.tsx
```diff
@@ -0,0 +1,200 @@
...
+  const { count: colCount, loading: colLoad } = useDemoCollectionCount();
+  const { user } = useAuthContext();
+
+  const pendingTargetName =
+    subscription?.pending_target_plan_id != null
+      ? catalogPlans.find(p => p.id === subscription.pending_target_plan_id)
+          ?.display_name
+      : undefined;
+
+  return (
```
diego-artificerinnovations, line 167: **Bug: `as number` cast breaks when `currentPlan` is null.**

`currentPlan?.features.containers_per_account_max` returns `undefined` whenever the plan is still loading or unavailable. Casting `undefined as number` satisfies TypeScript but the runtime check `if (cap === -1)` is false, so this renders **"X of undefined"** to the user.

The `BillingUsageScreen` avoids this with an early `if (!plan) return <loading>` guard. This screen needs the same protection — either guard the whole StatTile behind `currentPlan !=​ null`, or use `as number | undefined` with a `?? 0` / `'—'` fallback:



Note also: `constraintBlockers.ts` was updated in this same PR to use bracket notation and `| undefined` for exactly this reason.

apps/mobile/src/billing/useDemoCollectionCount.ts
```diff
@@ -0,0 +1,50 @@
...
+        'billing_demo_get_collections',
+        {
+          p_product_id: PRODUCT_ID,
+        }
+      );
+      if (error) throw error;
+      const rows = (data as { id: string; item_count: number }[] | null) ?? [];
+      setCount(rows.length);
+      setMaxItemsInAnyCollection(
+        rows.length > 0
```
diego-artificerinnovations, line 30: **Silent error swallow produces incorrect zeros.**

The `catch` block resets counts to `0` but returns no `error` state. A caller in `BillingOverviewScreen` showing "0 of 2 collections" is indistinguishable from a failed RPC. Other hooks in this codebase (e.g., `useDemoCollections` on the web side) expose an `error` field.

Suggested fix — add error state:

apps/mobile/src/screens/billing/BillingUsageScreen.tsx
```diff
@@ -0,0 +1,175 @@
...
+  usePlan,
+  useBillingState,
+  useBillingConfig,
+} from '@beakerstack/billing';
+import {
+  booleanFeatureLabel,
+  mergeUsageLimitsCopy,
+  mergeUsageMeterCopy,
+} from '@beakerstack/billing/presentation';
+import {
```
diego-artificerinnovations, line 20: **Hardcoded `feature_a`/`feature_b` keys couple this screen to the demo config.**

Every other feature-list UI (web `PlanFeatureList`, native `PlanFeatureList.native`) derives its rows from `mergePlanFeatureRows(billingConfig)`. Doing that here too would make this screen self-updating when features change in the billing config, and removes the need for the `PlanFeatureRow` local component entirely since `PlanFeatureList.native` already handles the rendering.

Non-blocking for the demo, but worth a TODO comment at minimum so it does not get cargo-culted into a real product deployment.

apps/mobile/src/screens/billing/BillingUsageScreen.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const at = u >= cap;
+  const rightColor = at
+    ? '#b91c1c'
+    : heavy
+      ? '#b45309'
+      : billingColors.textMuted;
+  return (
+    <View style={billingStyles.rowBetween}>
+      <Text style={{ color: billingColors.textPrimary }}>{name}</Text>
+      <Text style={{ color: rightColor }}>
```
diego-artificerinnovations, line 60: **Inline color literals diverge from `billingColors` tokens.**

`#b91c1c` (red-700) and `#b45309` (amber-700) are close but not the same as `billingColors.errorText` (`#991b1b` red-800) and `billingColors.warnText` (`#92400e` amber-800). Using the token constants keeps the palette consistent across both billing screens:

apps/mobile/src/screens/billing/BillingLayout.tsx
```diff
@@ -0,0 +1,35 @@
...
+export function BillingLayout({
+  children,
+}: {
+  children: ReactNode;
+}): React.ReactElement {
+  const navigation = useNavigation();
+  return (
+    <SafeAreaView style={billingStyles.safe} edges={['top', 'bottom']}>
+      <ScrollView contentContainerStyle={billingStyles.scrollContent}>
+        <Pressable
```
diego-artificerinnovations, line 17: **Back `Pressable` is missing accessibility attributes.**

Without `accessibilityRole` and `accessibilityLabel`, screen readers will announce the raw text including the arrow character. Suggest:

packages/billing/src/utils/openExternalUrl.native.ts
```diff
@@ -0,0 +1,13 @@
+import { billingError } from '../errors.js';
+
+/**
+ * Open an https (or custom-scheme) URL in the system browser / handler.
+ */
+export async function openExternalUrl(url: string): Promise<void> {
+  const { Linking } = await import('react-native');
+  const canOpen = await Linking.canOpenURL(url);
+  if (!canOpen) {
+    throw billingError('stripe', `Unable to open URL: ${url}`);
```
diego-artificerinnovations: **Error message leaks the full URL (may include Stripe session IDs).**

Stripe checkout URLs contain session identifiers (e.g. `/v/pay/cs_live_...`). Surfacing the raw URL in a thrown `BillingError` means it can end up in crash reporters or logs. Consider redacting to just the scheme+host:

diego-artificerinnovations (changes requested):
Good scope and clear separation between read-only mobile billing UI and the shared presentation layer. Two change requests below (one rendering bug, one missing error state). Four non-blocking notes as well.

apps/mobile/src/screens/billing/BillingUsageScreen.tsx
```diff
@@ -0,0 +1,175 @@
...
+    useDemoCollectionCount();
+
+  if (!plan) {
+    return (
+      <BillingLayout>
+        <Text style={billingStyles.body}>Loading plan…</Text>
+      </BillingLayout>
+    );
+  }
+  const containers = plan.features.containers_per_account_max as number;
```
mei-artificerinnovations: **Bug:** `as number` cast without a null check. `plan` being truthy only means the billing hook resolved — it doesn't guarantee `containers_per_account_max` is present on this specific plan record. If it's missing, `containers` is `undefined`, `cap > 0` in `FeatureLimitRow` evaluates to `false`, and the row renders `"N of "` (undefined coerces to empty string in JSX). Same issue on line 100 for `itemsCap`.

Suggested fix:
```typescript
const containers = (plan.features.containers_per_account_max as number | undefined) ?? -1;
const itemsCap = (plan.features.items_per_container_max as number | undefined) ?? -1;
```
Using -1 as the fallback matches your existing unlimited-sentinel convention and keeps the display correct.

apps/mobile/src/screens/billing/BillingOverviewScreen.tsx
```diff
@@ -0,0 +1,200 @@
...
+        }
+      />
+      <StatTile
+        label='Collections'
+        value={
+          colLoad
+            ? '—'
+            : (() => {
+                const cap = currentPlan?.features
+                  .containers_per_account_max as number;
```
mei-artificerinnovations: Same unsafe cast as `BillingUsageScreen.tsx` line 99. `currentPlan?.features.containers_per_account_max` could be `undefined` if the plan record omits this field. When `cap` is `undefined`, `cap === -1` is `false`, so the fallback renders `"N of undefined"`.

```typescript
const cap = (currentPlan?.features.containers_per_account_max as number | undefined) ?? -1;
```

supabase/functions/_shared/billing-origins.ts
```diff
@@ -73,12 +76,27 @@ function mergeEnvOrigins(set: Set<string>): void {
...
+ * have `origin === "null"` in the URL spec, so we use `protocol//host` instead.
+ */
+export function billingRedirectAllowlistKey(url: URL): string {
+  if (url.protocol === 'http:' || url.protocol === 'https:') {
+    return url.origin;
+  }
+  if (url.host) {
+    return `${url.protocol}//${url.host}`;
+  }
+  return url.href;
```
mei-artificerinnovations: **Nit:** The `url.href` fallback (for URLs with no host and non-http scheme) produces a key that includes path + query, so `beaker-stack:///some-path` would use its full href as the key — which wouldn't match any entry in `MOBILE_APP_REDIRECT_KEYS`. This is unreachable with the current `beaker-stack://billing` scheme (which parses `host = "billing"` correctly), but the silent fallback is easy to misread as safe.

Consider making the unrecognized-scheme case explicit:
```typescript
// or throw a RedirectValidationError() to fail fast on unrecognized schemes
return `${url.protocol}//${url.pathname}`;
```
or add a comment noting this branch shouldn't be reached with known schemes.

apps/mobile/src/billing/useDemoCollectionCount.ts
```diff
@@ -0,0 +1,50 @@
...
+          ? Math.max(...rows.map(r => Number(r.item_count ?? 0)))
+          : 0
+      );
```
mei-artificerinnovations, line 33: **Nit:** The `catch` block silently resets counts to 0 without exposing an `error` state. If `billing_demo_get_collections` fails (auth, network, RLS), the UI shows `0 of N` which looks like the user has no data rather than a fetch error. The web version of this hook — is it the same? Might be worth an `error` field to let callers show a retry or error banner, consistent with `useDemoCollectionCount` on web.

mei-artificerinnovations (changes requested):
Overall the architecture is solid — the presentation-layer extraction to `@beakerstack/billing/presentation` is clean, the App Store read-only billing UI is well-scoped, and the `billingRedirectAllowlistKey` logic for custom schemes is correct.

Two issues worth fixing before merge (both short); the rest are suggestions/nits.

**Missing coverage:** `BillingScreen.test.tsx` (251 lines) was deleted with no replacement tests for `BillingOverviewScreen` or `BillingUsageScreen`. The new screens have meaningful conditional rendering (`OverviewBanners`, payment-failed states, usage limit display) that isn't covered anywhere. Happy to merge after the two inline fixes if new screen tests land in a follow-up, but worth calling out.

apps/mobile/src/billing/useDemoCollectionCount.ts
```diff
@@ -0,0 +1,50 @@
...
+      const { data, error } = await supabase.rpc(
+        'billing_demo_get_collections',
+        {
+          p_product_id: PRODUCT_ID,
+        }
+      );
+      if (error) throw error;
+      const rows = (data as { id: string; item_count: number }[] | null) ?? [];
+      setCount(rows.length);
+      setMaxItemsInAnyCollection(
```
sarah-artificerinnovations, line 29: The catch block silently resets both counts to 0 with no error state exposed to callers. On `BillingOverviewScreen`, a Supabase failure looks identical to an empty account ("0 of 2 collections"). Suggest adding `error: string | null` to the returned value, set in the catch branch, so callers can surface a load failure rather than silently showing zeros.

apps/mobile/src/screens/billing/BillingTabBar.tsx
```diff
@@ -0,0 +1,81 @@
...
+        showsHorizontalScrollIndicator={false}
+        contentContainerStyle={[styles.scrollInner, { minWidth: width - 32 }]}
+      >
+        {TABS.map(t => {
+          const isActive = active === t.name;
+          return (
+            <Pressable
+              key={t.name}
+              onPress={() => navigation.navigate(t.name)}
+              style={[styles.tab, isActive && styles.tabActive]}
```
sarah-artificerinnovations, line 43: These `Pressable` elements act as tabs but are missing accessibility attributes. Add `accessibilityRole="tab"`, `accessibilityState={{ selected: isActive }}`, and `accessibilityLabel={t.label}` to each tab. Without them VoiceOver (iOS) and TalkBack (Android) announce unlabelled buttons with no selected state — screen reader users cannot tell which tab is active or that the controls form a tab group.

apps/mobile/src/screens/billing/BillingOverviewScreen.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+  const pendingTargetName =
+    subscription?.pending_target_plan_id != null
+      ? catalogPlans.find(p => p.id === subscription.pending_target_plan_id)
+          ?.display_name
+      : undefined;
+
+  return (
+    <BillingLayout>
+      <OverviewBanners
```
sarah-artificerinnovations, line 169: `plan.features.containers_per_account_max as number` is a type assertion without a runtime check. Billing config feature values can be `number | boolean | null`, so if this feature is configured as `true` or `null` on any plan, the assertion silently passes and `cap === -1` / `colCount of cap` produces `NaN` or incorrect output. Prefer `Number(plan.features.containers_per_account_max ?? 0)` or a `typeof` guard before use.

apps/mobile/src/screens/billing/BillingUsageScreen.tsx
```diff
@@ -0,0 +1,175 @@
...
+  return (
+    <BillingLayout>
+      {kind === 'payment_failed' ? (
+        <View
+          style={[
+            billingStyles.card,
+            {
+              backgroundColor: billingColors.errorBg,
+              borderColor: billingColors.errorBorder,
+            },
```
sarah-artificerinnovations: Same type-assertion issue as in `BillingOverviewScreen`. `plan.features.containers_per_account_max as number` is a type assertion without a runtime check. Feature values in the billing config can be `number | boolean | null`, so if either feature is boolean or null, `containers === -1` is `false`, `cap > 0` in `FeatureLimitRow` behaves unexpectedly, and `u / cap` produces `Infinity` (feeding the `ratio >= 0.8` and `u >= cap` color guards incorrectly). Coerce with `Number()` or narrow with a type guard before use.

sarah-artificerinnovations (changes requested):
## Overview

The architecture is solid: splitting the monolithic `BillingScreen` into a proper navigator with Overview/Usage tabs is the right call for both UX and App Store compliance. The `@beakerstack/billing/presentation` consolidation is clean — package exports are properly configured and TypeScript resolves them correctly via `moduleResolution: "bundler"`. CI is green.

Four issues worth addressing before merge, ordered by severity:

### 1. Type assertions on plan feature values may panic at runtime (blocking)

See inline comments on `BillingOverviewScreen` and `BillingUsageScreen`. Feature values are typed as `number | boolean | null` in the billing config, but the code asserts `as number` and uses the result arithmetically. If a plan has `containers_per_account_max: true` (boolean) or `null`, `cap > 0` evaluates unexpectedly and `u / cap` is `Infinity`. Use a runtime coercion (`Number(...)`) or narrow with a type guard.

### 2. `BillingTabBar` is inaccessible to screen readers (blocking)

See inline comment. `Pressable` elements acting as tabs need `accessibilityRole="tab"` and `accessibilityState={{ selected: isActive }}`. Without them VoiceOver/TalkBack users get unlabelled buttons with no selected state.

### 3. `useDemoCollectionCount` swallows errors silently (change request)

See inline comment. The catch block resets to 0 with no exposed error state. Callers cannot distinguish "0 collections" from "fetch failed" — a broken DB connection looks identical to an empty account on the Overview screen.

### 4. Missing test coverage for new screens (follow-up)

The PR removes 251 lines of `BillingScreen.test.tsx` and replaces them with nothing — `BillingOverviewScreen`, `BillingUsageScreen`, `BillingTabBar`, `BillingLayout`, and `useDemoCollectionCount` have zero coverage. If tests are coming in a follow-up PR, please note it in the description so reviewers know it is tracked.

packages/billing/src/components/PlanFeatureList.native.tsx
```diff
@@ -0,0 +1,68 @@
...
+export function PlanFeatureList<P extends ProductBillingConfig>({
+  plan,
+  style,
+}: PlanFeatureListProps<P>): ReactElement {
+  const billingConfig = useBillingConfig<P>();
+  const rows = mergePlanFeatureRows(billingConfig);
+
+  return (
+    <View style={style}>
+      <Text style={styles.title}>What&apos;s included</Text>
```
rahul-artificerinnovations: **Bug:** React Native doesn't process HTML entities. `&apos;` will render as a literal 6-character string. Use a real apostrophe:
```tsx
<Text style={styles.title}>What's included</Text>
```

apps/mobile/src/screens/billing/BillingUsageScreen.tsx
```diff
@@ -0,0 +1,175 @@
...
+  name,
+  available,
+}: {
+  name: string;
+  available: boolean;
+}) {
+  return (
+    <View style={billingStyles.rowBetween}>
+      <Text style={{ color: billingColors.textPrimary }}>
+        {available ? '✓ ' : '✗ '}
```
rahul-artificerinnovations: **Inconsistency:** `FeatureAvailabilityIcon.native.tsx` (introduced in this PR) exports `FeatureCheckIcon` and `FeatureXIcon` for exactly this purpose. `PlanFeatureList.native.tsx` uses them correctly. This component should too:
```tsx
import { FeatureCheckIcon, FeatureXIcon } from '@beakerstack/billing/native';
// ...
{available ? <FeatureCheckIcon /> : <FeatureXIcon />}
```

apps/mobile/src/screens/billing/BillingUsageScreen.tsx
rahul-artificerinnovations: **Unsafe cast:** `as number` will silently produce `undefined as number` if `containers_per_account_max` is ever absent. The `-1` sentinel check below will then fail. Use a guard:
```ts
const containers = typeof plan.features.containers_per_account_max === 'number'
  ? plan.features.containers_per_account_max
  : -1;
```
Same issue on the next line for `items_per_container_max`.

packages/billing/src/components/UsageIndicator.native.tsx
```diff
@@ -70,14 +70,17 @@ export function UsageIndicator<P extends ProductBillingConfig>(
 
 const styles = StyleSheet.create({
   text: { fontSize: 14 },
```
rahul-artificerinnovations, line 73: **Token inconsistency:** `billingUiTokens` was introduced in this PR to centralise these values, but this file still hardcodes `'#111827'` here, `'#6b7280'` (line 74), `'#4b5563'` (line 76), `'#e5e7eb'` (line 81), `'#4f46e5'` (line 84). Use `billingUi.*` throughout:
```ts
import { billingUi } from './billingUiTokens.js';
// ...
color: billingUi.gray900,  // #111827
color: billingUi.gray500,  // #6b7280
```

apps/mobile/src/screens/billing/BillingOverviewScreen.tsx
```diff
@@ -0,0 +1,200 @@
...
+            text: billingColors.warnText,
+          }
+        : {
+            bg: billingColors.infoBg,
+            border: billingColors.infoBorder,
+            text: billingColors.infoText,
+          };
+  return (
+    <View
+      style={{
```
rahul-artificerinnovations: **Style allocation on every render:** structural styles (border radius, padding, margin) should go in `StyleSheet.create` or the shared `styles.ts`, not inline. Only dynamic/computed values (e.g. `borderColor: colors.border`) need to stay inline. The `Banner` component renders on every state change and this creates a new object each time.

apps/mobile/src/screens/billing/BillingOverviewScreen.tsx
rahul-artificerinnovations: **Unsafe cast:** same pattern as `BillingUsageScreen` — `as number` silently turns `undefined` into a number and the `-1` sentinel check breaks. Use a typeof guard.

apps/mobile/src/billing/useDemoCollectionCount.ts
```diff
@@ -0,0 +1,50 @@
...
+        rows.length > 0
+          ? Math.max(...rows.map(r => Number(r.item_count ?? 0)))
+          : 0
+      );
+    } catch {
```
rahul-artificerinnovations: **Silent failure:** the catch block resets to 0 without exposing an error state. Callers can't distinguish between 'no collections' and 'request failed', which means failed fetches look identical to an empty account. Add an `error` state to match the web hook's shape:
```ts
const [error, setError] = useState<Error | null>(null);
// ...
} catch (err) {
  setError(err instanceof Error ? err : new Error('Unknown error'));
  setCount(0);
  setMaxItemsInAnyCollection(0);
}
```

packages/billing/src/utils/openExternalUrl.native.ts
```diff
@@ -0,0 +1,13 @@
+import { billingError } from '../errors.js';
+
+/**
+ * Open an https (or custom-scheme) URL in the system browser / handler.
+ */
+export async function openExternalUrl(url: string): Promise<void> {
+  const { Linking } = await import('react-native');
```
rahul-artificerinnovations: **Unnecessary dynamic import:** `react-native` is always available in the `.native` bundle — there's no lazy-loading benefit here and it adds async overhead before `canOpenURL` runs. Use a static import at the top of the file:
```ts
import { Linking } from 'react-native';
```

supabase/functions/_shared/billing-origins.ts
```diff
@@ -1,3 +1,6 @@
+/** Expo app scheme (`app.config.js` `scheme`) — used for Stripe return URLs on native. */
+const MOBILE_APP_REDIRECT_KEYS = ['beaker-stack://billing'] as const;
```
rahul-artificerinnovations, line 6: **Env var coupling:** `mobileBillingUrls.ts` already reads `EXPO_PUBLIC_BILLING_DEMO_BASE_URL` to allow scheme overrides (staging builds use a different deep-link scheme). This constant hardcodes `beaker-stack://billing` without checking that env var, so staging/CI redirects will be rejected by `assertRedirectUrlAllowed`. Either derive this from the same env var, or at minimum add a comment documenting the required coupling.

rahul-artificerinnovations (changes requested):
## Overview

This PR makes good progress on the read-only billing UI for native and the new `@beakerstack/billing/presentation` shared entrypoint. The architecture is right — keeping upgrade CTAs off native for App Store compliance, sharing formatters across platforms, and the tab-based billing navigator is clean. Several issues need to be addressed before this merges.

## Required changes

### 1. HTML entity in React Native JSX (`PlanFeatureList.native.tsx`)

`What&apos;s included` renders as a literal string in React Native — RN doesn't process HTML entities. Use the literal apostrophe: `What's included`.

### 2. `PlanFeatureRow` bypasses `FeatureAvailabilityIcon` (`BillingUsageScreen.tsx`)

`FeatureAvailabilityIcon.native.tsx` was introduced in this very PR with `FeatureCheckIcon` / `FeatureXIcon`. `PlanFeatureRow` renders raw emoji `'✓ '` / `'✗ '` inline instead of using those components. `PlanFeatureList.native.tsx` (same PR) gets this right. `PlanFeatureRow` should match.

### 3. Hardcoded hex colors in `UsageIndicator.native.tsx`

`billingUiTokens` is introduced in this PR specifically to centralise these values, but `UsageIndicator.native.tsx` still hardcodes `#111827`, `#6b7280`, `#4b5563`, `#e5e7eb`, `#4f46e5` in its updated `StyleSheet.create`. These should use `billingUi.*` tokens.

### 4. Inline style objects in `BillingOverviewScreen.tsx` and `BillingUsageScreen.tsx`

Both files create new style objects on every render for structural layout (border radius, padding, margin, etc.). The project uses `StyleSheet.create` for this. Move the structural styles into `StyleSheet.create` (or the shared `styles.ts`). Computed/dynamic values (`borderColor: colors.border`) are fine inline.

### 5. Unsafe `as number` casts on plan features

`containers_per_account_max as number` and `items_per_container_max as number` will silently produce `undefined as number` if the plan schema ever changes or a feature key is missing. Use a numeric guard instead:

```ts
const containers = typeof plan.features.containers_per_account_max === 'number'
  ? plan.features.containers_per_account_max
  : -1;
```

This pattern appears in both `BillingOverviewScreen.tsx` and `BillingUsageScreen.tsx`.

### 6. Silent error handling in `useDemoCollectionCount`

The catch block resets count to 0 with no error surface. Callers get stale-looking zeros with no way to distinguish "no collections" from "request failed". Add an `error` state (mirrors the web hook's shape) so callers can show appropriate feedback.

### 7. Dynamic `import('react-native')` in `openExternalUrl.native.ts`

React Native should be a static top-level import. The dynamic import is unusual and adds unnecessary async overhead before `canOpenURL` runs. `react-native` is always available in the .native bundle — no tree-shaking benefit either.

### 8. Hardcoded scheme in `billing-origins.ts`

`MOBILE_APP_REDIRECT_KEYS` hardcodes `beaker-stack://billing`. `mobileBillingUrls.ts` already reads `EXPO_PUBLIC_BILLING_DEMO_BASE_URL` for overrides. If that env var points at a different scheme (staging, e.g. `beaker-stack-staging://billing`), the redirect will be rejected. Either read the same env var here or document this coupling explicitly.

### 9. Missing tests for new screens

`BillingScreen.test.tsx` (251 lines) was deleted with no replacement. `BillingOverviewScreen` and `BillingUsageScreen` are new screens with meaningful branching logic (banner variants, feature flags, loading states, the `as number` cap logic). At minimum the happy-path render and the loading/error states should have coverage.

supabase/functions/_shared/billing-origins.ts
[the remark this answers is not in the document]
ZappoMan: Good catch — mergeEnvOrigins now normalizes parseable env entries with billingRedirectAllowlistKey(new URL(t)) (with a try/catch so malformed custom-scheme strings are skipped rather than added raw). That keeps deep-link success/cancel URLs aligned with assertRedirectUrlAllowed. (49861e1)

packages/billing/src/utils/openExternalUrl.native.ts
```diff
@@ -0,0 +1,13 @@
...
+  const canOpen = await Linking.canOpenURL(url);
+  if (!canOpen) {
+    throw billingError('stripe', `Unable to open URL: ${url}`);
+  }
+  await Linking.openURL(url);
```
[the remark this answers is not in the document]
ZappoMan: Agreed. Errors now use redactedUrlForMessage() (scheme + host only); the full URL is passed as a non-display cause field on billingError. Test updated in openExternalUrl.native.test.ts. (49861e1)

packages/billing/src/components/CustomerPortalLink.native.tsx
```diff
@@ -13,7 +14,14 @@ export function CustomerPortalLink<P extends ProductBillingConfig>({
...
       style={style}
       disabled={pending}
-      onPress={() => void openPortal()}
+      onPress={() => {
+        void (async () => {
+          const url = await openPortal();
+          if (url) {
+            await openExternalUrl(url);
+          }
+        })();
```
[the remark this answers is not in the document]
ZappoMan: Wrapped the portal open flow in try/catch and surface failures via local openError state under the button, so we avoid unhandled rejections. (49861e1)

packages/billing/src/components/UpgradePrompt.native.tsx
```diff
@@ -16,11 +17,7 @@ export function UpgradePrompt<P extends ProductBillingConfig>({
 
   const onUpgrade = async () => {
     if (!planId) return;
-    const r = await startCheckout(planId);
-    if (r?.checkoutUrl) {
-      const { Linking } = await import('react-native');
-      await Linking.openURL(r.checkoutUrl);
-    }
+    await launchStripeCheckout(startCheckout, planId);
   };
```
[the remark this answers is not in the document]
ZappoMan: Same pattern: onUpgrade is now async with try/catch and a local checkoutOpenError message when launchStripeCheckout / openExternalUrl fails. (49861e1)

apps/mobile/src/billing/mobileBillingUrls.ts
```diff
@@ -10,9 +12,17 @@ export function getMobileBillingProviderUrls(): {
   const base =
     (typeof process !== 'undefined' &&
       process.env?.['EXPO_PUBLIC_BILLING_DEMO_BASE_URL']) ||
-    'http://127.0.0.1:8081';
+    'beaker-stack://billing';
+  const isDeepLink = base.includes('://') && !base.startsWith('http');
+  if (isDeepLink) {
+    return {
+      checkoutSuccessUrl: `${base}?checkout=success`,
+      checkoutCancelUrl: `${base}/plans?checkout=cancel`,
```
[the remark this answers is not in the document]
ZappoMan: Updated deep-link cancel to ${base}?checkout=cancel (no /plans segment). Mobile billing only has Overview/Usage routes. (49861e1)

apps/mobile/src/screens/billing/BillingOverviewScreen.tsx
ZappoMan: Fixed with numericPlanFeature() and guarding the collections StatTile when currentPlan is missing — shows em dash while loading and avoids N of undefined. (49861e1)

apps/mobile/src/billing/useDemoCollectionCount.ts
ZappoMan: Added error to the hook return value; Overview shows an "Unable to load collection counts" tile when the RPC fails instead of silent zeros. (49861e1)

apps/mobile/src/screens/billing/BillingUsageScreen.tsx
ZappoMan: Replaced hardcoded feature_a/feature_b rows with mergePlanFeatureRows + planFeatureLine and the shared check/x icons (same pattern as PlanFeatureList.native). (49861e1)

apps/mobile/src/screens/billing/BillingUsageScreen.tsx
ZappoMan: Swapped inline hex literals for billingColors.errorText / billingColors.warnText. (49861e1)

apps/mobile/src/screens/billing/BillingLayout.tsx
ZappoMan: Added accessibilityRole="button" and accessibilityLabel="Go back" on the back Pressable. (49861e1)

packages/billing/src/utils/openExternalUrl.native.ts
ZappoMan: Same fix as above — user-facing message is redacted to scheme+host. (49861e1)

apps/mobile/src/screens/billing/BillingUsageScreen.tsx
ZappoMan: Using numericPlanFeature(features, key) which only accepts finite numbers (otherwise falls back to -1 unlimited sentinel). (49861e1)

apps/mobile/src/screens/billing/BillingOverviewScreen.tsx
ZappoMan: Same numericPlanFeature helper on Overview collections StatTile. (49861e1)

supabase/functions/_shared/billing-origins.ts
ZappoMan: Replaced the url.href fallback with throw new RedirectValidationError() so unrecognized custom schemes fail fast. Documented the default beaker-stack://billing coupling in a comment. (49861e1)

apps/mobile/src/billing/useDemoCollectionCount.ts
ZappoMan: Hook now exposes error; Usage screen shows a banner when collection counts fail to load. (49861e1)

apps/mobile/src/billing/useDemoCollectionCount.ts
ZappoMan: Done — error is returned from useDemoCollectionCount and surfaced on Overview/Usage. (49861e1)

apps/mobile/src/screens/billing/BillingTabBar.tsx
ZappoMan: Added accessibilityRole="tab", accessibilityState={{ selected }}, and accessibilityLabel on each tab Pressable. (49861e1)

apps/mobile/src/screens/billing/BillingOverviewScreen.tsx
ZappoMan: Addressed with numericPlanFeature() — only finite numbers are used; non-numeric/missing values fall back to the unlimited sentinel. (49861e1)

apps/mobile/src/screens/billing/BillingUsageScreen.tsx
ZappoMan: Same numericPlanFeature helper on Usage limits rows. (49861e1)

packages/billing/src/components/PlanFeatureList.native.tsx
ZappoMan: Fixed — title now uses a literal apostrophe (What's included). (49861e1)

apps/mobile/src/screens/billing/BillingUsageScreen.tsx
ZappoMan: Usage screen now uses FeatureCheckIcon / FeatureXIcon (exported from @beakerstack/billing/native) instead of unicode marks. (49861e1)

apps/mobile/src/screens/billing/BillingUsageScreen.tsx
ZappoMan: See numericPlanFeature in apps/mobile/src/billing/planFeatureValue.ts. (49861e1)

packages/billing/src/components/UsageIndicator.native.tsx
ZappoMan: UsageIndicator.native now reads bar/label colors from billingUi tokens. (49861e1)

apps/mobile/src/screens/billing/BillingOverviewScreen.tsx
ZappoMan: Moved Banner/StatTile structural styles into styles.ts (banner*, statTile*). (49861e1)

apps/mobile/src/screens/billing/BillingOverviewScreen.tsx
ZappoMan: Overview collections cap uses numericPlanFeature as well. (49861e1)

apps/mobile/src/billing/useDemoCollectionCount.ts
ZappoMan: Hook exposes error; callers show failure UI instead of misleading zeros. (49861e1)

packages/billing/src/utils/openExternalUrl.native.ts
ZappoMan: Switched to a static import { Linking } from 'react-native'. (49861e1)

supabase/functions/_shared/billing-origins.ts
ZappoMan: Added a comment that MOBILE_APP_REDIRECT_KEYS must match EXPO_PUBLIC_BILLING_DEMO_BASE_URL host, with BILLING_ALLOWED_ORIGINS for other schemes. Edge functions cannot read Expo env at runtime, so staging overrides belong in Supabase env. (49861e1)

diego-artificerinnovations (approved):
All six points addressed — happy to approve.

- **Collections StatTile null-safety**: `numericPlanFeature()` helper + `formatCollectionsStat()` correctly handles null plan, loading, and error cases. Clean.
- **Error state in `useDemoCollectionCount`**: `error` is now exposed, cleared on each refresh, and consumed by both screens (Usage shows an error card, Overview shows "Unable to load").
- **Hardcoded feature keys**: `BooleanPlanFeatures` now derives rows from `mergePlanFeatureRows(...).filter(row => row.kind === 'boolean')` — fully config-driven and consistent with `PlanFeatureList.native`.
- **Color tokens**: `billingColors.errorText` / `billingColors.warnText` used throughout, inline literals gone.
- **Back button a11y**: `accessibilityRole='button'` + `accessibilityLabel='Go back'` added.
- **URL redaction**: `redactedUrlForMessage()` strips path and query, keeping scheme+host only in the error message. The static `Linking` import is also a nice cleanup.

sarah-artificerinnovations (approved):
All four items addressed — approving.

- **Type assertions** → `numericPlanFeature()` helper in `planFeatureValue.ts` with a real `typeof v === 'number' && Number.isFinite(v)` guard. Both screens now use it. Clean.
- **Tab accessibility** → `accessibilityRole='tab'`, `accessibilityState={{ selected: isActive }}`, `accessibilityLabel` all added to `BillingTabBar`.
- **Error state** → `useDemoCollectionCount` now returns `error: Error | null`; `BillingUsageScreen` surfaces it as a visible error card; `BillingOverviewScreen` shows "Unable to load" via `formatCollectionsStat`.
- **Tests** → `BillingScreens.test.tsx` covers both Overview and Usage screens with mocked hooks.

Also appreciated the bonus fixes: inline styles lifted into `StyleSheet.create`, URL redaction in `openExternalUrl` error messages, and the `billing-origins.ts` edge-case tightening. CI green.

apps/mobile/src/screens/billing/BillingOverviewScreen.tsx
```diff
@@ -0,0 +1,209 @@
...
+        kind={kind}
+        subscription={subscription}
+        pendingTargetName={pendingTargetName}
+      />
+      {planLoad && kind === 'loading' ? (
+        <ActivityIndicator style={{ marginVertical: 24 }} />
+      ) : currentPlan ? (
+        <View style={billingStyles.card}>
+          <Text style={billingStyles.cardTitle}>
+            You&apos;re on the {currentPlan.display_name} plan
```
rahul-artificerinnovations: **HTML entity bug (same fix applied in `PlanFeatureList.native.tsx` was missed here):** `You&apos;re` will render as a literal 7-character string in React Native. The `BillingOverviewScreen.test.tsx` test even checks for `/You're on the Free plan/` (literal apostrophe), so this will cause a test failure too.

Use a real apostrophe:
```tsx
You're on the {currentPlan.display_name} plan
```

rahul-artificerinnovations (changes requested):
All 8 of my original comments are addressed — great work on the `numericPlanFeature` helper, the `error` state in `useDemoCollectionCount`, swapping to `billingUi.*` tokens, and the new test coverage. One issue was missed and needs fixing before this can be approved.

mei-artificerinnovations (approved):
All comments addressed — approving.

Quick verification of each fix:

- **`as number` casts** → `numericPlanFeature()` helper is correct: `typeof v === 'number' && Number.isFinite(v)`, fallback to -1 sentinel. Used consistently in both screens. ✓
- **`useDemoCollectionCount` silent errors** → `error: Error | null` exposed; both Overview and Usage show an error banner when the RPC fails. ✓
- **`billing-origins.ts` fallback** → `url.href` replaced with `throw new RedirectValidationError()`. `mergeEnvOrigins` now routes all env entries through `billingRedirectAllowlistKey` with an inner try/catch. ✓
- **`openExternalUrl` URL leakage** → `redactedUrlForMessage()` strips path+query; full URL passed as non-display cause field. Test updated to assert `cs_` token doesn't appear in message. ✓
- **`CustomerPortalLink` / `UpgradePrompt` unhandled rejections** → both now have try/catch with local error state shown under the button. ✓
- **`BillingUsageScreen` hardcoded feature keys** → `BooleanPlanFeatures` component drives rows from `mergePlanFeatureRows` + `planFeatureLine`, same pattern as web. ✓
- **Inline color literals** → `billingColors` tokens used throughout; `billingUiTokens` extended with missing gray500. `UsageIndicator` now fully tokenized. ✓
- **Inline structural styles** → extracted to `billingStyles` (`banner`, `bannerTitle`, `statValue`, `paymentFailedCard`, etc.). ✓
- **Accessibility** → `BillingLayout` back button has `accessibilityRole='button'` + label; `BillingTabBar` tabs have `role='tab'` + `selected` state + label. ✓
- **`mobileBillingUrls` cancel URL** → fixed to `${base}?checkout=cancel` (no stale `/plans` segment). ✓
- **`PlanFeatureList` HTML entity** → literal apostrophe. ✓
- **`openExternalUrl` dynamic import** → converted to static `import { Linking } from 'react-native'`. ✓
- **Test coverage** → `BillingScreens.test.tsx` added for both new screens. ✓

One very minor leftover: `CustomerPortalLink.native.tsx` line 678 still uses a hardcoded `'#b91c1c'` for the `openError` text color. Suggest swapping to `billingColors.errorText` in a follow-up — not worth blocking this.

apps/mobile/src/screens/billing/BillingOverviewScreen.tsx
ZappoMan: Good catch — missed this when fixing PlanFeatureList. Replaced `You&apos;re` with a literal apostrophe so it matches `BillingScreens.test.tsx` and renders correctly in RN. (pushed on this branch just now)

rahul-artificerinnovations (approved):
All 9 issues resolved — the `You&apos;re` entity is fixed, no remaining HTML entities anywhere in the additions. LGTM.

PR #224, opened 2026-05-16
fix(web): landing and app UI polish — images, layout, nav, and branding

ZappoMan:
## Summary

Marketing and app chrome polish across landing, auth, dashboard, and billing:

### Landing images
- Replace `mobile-hero.avif` with a web + mobile composite on a white background for the "Web and mobile from a single codebase" feature row (contain scaling at 560×315 so full device frames are visible).
- Add dedicated `hero.avif` for the headline carousel.
- Replace `seo.avif` with Lighthouse score artwork for the SEO feature row.

### Layout
- Standardize marketing and app chrome on `max-w-content` (1024px) via shared `ContentContainer` and Tailwind tokens, with a prose variant (800px) for policy/FAQ pages.
- Ensure JIT generates width utilities by scanning shared layout config.

### Navigation & scroll
- Fix `ScrollToTop` resetting to top on every pathname change, which broke Features/Pricing/FAQ links from policy pages; honor `location.hash` and retry until lazy landing sections mount.
- Keep app header sticky on dashboard, billing, and auth pages (scroll-gated shadow, matching public Nav).

### Branding
- Wire marketing landing and SEO meta to `BRANDING.displayName` ("Beaker Stack") so app chrome, policies, and landing copy stay consistent; keep `BeakerStack` for technical IDs.
- Update mobile/web billing copy, seeds, legal policies, and docs to match.

## Test plan

- [ ] Landing: verify hero carousel, single-codebase feature row, and SEO feature row images look correct (full device frames, no clipping).
- [ ] Layout: confirm landing sections, auth pages, dashboard, and billing align to 1024px; policy/FAQ pages use narrower prose width.
- [ ] From a policy page, follow Features / Pricing / FAQ nav links and confirm scroll lands on the correct section (not page top).
- [ ] Dashboard, billing, and auth: header stays sticky with shadow on scroll.
- [ ] User-facing copy shows "Beaker Stack" (not "BeakerStack") where appropriate.
- [ ] `npx vitest run` for affected web/shared tests (`FeatureRows`, `ScrollToTop`, `AppHeader`, `landing`, dashboard).
- [ ] E2E: `tests/e2e/web/flows/home.yaml` passes.


apps/web/src/components/ScrollToTop.tsx
```diff
@@ -1,20 +1,48 @@
...
+  const id = decodeURIComponent(hash.replace(/^#/, ''));
+  if (!id) return false;
+  const el = document.getElementById(id);
+  if (!el) return false;
+  el.scrollIntoView();
+  return true;
+}
+
+/** Retry until lazy below-fold sections (pricing, FAQ) have mounted. */
+export function scrollToHash(hash: string, frame = 0): void {
```
diego-artificerinnovations: Silent fallback after 60 frames is acceptable, but a `console.warn` in `__DEV__` mode would help diagnose cases where a hash anchor never resolves (e.g. a typo in the `id`). Not a blocker — just something to consider.

apps/web/src/config/__tests__/landing.test.ts
```diff
@@ -0,0 +1,18 @@
+import { describe, it, expect } from 'vitest';
+import { BRANDING } from '@beakerstack/shared/config/branding';
+import { landingConfig } from '../landing';
+
+const GITHUB_REPO = 'Artificer-Innovations/BeakerStack';
```
diego-artificerinnovations: Nit: `GITHUB_REPO` is a hardcoded string that will silently go stale if the repo is ever renamed or forked. If there is a shared config for the repo URL (e.g. in `BRANDING`), wiring it from there would keep this self-maintaining. Fine as-is for now.

apps/web/src/components/landing/sections/FeatureGrid.tsx
```diff
@@ -10,7 +11,7 @@ export function FeatureGrid({ config }: FeatureGridProps) {
       id='features'
       className='py-20 md:py-24 bg-gray-50 dark:bg-gray-900'
     >
-      <div className='max-w-[1200px] mx-auto px-6'>
+      <ContentContainer>
```
diego-artificerinnovations, line 14: Previous width was `max-w-[1200px]`; `ContentContainer` defaults to `max-w-content` (1024px). Narrowing is intentional per the PR description — just flagging for visual QA since this is the widest grid on the page.

diego-artificerinnovations (approved):
Solid polish PR. Branding centralization, `ContentContainer` abstraction, and the `ScrollToTop` hash-routing fix are all the right moves. A couple of inline observations — nothing blocking.

rahul-artificerinnovations (approved):
## Review

Clean, well-tested polish PR. No blockers.

### ScrollToTop hash fix
Correct approach. The `useLayoutEffect` deps include both `pathname` and `hash`, so every navigation variant (pathname change + hash, hash-only change, hash removed) is handled correctly. The rAF retry loop capping at 60 frames (~1 second at 60fps) is the right bound for lazy-mounted landing sections. Tests cover the important cases: hash suppresses scroll-to-top, pathname change to hash scrolls the target, retry until element exists.

One minor note: `scrollToHash` doesn't cancel in-flight rAF chains if the component unmounts mid-retry. In practice the loop runs for at most ~1 second and `scrollIntoView` on a detached element fails silently, so this is harmless — just worth knowing.

### ContentContainer / layoutWidth
Good abstraction. The `as` prop keeps it flexible (`DashboardPage` uses `as='main'`). The `LAYOUT_WIDTH` values match what was previously hardcoded: 64rem (1024px) for content, 50rem (800px) for prose/policy — confirmed by the `PolicyPage` swap to `variant='prose'`. Adding `../../packages/shared/src/config/**` to Tailwind's content scan + safelisting `max-w-content`/`max-w-prose` covers the JIT concern.

### AppHeader sticky
`window.scrollY > 8` threshold, passive listener, cleanup on unmount — correct. The element swap to `<header>` is a semantic improvement (implicit `banner` role, which is why `getByRole('banner')` works in the tests). Shadow/no-shadow state tests using `Object.defineProperty` on `scrollY` are solid.

### Branding
`BRANDING.displayName` is now the single source of truth across landing copy, meta tags, `LEGAL_CONFIG.brandName`, and `DemoBanner`. The landing test's serialization check (`JSON.stringify(landingConfig)` must not contain `pascalName`, excluding GitHub URLs) is a good regression guard.

### BillingPageShell
`maxWidthClass` prop removed cleanly — both callers (`BillingInvoicesPage`, `BillingPlansPage`) were already passing the default value, so no behavioral change.

### Minor nit (non-blocking)
`joinClasses` in `ContentContainer.web.tsx` is a manual re-implementation of `clsx`. If the project already has `clsx` as a dep, might as well use it; if not, the simple version is fine for two arguments.

sarah-artificerinnovations (approved):
## Review: PR #224 — Landing and app UI polish

**Verdict: APPROVE ✅**

This is a well-executed polish pass. Architecture is clean, no blockers.

### Branding centralization
`landing.ts` now derives all user-facing display strings from `BRANDING.displayName` — no more hardcoded BeakerStack scattered through copy. The test in `landing.test.ts` is a nice safety net: serializes the config, strips GitHub URLs, and asserts `BRANDING.pascalName` doesn't appear in the output. That will catch regressions immediately.

### Layout standardization
Replacing ad-hoc `max-w-[1200px]` with the `max-w-content` token (1024px via `ContentContainer`) is the right call — consistent with the rest of the app and the shared design system. The narrowing is intentional and the Tailwind safelist already covers both tokens.

### Hash navigation (`scrollToHash`)
The 60-frame rAF retry loop is a pragmatic fix for lazy-mounted landing sections. At 60fps that's ~1 second of patience before giving up — reasonable. Exporting the function and testing it in isolation is good practice.

### `BillingPageShell` prop removal
Verified all callers (`BillingInvoicesPage`, `BillingPlansPage`) — neither passes `maxWidthClass`. Safe removal.

### Meta tags
The `og:description` and `twitter:description` in `index.html` are left as generic copy — fine since they don't contain the brand name and the `%OG_TITLE%` placeholder is correctly wired through the Vite plugin.

CI green across all checks. Ship it.

apps/web/src/components/ScrollToTop.tsx
```diff
@@ -1,20 +1,48 @@
...
+  if (scrollToHashElement(hash)) return;
+  if (frame >= HASH_SCROLL_MAX_FRAMES) return;
+  requestAnimationFrame(() => scrollToHash(hash, frame + 1));
```
mei-artificerinnovations: **Blocking: RAF callbacks not cancelled on route change (race condition)**

`scrollToHash` chains `requestAnimationFrame` calls recursively up to 60 frames but never stores or cancels the pending ID. If the user navigates away mid-chain, callbacks queued before the navigation will still fire — calling `scrollIntoView()` in the context of the new page on an element that may not even exist there.

Suggested fix — make `scrollToHash` return the RAF id and cancel it in the effect cleanup:

```tsx
export function scrollToHash(hash: string, frame = 0): number | null {
  if (scrollToHashElement(hash)) return null;
  if (frame >= HASH_SCROLL_MAX_FRAMES) return null;
  return requestAnimationFrame(() => scrollToHash(hash, frame + 1));
}

// Inside ScrollToTop:
useLayoutEffect(() => {
  if (hash) {
    const rafId = scrollToHash(hash);
    return () => { if (rafId !== null) cancelAnimationFrame(rafId); };
  }
  window.scrollTo(0, 0);
  topRef.current?.focus({ preventScroll: true });
}, [pathname, hash]);
```

Note: the recursive calls inside the chain also each return an id — to cancel a full in-flight chain you'd need to thread the cancellation through (e.g. via a `ref`-held flag or a single cancel token). The simplest approach is a `cancelled` flag in the closure:

```tsx
useLayoutEffect(() => {
  if (!hash) {
    window.scrollTo(0, 0);
    topRef.current?.focus({ preventScroll: true });
    return;
  }
  let cancelled = false;
  function attempt(frame: number) {
    if (cancelled) return;
    if (scrollToHashElement(hash)) return;
    if (frame < HASH_SCROLL_MAX_FRAMES) requestAnimationFrame(() => attempt(frame + 1));
  }
  attempt(0);
  return () => { cancelled = true; };
}, [pathname, hash]);
```

packages/shared/src/components/navigation/AppHeader.web.tsx
```diff
@@ -16,6 +18,13 @@ export interface AppHeaderProps {
 export function AppHeader({ supabaseClient: _supabaseClient }: AppHeaderProps) {
   const auth = useAuthContext();
   const profile = useProfileContext();
+  const [scrolled, setScrolled] = useState(false);
+
+  useEffect(() => {
+    const onScroll = () => setScrolled(window.scrollY > 8);
```
mei-artificerinnovations: **Suggestion: skip redundant state updates in scroll listener**

`setScrolled(window.scrollY > 8)` is called on every scroll event — even when the boolean value hasn't changed (e.g. scrolling from Y=100 to Y=200 keeps firing `setScrolled(true)`). React batches and bails out on same-value state, so this is unlikely to cause visible bugs, but it does allocate a new closure and queue a reconciliation check on every frame.

A simple guard avoids the noise:

```ts
const onScroll = () => {
  const next = window.scrollY > 8;
  setScrolled(prev => prev === next ? prev : next);
};
```

Or equivalently with a functional updater: `setScrolled(prev => (window.scrollY > 8) || false)` — but the explicit comparison is clearer.

mei-artificerinnovations (changes requested):
## Review of PR #224 — Landing and App UI Polish

Good overall: sticky header, hash-aware scroll, image assets, and the ContentContainer abstraction are all solid directions. One blocking issue in `scrollToHash` and a few suggestions below.

### Other observations

**`ContentContainer.web.tsx` — polymorphic `as` prop typing**
The `as` prop accepts any `React.ElementType` but the prop spread isn't typed generically against that element's own props, so passing invalid HTML attributes won't be caught at compile time. Not a blocker for current usages (all callers pass no extras), but worth tightening if this is intended as a shared primitive.

**`joinClasses` utility**
If the codebase already has `cn`, `clsx`, or a similar helper, prefer that over a new one-liner to avoid two diverging utilities.

**`packages/shared/src/generated/policies.ts`**
This looks hand-edited. If the file is generated, the generator should produce the update; if it's intentionally hand-maintained, renaming it away from `generated/` would prevent confusion.

**Tests for `scrollToHash` retry logic**
The new test mocks cover the happy path but don't exercise the max-frames guard or verify that stale frames don't fire after cleanup (which is exactly the bug fixed in the blocking item below). Once the cancellation fix lands, a test covering the cleanup case would be valuable.

apps/web/src/config/__tests__/landing.test.ts
ZappoMan: Agreed here, this is actually not a very useful test because as a template, most people will change this landing page branding to not include a github URL, maybe we should just remove this test.

apps/web/src/components/landing/sections/FeatureGrid.tsx
ZappoMan: @diego-artificerinnovations - yes this was intentional.

apps/web/src/components/ScrollToTop.tsx
```diff
@@ -1,20 +1,48 @@
 import { useLayoutEffect, useRef } from 'react';
 import { useLocation } from 'react-router-dom';
 
+const HASH_SCROLL_MAX_FRAMES = 60;
+
+function scrollToHashElement(hash: string): boolean {
+  const id = decodeURIComponent(hash.replace(/^#/, ''));
```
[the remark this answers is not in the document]
ZappoMan: Fixed in ef7210b: `parseHashId` wraps `decodeURIComponent` in try/catch and treats decode failures as a missing target (returns false, no throw).

supabase/seed.sql
```diff
@@ -5,7 +5,7 @@
 INSERT INTO public.billing_products (id, display_name, description)
 VALUES (
     'beakerstack',
-    'BeakerStack',
+    'Beaker Stack',
     'Template billing demo product'
 )
 ON CONFLICT (id) DO NOTHING;
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 31633ab: `ON CONFLICT (id) DO UPDATE SET display_name, description` so re-seeding applies the branding copy.

apps/mobile/supabase/seed.sql
```diff
@@ -3,7 +3,7 @@
 INSERT INTO public.billing_products (id, display_name, description)
 VALUES (
     'beakerstack',
-    'BeakerStack',
+    'Beaker Stack',
     'Template billing demo product'
 )
 ON CONFLICT (id) DO NOTHING;
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 31633ab: same `DO UPDATE` upsert in the mobile mirror seed.

apps/web/src/pages/__tests__/HomePage.test.tsx
```diff
@@ -44,7 +45,7 @@ vi.mock('@beakerstack/billing', async importOriginal => {
 
 vi.mock('../../config/landing', () => ({
   landingConfig: {
-    brand: { name: 'BeakerStack', tagline: 'Test tagline' },
+    brand: { name: BRANDING.displayName, tagline: 'Test tagline' },
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 31633ab: removed the top-level `BRANDING` import from the mock factory; the mock now uses a local `'Beaker Stack'` string.

docs/BRANDING.md
```diff
@@ -120,18 +120,25 @@ The favicon files are referenced in `apps/web/index.html` with appropriate link
...
+- **Welcome strings**: `packages/shared/src/utils/strings.ts` (`HOME_TITLE`, etc.)
+- **Legal policies**: `LEGAL_CONFIG.brandName` in `packages/shared/src/config/legal.ts`
+- **Browser tab and SEO meta**: `apps/web/vite.config.ts` injects `displayName` into `index.html`
 
-To change the app name, update the text in both header components.
+To rebrand, update `branding.ts` (or run `npm run rename`) — do not hardcode the display name in landing config or headers.
+
+### PascalCase / technical name
+
+**`BRANDING.pascalName`** (`BeakerStack`) is for identifiers only: GitHub repo URLs, npm scope (`@beakerstack/*`), domain (`beakerstack.com`), database product ids (`beakerstack`), build artifacts, and CI labels. Do not use it in sentences shown to end users.
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 31633ab: clarified that `BRANDING.flatName` is for lowercase DB/billing ids and `pascalName` is for code-style identifiers; GitHub URLs called out separately.

apps/web/src/components/__tests__/ScrollToTop.test.tsx
```diff
@@ -53,4 +53,71 @@ describe('ScrollToTop', () => {
...
+    const target = document.createElement('section');
+    target.id = 'faq';
+    target.scrollIntoView = vi.fn();
+
+    scrollToHash('#faq');
+    expect(target.scrollIntoView).not.toHaveBeenCalled();
+
+    document.body.append(target);
+    scrollToHash('#faq');
+    expect(target.scrollIntoView).toHaveBeenCalled();
```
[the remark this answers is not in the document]
ZappoMan: Fixed in ef7210b: the retry test now appends the target between queued rAF frames (no second direct `startHashScroll` call). Also added cleanup and malformed-hash coverage.

apps/web/src/components/ScrollToTop.tsx
```diff
@@ -1,20 +1,48 @@
...
+
 export function ScrollToTop() {
-  const { pathname } = useLocation();
+  const { pathname, hash } = useLocation();
   const topRef = useRef<HTMLSpanElement>(null);
 
   useLayoutEffect(() => {
+    if (hash) {
+      scrollToHash(hash);
+      return;
```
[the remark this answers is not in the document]
ZappoMan: Fixed in ef7210b: hash scroll uses a `cancelled` flag in `startHashScroll`, returned as effect cleanup so stale rAF chains stop on route change/unmount.

apps/web/src/pages/ProfilePage.tsx
```diff
@@ -20,7 +21,7 @@ export default function ProfilePage() {
       <AppHeader supabaseClient={supabase} />
 
       {/* Main Content */}
-      <div className='max-w-[1024px] mx-auto py-6 sm:px-6 lg:px-8'>
+      <ContentContainer className='py-6'>
         <div className='px-4 py-6 sm:px-0'>
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 31633ab: removed the inner `px-4` wrapper on ProfilePage so horizontal padding comes only from `ContentContainer` (16px on mobile, aligned with other app shells).

apps/web/src/components/ScrollToTop.tsx
ZappoMan: Fixed in ef7210b: logs `console.warn` when the hash anchor is not found after max frames, gated on `import.meta.env.DEV`.

apps/web/src/config/__tests__/landing.test.ts
ZappoMan: Done in 31633ab: dropped the GitHub-URL stripping / pascalName serialization check; kept a single assertion that `landingConfig.brand.name` uses `BRANDING.displayName`.

apps/web/src/components/landing/sections/FeatureGrid.tsx
ZappoMan: Acknowledged — the 1024px `max-w-content` narrowing is intentional (per PR description). No code change.

apps/web/src/components/ScrollToTop.tsx
ZappoMan: Fixed in ef7210b: implemented the `cancelled` closure pattern in `startHashScroll` with effect cleanup, plus tests for cleanup and hash-change cancellation.

packages/shared/src/components/navigation/AppHeader.web.tsx
ZappoMan: Fixed in 31633ab: scroll listener now uses `setScrolled(prev => prev === next ? prev : next)` to skip redundant updates.

mei-artificerinnovations (approved):
Both blocking and suggestion items are resolved. The `cancelled`-flag closure in `startHashScroll` is exactly the right fix — it cleanly propagates cancellation through the full recursive RAF chain, and returning the cleanup function directly from `useLayoutEffect` is the idiomatic hook pattern. The `import.meta.env.DEV` warning for unfound anchors is a nice touch. Scroll listener guard is in too. Approving.

diego-artificerinnovations (approved):
Re-reviewed after the follow-up commits. Mei's blocking RAF cancellation issue is properly addressed — `startHashScroll` uses a `cancelled` flag closure and returns a cleanup function that `useLayoutEffect` propagates directly. The `console.warn` on timeout is now gated on `import.meta.env.DEV`. The seed upsert, scroll listener guard, and test cleanups all look good. CI is fully green. Approving.

PR #249, opened 2026-05-17
feat(shared): centralize theme colors in @beakerstack/shared

### LINKED ISSUE #243: Centralize web and mobile theme colors in @beakerstack/shared
## Summary

Introduce a single source of named color tokens in `@beakerstack/shared` and wire both the web (Tailwind) and mobile (React Native `StyleSheet`) apps to consume it.

## Business reasons

- **Brand consistency:** Web auth/landing uses Tailwind `primary-*` (blue `#3b82f6`) while billing, dashboard, and most mobile surfaces use indigo (`#4f46e5`). Users see different “product colors” on sign-in vs. billing — undermines template polish and clone confidence.
- **Maintainability:** Hardcoded hex values are scattered across mobile screens (`HomeScreen`, `LoginScreen`, `SignupScreen`, `DashboardScreen`, `ProfileScreen`, dashboard components) and `billingColors`, with inconsistent casing (`#f9fafb` vs `#F9FAFB`) for the same color.
- **Clone / fork cost:** Template adopters who rebrand must hunt dozens of literals across web and mobile. A shared token file is the same pattern we already use for layout (`packages/shared/src/config/layoutWidth.ts`).
- **Cross-platform parity:** Mobile cannot share CSS with web, but both can import the same TS palette — one change updates both surfaces.

## Current state

| Surface | Brand / CTA | Page background |
|---------|-------------|-----------------|
| Web auth/landing (`primary-*` in `apps/web/tailwind.config.js`) | Blue `#3b82f6` | Tailwind `gray-*` |
| Web billing/dashboard (`indigo-*`) | Indigo `#4f46e5` | Tailwind `gray-*` |
| Mobile auth screens | Blue `#3b82f6` | `#f9fafb` |
| Mobile shared + billing (`Button.native`, `billing/styles.ts`) | Indigo `#4f46e5` | `#f9fafb` |

## Recommended approach

1. **Add** `packages/shared/src/theme/colors.ts` with:
   - Scales: `gray`, `indigo` (plus billing banner hues currently only in `apps/mobile/src/screens/billing/styles.ts`).
   - Semantic aliases: `pageBg`, `cardBg`, `border`, `rowDivider` (`#f3f4f6`), `textPrimary`, `textMuted`, `brand`, `brandDark`, and billing `error*` / `info*` / `warn*`.

2. **Wire web:** Import palette in `apps/web/tailwind.config.js` and map `primary` → shared indigo scale (keeps existing `primary-*` class names; values unify at config layer). Override `gray` and `indigo` from the same source.

3. **Wire mobile:** Refactor `billingColors` to re-export shared semantics; replace hardcoded hex in billing `rowBetween` divider and all grep-listed consumers.

4. **Brand decision:** Use **indigo as canonical brand** (`#4f46e5`). Auth/landing will shift from blue to indigo — intentional unification with billing/dashboard.

### Files to touch (non-exhaustive)

- `packages/shared/src/theme/colors.ts` (new)
- `apps/web/tailwind.config.js`
- `apps/mobile/src/screens/billing/styles.ts`
- Mobile screens: `HomeScreen`, `LoginScreen`, `SignupScreen`, `DashboardScreen`, `ProfileScreen`
- Mobile dashboard: `CollectionDetail`, `BooleanFeatureTiles`, `UsageStrip`
- Shared native: `Button.native.tsx`, `AppHeader.native.tsx`, `ProtectedRoute.native.tsx`

### Out of scope

- Replacing every web `indigo-*` class with `primary-*` (unnecessary once scales share source)
- Refactoring auth screens to use shared `Button` (nice follow-up)

## Acceptance criteria

- [ ] No hardcoded brand/page hex in migrated mobile files; all reference shared tokens
- [ ] `billingColors.rowDivider` (or equivalent) replaces `#f3f4f6` literal in `rowBetween`
- [ ] Tailwind `primary-*`, `gray-*`, and `indigo-*` derive from `packages/shared/src/theme/colors.ts`
- [ ] `npm run type-check` passes for `packages/shared`, `apps/web`, `apps/mobile`
- [ ] Smoke-test: web Login/Signup + billing plans; mobile Home/Login + billing overview

## Origin

AI code review finding on `apps/mobile/src/screens/HomeScreen.tsx` and related color drift across the monorepo.

mei-artificerinnovations:
## Summary

Closes #243

Creates `packages/shared/src/theme/colors.ts` as the single source of truth for all brand and semantic colors across web and mobile, and wires it into every file that had hardcoded hex values.

### What changed

**New file: `packages/shared/src/theme/colors.ts`**
- Gray and indigo scales (`colors.gray`, `colors.indigo`)
- Semantic surface tokens: `pageBg`, `cardBg`, `border`, `rowDivider`
- Text hierarchy: `textPrimary` → `textFaint`
- Brand: `brand` (#4f46e5), `brandDark`, `brandLight`
- Status: error/success/warn/info families; `errorBadgeBg` (red-100) separate from `errorBg` (red-50)
- Feature gate UI: `featureOnBg/Border/IconBg/Icon`, `featureBOnBg`, `featureOffBg`, `iconOffBg`
- Code display: `codeBg`, `codeText`, `codeValTrue`, `codeValFalse`
- Flask illustration: `iconBg/Fill/Stroke/Neck/Liquid`

**Web: `tailwind.config.js` → `tailwind.config.ts`**
- Renamed to `.ts` so jiti handles the full TypeScript import graph
- `primary`, `gray`, and `indigo` color scales now import from `colors.indigo`/`colors.gray` — future changes to `colors.ts` propagate to all `gray-*`/`indigo-*`/`primary-*` Tailwind classes automatically
- Fixes blue→indigo brand divergence between auth/landing and billing/dashboard

**Mobile screens** (`HomeScreen`, `LoginScreen`, `SignupScreen`, `DashboardScreen`, `ProfileScreen`)
- All `#3b82f6` (blue) brand colors replaced with `colors.brand` (indigo)
- Semantic gray values replaced with `colors.pageBg`, `colors.textMuted`, etc.
- `DashboardScreen` badge background: `colors.warnBg` (was hardcoded `#fef3c7`)

**Shared components** (`AppHeader.native`, `Button.native`, `ProtectedRoute.native`)
- `AppHeader.native`: SVG flask uses icon token references
- `Button.native`: `variantColors` returns `{ bg, text, border, borderWidth }` — all variant shape logic consolidated
- `ProtectedRoute.native`: `ActivityIndicator` and container bg use tokens

**Dashboard components** (`UsageStrip`, `CollectionDetail`, `BooleanFeatureTiles`)
- `UsageStrip` `limitBadge`: `colors.errorBadgeBg` (red-100, not lighter `errorBg`)
- `CollectionDetail` `featBtnOff`: `colors.featureOffBg` (not `rowDivider`)
- `BooleanFeatureTiles` `iconOff`: `colors.iconOffBg`; code block colors fully tokenized

**`apps/mobile/src/screens/billing/styles.ts`**
- `billingColors` now delegates to shared tokens
- Adds `rowDivider` token (was hardcoded `#f3f4f6` on `rowBetween.borderBottomColor`)

### Import paths

- Mobile: `@beakerstack/shared/theme/colors` (resolved via existing tsconfig path mapping)
- Shared package internals: `../../theme/colors` (relative)
- Tailwind config: `../../packages/shared/src/theme/colors` (relative, processed by jiti)

## Test plan

- [ ] Web: do a visual pass of any page using `primary-*`, `gray-*`, or `indigo-*` Tailwind classes — the scale values should be unchanged (jiti resolves `tailwind.config.ts` identically to the old `.js` file, but verify no class regressions)
- [ ] Web: `primary-*` Tailwind classes still apply correct indigo shades (not blue)
- [ ] Mobile: auth screens show indigo brand color (#4f46e5) instead of blue (#3b82f6)
- [ ] Mobile: billing `rowBetween` separator uses gray-100 row divider
- [ ] Mobile: `ActivityIndicator` across protected screens shows indigo
- [ ] Mobile: `UsageStrip` "Limit reached" badge is visibly red-100 (darker than the card bg)
- [ ] Mobile: feature gate off-state buttons use a distinct gray-100 surface, not the row divider shade
- [ ] TypeScript: no new type errors in modified files


packages/shared/src/theme/colors.ts
```diff
@@ -0,0 +1,89 @@
...
+  featureOnIconBg: '#dcfce7',
+  featureOnIcon: '#16a34a',
+  featureBOnBg: '#f3e8ff',
+
+  // Flask illustration tokens
+  iconBg: indigo[600],
+  iconFill: indigo[100],
+  iconStroke: '#8b5cf6',
+  iconNeck: indigo[200],
+  iconLiquid: indigo[400],
```
ZappoMan, line 117: **Brand consistency:** `iconStroke` is `#8b5cf6` (Tailwind violet-500), not an indigo step. The previous `AppHeader` SVG used the same value, so this preserves the illustration—but it diverges from the PR’s “indigo brand” story. Consider `indigo[400]` / a named token tied to the palette, or document why violet is intentional for the flask.

packages/shared/src/theme/colors.ts
```diff
@@ -0,0 +1,89 @@
...
+  errorBorder: '#fecaca',
+  errorText: '#991b1b',
+  errorTextAlt: '#b91c1c',
+  errorIcon: '#dc2626',
+  errorBorderBright: '#ef4444',
+
+  // Status: success
+  successBg: '#d1fae5',
+  successBorder: '#6ee7b7',
+  successText: '#065f46',
```
ZappoMan: **Token accuracy:** `successBg: '#d1fae5'` is closer to emerald-200 than green-50 (`#f0fdf4`). Nothing in this PR consumes it yet, but aligning status tokens to Tailwind’s green scale now avoids drift when someone reaches for `colors.successBg` later.

packages/shared/src/theme/colors.ts
```diff
@@ -0,0 +1,89 @@
...
+  400: '#818cf8',
+  500: '#6366f1',
+  600: '#4f46e5',
+  700: '#4338ca',
+  800: '#3730a3',
+  900: '#312e81',
+} as const;
+
+export const colors = {
+  gray,
```
ZappoMan, line 48: **Discoverability:** Consider a short module doc (import path + “semantic tokens for RN; Tailwind imports `colors.indigo`”) and/or a `package.json` `"exports"` entry for `./theme/colors` so resolution isn’t only via tsconfig path aliases.

apps/web/tailwind.config.ts
```diff
@@ -1,4 +1,5 @@
 import typography from '@tailwindcss/typography';
+import { colors } from '../../packages/shared/src/theme/colors';
```
ZappoMan, line 2: **Build check:** Importing TS from `tailwind.config.js` depends on Tailwind/jiti. Worth `pnpm --filter web build` in CI—if it fails on a clean machine, add the `.ts` extension and keep this file free of RN imports (already true).

apps/web/tailwind.config.ts
```diff
@@ -21,18 +22,7 @@ export default {
...
-          200: '#bfdbfe',
-          300: '#93c5fd',
-          400: '#60a5fa',
-          500: '#3b82f6',
-          600: '#2563eb',
-          700: '#1d4ed8',
-          800: '#1e40af',
-          900: '#1e3a8a',
-        },
+        primary: colors.indigo,
```
ZappoMan, line 30: **High-value change:** `primary: colors.indigo` makes existing `primary-600` / `hover:bg-primary-700` classes match mobile `colors.brand`—this fixes the auth/landing vs billing blue divergence.

apps/mobile/src/components/dashboard/BooleanFeatureTiles.tsx
```diff
@@ -90,31 +91,31 @@ const styles = StyleSheet.create({
...
+  iconOn: { backgroundColor: colors.featureOnIconBg },
+  iconOff: { backgroundColor: colors.border },
   iconPlaceholder: { width: 14, height: 14 },
-  iconText: { fontSize: 12, fontWeight: '800', color: '#9ca3af' },
-  iconTextOn: { color: '#16a34a' },
+  iconText: { fontSize: 12, fontWeight: '800', color: colors.textFaint },
+  iconTextOn: { color: colors.featureOnIcon },
   codeBox: {
     marginTop: 10,
     backgroundColor: '#0f172a',
```
ZappoMan: **Incomplete migration:** `codeBox` / `codeText` / `codeValOn` / `codeValOff` still use slate/green hex. Consider gray/slate tokens (or add a small `code` group) so this file is fully on the shared palette.

apps/mobile/src/screens/DashboardScreen.tsx
```diff
@@ -340,42 +341,42 @@ const styles = StyleSheet.create({
     paddingVertical: 2,
     borderRadius: 4,
   },
-  badgeText: { fontSize: 10, fontWeight: '600', color: '#92400e' },
+  badgeText: { fontSize: 10, fontWeight: '600', color: colors.warnText },
```
ZappoMan, line 344: Nice use of `colors.warnText` for the badge label. `badge.backgroundColor` is still `#fef3c7` while `colors.warnBg` exists—small follow-up for consistency.

packages/shared/src/components/primitives/Button.native.tsx
```diff
@@ -64,7 +65,7 @@ export function Button({
 }: ButtonProps) {
   const isDisabled = disabled || loading;
   const { v, h, font } = sizePadding[size];
-  const colors = variantColors(variant);
+  const vc = variantColors(variant);
```
ZappoMan, line 69: **Good catch:** `colors` → `vc` avoids shadowing the theme import—easy regression without this rename.

apps/mobile/src/screens/billing/styles.ts
```diff
@@ -1,22 +1,24 @@
...
-  warnBg: '#fffbeb',
-  warnBorder: '#fde68a',
-  warnText: '#92400e',
+  pageBg: colors.pageBg,
+  cardBg: colors.cardBg,
+  border: colors.border,
+  rowDivider: colors.rowDivider,
+  textPrimary: colors.textPrimary,
+  textMuted: colors.textMuted,
+  indigo: colors.brand,
```
ZappoMan, line 11: **Alias naming:** `billingColors.indigo` → `colors.brand` keeps call sites stable while fixing the hue. `rowDivider` on `rowBetween` is a nice semantic addition.

ZappoMan (commented):
## Review summary

Solid step toward a single color source of truth; the blue → indigo brand fix is the right call. The token file is clear, the `Button.native` shadowing fix is good, and Tailwind `primary` → `colors.indigo` aligns web with mobile.

Main gaps: migration is partial (several dashboard files still hardcode hex), a couple of semantic tokens look unused or slightly off, and there is no `package.json` subpath export yet—worth confirming Metro/Vite resolution in CI.

Inline notes on the highest-impact items below.

sarah-artificerinnovations (approved):
Review: PR #249 — centralize theme colors in @beakerstack/shared

Verdict: APPROVE (pending test run)

Clean, well-structured centralization. No blockers. A few things worth calling out:

### colors.ts values
All gray and indigo scale values match Tailwind defaults exactly. The semantic token names (textPrimary/Secondary/Body/Subtle/Muted/Faint) form a sensible hierarchy. One minor observation: errorText (#991b1b = red-800) is darker than errorTextAlt (#b91c1c = red-700), which reads as "Alt is lighter" rather than the usual convention of Alt being a variant. Not a bug — both were already in the codebase with these values — but worth a comment if this token set grows.

### Tailwind primary scale: blue → indigo
The old primary scale was blue (#3b82f6 at 500); the new one is colors.indigo (#4f46e5 at 600). This is the right fix — web and mobile are now using the same brand hue. The import of a .ts file from tailwind.config.js works via Vite/jiti (confirmed: deploy-preview check passed, which means the build succeeded).

The colors.indigo keys are numeric literals (50, 100, ...). JavaScript coerces numeric object keys to strings at runtime, so Tailwind sees { "50": "#eef2ff", ... } and generates primary-50, primary-100, etc. correctly.

### Button.native.tsx rename (colors → vc)
Renaming the local variable from colors to vc to avoid shadowing the import is the right call. The rename is purely mechanical — all six references updated consistently.

### billingColors delegation
All existing billingColors string values are preserved: billingColors.indigo delegates to colors.brand (#4f46e5), same hex as before. BillingTabBar (not modified) will continue working. The one addition — rowDivider: colors.rowDivider — fills a slot that was previously hardcoded directly in the rowBetween style (#f3f4f6 = gray-100); now it gets a token name.

### CollectionDetail toast colors
The toast background changed from "#eef2ff" to colors.indigo[50] — same hex (#eef2ff). The border changed from "#c7d2fe" to colors.indigo[200] (#c7d2fe). Both are identity replacements. The toastText changed from "#3730a3" to colors.indigo[800] (#3730a3). All correct.

### Tests
Still in_progress at review time. Hold merge until green. Everything else (deploy-preview, Lighthouse, CodeQL) passed.

packages/shared/src/theme/colors.ts
```diff
@@ -0,0 +1,89 @@
...
+  brandDark: indigo[700],
+  brandLight: indigo[100],
+
+  // Status: error
+  errorBg: '#fef2f2',
+  errorBorder: '#fecaca',
+  errorText: '#991b1b',
+  errorTextAlt: '#b91c1c',
+  errorIcon: '#dc2626',
+  errorBorderBright: '#ef4444',
```
rahul-artificerinnovations: `errorBorderBright` (#ef4444) is never referenced in this PR or in any of the changed files. Either wire it up somewhere (e.g. `Button.native` destructive border?) or remove it — dead tokens in a new canonical file add noise and raise questions for future readers.

packages/shared/src/theme/colors.ts
```diff
@@ -0,0 +1,89 @@
...
+  900: '#312e81',
+} as const;
+
+export const colors = {
+  gray,
+  indigo,
+
+  // Surfaces
+  pageBg: gray[50],
+  cardBg: '#ffffff',
```
rahul-artificerinnovations: Nit: `cardBg` is the only surface token that hard-codes a hex string rather than deriving from the `gray` scale. Could be `'#fff'` for brevity or left as-is; just flag that it breaks the consistent derivation pattern in this section.

apps/mobile/src/screens/billing/styles.ts
rahul-artificerinnovations, line 11: `indigo: colors.brand` — the field name `indigo` now holds `colors.brand` (which is `indigo[600]` under the hood, so it's accurate), but a reader scanning `billingColors.indigo` might expect a palette entry, not a semantic alias. Not a blocker given this is a pre-existing name, but worth a follow-up to rename to `brand` (matching its callers' intent) when you next touch this file.

apps/web/tailwind.config.ts
rahul-artificerinnovations, line 30: This changes `primary-*` across the entire web app from blue (#3b82f6) to indigo (#4f46e5). The test plan should include an explicit visual pass over the landing page, marketing CTAs, auth forms, and any page that uses `text-primary-*` / `bg-primary-*` / `border-primary-*` classes — not just a smoke-check that the classes resolve.

rahul-artificerinnovations (commented):
## Review: feat(shared): centralize theme colors

Clean, well-scoped refactor. The new `colors.ts` is the right move — single source of truth, sensible semantic hierarchy (surfaces → text → brand → status), and the `billingColors` delegation means billing UI stays on the same tokens without a big callsite migration. A few things worth addressing before merge:

**Bigger concern — Tailwind `primary` scale is a visually breaking change**

Every web element that uses `primary-*` classes (buttons, links, badges, auth pages, landing CTAs) shifts from blue (#3b82f6) to indigo (#4f46e5). This is intentionally fixing a blue/indigo divergence, but the test plan only says "primary-* classes still apply" — it doesn't explicitly call for a visual pass over the landing page, auth flows, and marketing copy. Please add that to the test plan before merge.

**Minor issues (inline below):**
- `errorBorderBright` in `colors.ts` is defined but unreferenced in this PR.
- `billingColors.indigo` field name now delegates to `colors.brand` — the mismatch between key name and value semantics is a pre-existing issue this PR surfaces more visibly.

Otherwise the implementation looks correct — token values match the originals, `Button.native` variable rename to `vc` avoids the shadowing correctly, and the flask SVG token chain is tidy.

diego-artificerinnovations (approved):
Solid work overall — the centralization is clean, the token naming is semantic and consistent across web and native, and the blue→indigo brand alignment is a meaningful correctness fix.

Three non-blocking observations:

**1. `billingColors.indigo` / `billingColors.indigoDark` naming in `styles.ts`**

The alias layer maps `colors.brand → billingColors.indigo`. That works today, but callers of `billingStyles` are now coupling to a color name (`indigo`) instead of a semantic role (`brand`). If the brand color changes — or gets mapped to something that isn't indigo — the alias becomes a lie without callers noticing. Renaming to `brand`/`brandDark` here would stay consistent with the shared token system.

**2. `borderWidth` in `Button.native.tsx` is still caller-side logic**

```tsx
borderWidth: variant === 'secondary' ? 1 : 0,
borderColor: vc.border,
```

`vc.border` comes from `variantColors`, but `borderWidth` is still inlined. The border visibility decision is split across two places. Adding `borderWidth` to the `variantColors` return would keep all variant shape logic in one spot.

**3. `iconStroke: '#8b5cf6'` in `colors.ts`**

This is the only hardcoded hex in the token file that's neither gray nor indigo — it's violet-500, intentionally distinct from `indigo[500]` (`#6366f1`) for the flask illustration. A one-line comment (`// violet-500: warmer stroke against indigo fill`) would save the next person from wondering if this was an oversight.

No blockers — approve as-is.

ZappoMan:
## Recommendation: adopt `palette` + `semantic` instead of flat `colors`

Thanks for centralizing theme colors and fixing the blue → indigo brand drift — that’s the right fix for #243. After reviewing the approach against a local spike on the same files, I’d recommend **refactoring this PR** to use a two-layer model before merge rather than shipping the flat `export const colors` object.

### Why split palette vs semantic?

The current PR merges **scale steps** and **UI roles** on one object (`colors.gray`, `colors.pageBg`, `colors.featureOnBg`, `colors.iconStroke`, …). That works short-term, but it makes the token file grow awkwardly:

- **Scale steps** belong in a palette (Tailwind-like): `palette.indigo[200]`, `palette.green[50]`, `palette.slate[900]`.
- **Product meaning** belongs in semantic: `semantic.pageBg`, `semantic.brand`, `semantic.destructive`.

Keeping them separate makes it obvious whether a consumer should use a **role** (`semantic.textMuted`) or pick a **step** when the design is inherently “use indigo-50 for this toast” (`palette.indigo[50]`).

### Suggested shape for `packages/shared/src/theme/colors.ts`

```ts
/** Canonical colors — import via `@beakerstack/shared/theme/colors`. */

export const palette = {
  white: '#ffffff',
  gray: { /* 50–900 */ },
  indigo: { /* 50–900 */ },
  // Add scales only as needed (sparse is fine):
  red: { 50, 100, 200, 600, 700, 800 },
  blue: { 50, 200, 900 },
  amber: { 50, 100, 200, 800 },
  green: { 50, 100, 300, 400, 600 },
  slate: { 300, 500, 900 },
  purple: { 100 },
} as const;

/** Semantic tokens for RN StyleSheet / billing wrappers. */
export const semantic = {
  pageBg: palette.gray[50],
  cardBg: palette.white,
  border: palette.gray[200],
  rowDivider: palette.gray[100],
  textPrimary: palette.gray[900],
  textMuted: palette.gray[500],
  heading: palette.gray[800],
  brand: palette.indigo[600],
  brandDark: palette.indigo[700],
  brandLight: palette.indigo[100],
  destructive: palette.red[600],
  errorBg: palette.red[50],
  errorBorder: palette.red[200],
  errorText: palette.red[800],
  infoBg: palette.blue[50],
  infoBorder: palette.blue[200],
  infoText: palette.blue[900],
  warnBg: palette.amber[50],
  warnBorder: palette.amber[200],
  warnText: palette.amber[800],
} as const;
```

**Tailwind** should import `palette` only:

```js
import { palette } from '../../packages/shared/src/theme/colors.ts';

theme: {
  extend: {
    colors: {
      primary: palette.indigo,
      gray: palette.gray,
      indigo: palette.indigo,
    },
  },
},
```

Also add `../../packages/shared/src/theme/**/*.{js,ts}` to `content` so JIT sees theme files if classes are referenced there later.

### Mapping from this PR → palette/semantic

| Current PR token | Suggested replacement |
|------------------|----------------------|
| `colors.pageBg`, `colors.border`, `colors.brand`, … | `semantic.*` |
| `colors.textSecondary`, `textBody`, `textSubtle`, `textFaint` | `semantic.heading` or `palette.gray[700/600/400]` — avoid duplicating a full text ladder in semantic unless we standardize names app-wide |
| `colors.indigo[n]` (toast, etc.) | `palette.indigo[n]` |
| `colors.featureOnBg`, `featureOnBorder`, … | `palette.green[50]`, `palette.green[300]`, etc. — no one-off `featureOn*` keys |
| `colors.iconBg`, `iconStroke`, `iconLiquid`, … | `semantic.brand`, `palette.indigo[500]`, `palette.indigo[400]` — drops violet `#8b5cf6` stroke so the flask matches indigo brand |
| `colors.errorIcon` | `semantic.destructive` |
| `colors.successBg` (currently wrong hex) | `palette.green[50]` when needed — don’t pre-declare unused status tokens |

### Consumer import pattern

```ts
import { palette, semantic } from '@beakerstack/shared/theme/colors';

// Roles
backgroundColor: semantic.pageBg,
color: semantic.textMuted,

// Scale-specific UI (toasts, tiles, code blocks)
backgroundColor: palette.indigo[50],
borderColor: palette.indigo[200],
backgroundColor: palette.slate[900],  // code box
```

`billingColors` can keep delegating to `semantic` (same aliases as today: `indigo` → `semantic.brand`).

### Small finish items while touching files

These are easier with `palette` in place:

1. **`BooleanFeatureTiles`** — migrate `codeBox` / `codeText` / `codeValOn|Off` (still hex in this PR) to `palette.slate` / `palette.green`.
2. **`AppHeader` flask** — use `palette.indigo[*]` for stroke/fill instead of `iconStroke: '#8b5cf6'`.
3. **`DashboardScreen` demo badge** — `palette.amber[100]` for bg + `semantic.warnText` for label (PR already has warn text token; bg is still `#fef3c7`).
4. **Prefer `palette.white`** over `'#ffffff'` / `'#fff'` in `Button.native` and similar.

### What to drop from the flat `colors` object

- Unused or inaccurate pre-declared tokens (`successBg`, `errorTextAlt`, `errorBorderBright`, etc.) unless a consumer in this PR needs them.
- Domain-specific keys used once (`featureBOnBg`, `iconNeck`) — use palette steps at the call site instead.

### Scope (unchanged)

Either approach still leaves `CollectionsGrid`, `FeatureGateCard`, `AnnotatedPrimitive`, `SocialLoginButton` on hex — fine for a phased #243 if called out in the PR body.

---

**Summary:** Approve the direction (shared module + indigo brand + wired consumers). Request a refactor to **`palette` + `semantic`** before merge — same migration scope, better long-term shape, and it closes the few half-migrated spots above without growing a flat `colors` bag of keys.

Happy to pair on the refactor or review a follow-up commit on this branch.

ZappoMan:
### Follow-up: `palette.brand` alias (template + forks)

The recommendation above didn’t spell out one piece of the model that matters for **BeakerStack as a forkable template**.

**Problem:** Naming the accent ramp `indigo` and wiring UI to `palette.indigo[600]` is accurate today, but it’s confusing for forks that want teal (or anything else)—you shouldn’t put teal hex values in a key called `indigo`.

**Pattern:** Keep **named hue ramps** (`indigo`, `teal`, …) and add a **`brand` alias** that points at the ramp this product uses:

```ts
const indigo = { 50: '…', /* … */ 600: '#4f46e5', /* … */ } as const;
const teal   = { 50: '…', /* … */ 600: '#0d9488', /* … */ } as const; // optional until a fork needs it

export const palette = {
  gray: { /* … */ },
  indigo,
  teal,           // can ship later; doesn’t have to exist in v1
  brand: indigo,  // template default — same object reference as `indigo`
} as const;

export const semantic = {
  brand: palette.brand[600],
  brandDark: palette.brand[700],
  brandLight: palette.brand[100],
  // surfaces/text/errors still from gray / red / amber …
} as const;
```

**Consumers:** Prefer `semantic.brand` and `palette.brand[n]` for accent UI (buttons, links, usage bars, toasts). Reserve `palette.indigo` only when you truly mean that hue, not “whatever the product accent is.”

**Tailwind:** Wire `primary` to the alias, not the hue name:

```js
colors: {
  primary: palette.brand,
  gray: palette.gray,
},
```

**Fork to teal (later):** Change one line in `palette` (and add `teal` if it isn’t there yet):

```ts
brand: teal,  // was: indigo
```

`semantic` and `primary-*` classes stay unchanged. No need to grep-replace “indigo” across the app if call sites already use `palette.brand` / `semantic.brand`.

**What stays non-brand:** `palette.green` for “feature on” / success tiles, `palette.red` for errors, `palette.gray` for neutrals—those don’t have to become teal when the brand does.

That’s the template goal behind the palette/semantic split: **hue ramps are ingredients; `brand` + `semantic` are which accent this fork uses.**

apps/mobile/src/components/dashboard/UsageStrip.tsx
```diff
@@ -201,53 +202,53 @@ const styles = StyleSheet.create({
   sectionTitle: {
     fontSize: 14,
     fontWeight: '600',
-    color: '#111827',
+    color: colors.textPrimary,
   },
   limitBadge: {
-    backgroundColor: '#fee2e2',
+    backgroundColor: colors.errorBg,
```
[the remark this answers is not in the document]
mei-artificerinnovations: Fixed in r2. Added `errorBadgeBg: '#fee2e2'` (red-100) as a dedicated token distinct from `errorBg` (red-50). `UsageStrip` `limitBadge` now uses `colors.errorBadgeBg` — matches the original hardcoded value exactly.

apps/mobile/src/components/dashboard/BooleanFeatureTiles.tsx
```diff
@@ -90,31 +91,31 @@ const styles = StyleSheet.create({
...
+  tilePlan: { marginTop: 2, fontSize: 11, color: colors.textMuted },
   iconWrap: {
     borderRadius: 999,
     paddingHorizontal: 8,
     paddingVertical: 4,
   },
-  iconOn: { backgroundColor: '#dcfce7' },
-  iconOff: { backgroundColor: '#e5e7eb' },
+  iconOn: { backgroundColor: colors.featureOnIconBg },
+  iconOff: { backgroundColor: colors.border },
```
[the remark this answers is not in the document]
mei-artificerinnovations: Fixed in r2. Added `featureOffBg: gray[100]` as a dedicated token (same hex as `rowDivider`, correct semantic name). `CollectionDetail` `featBtnOff` now uses `colors.featureOffBg`.

apps/mobile/src/components/dashboard/CollectionDetail.tsx
```diff
@@ -319,44 +320,44 @@ export function CollectionDetail({ collection, addItem, onActivity }: Props) {
...
     paddingVertical: 8,
   },
-  featBtnAOn: { backgroundColor: '#e0e7ff' },
-  featBtnBOn: { backgroundColor: '#f3e8ff' },
-  featBtnOff: { backgroundColor: '#f3f4f6' },
-  featBtnText: { fontSize: 11, fontWeight: '700', color: '#4338ca' },
-  featBtnTextOff: { color: '#9ca3af' },
+  featBtnAOn: { backgroundColor: colors.brandLight },
+  featBtnBOn: { backgroundColor: colors.featureBOnBg },
+  featBtnOff: { backgroundColor: colors.rowDivider },
```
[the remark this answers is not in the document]
mei-artificerinnovations: Removed in r2. `errorBorderBright` was defined but never consumed in this PR — dead token. Dropped from `colors.ts`.

apps/web/tailwind.config.ts
```diff
@@ -21,18 +22,7 @@ export default {
...
       },
```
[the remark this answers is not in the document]
mei-artificerinnovations: Fixed in r2. `successBg` is now `'#f0fdf4'` (Tailwind green-50). The previous `#d1fae5` was emerald-100/200, not a standard Tailwind green scale value.

packages/shared/src/theme/colors.ts
```diff
@@ -0,0 +1,89 @@
...
+  featureOnBg: '#f0fdf4',
+  featureOnBorder: '#86efac',
+  featureOnIconBg: '#dcfce7',
+  featureOnIcon: '#16a34a',
+  featureBOnBg: '#f3e8ff',
+
+  // Flask illustration tokens
+  iconBg: indigo[600],
+  iconFill: indigo[100],
+  iconStroke: '#8b5cf6',
```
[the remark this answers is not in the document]
mei-artificerinnovations: Fixed in r2. Added `codeBg`, `codeText`, `codeValTrue`, `codeValFalse` tokens to `colors.ts` and wired them into `BooleanFeatureTiles` — no more hardcoded hex in the stylesheet.

packages/shared/src/theme/colors.ts
```diff
@@ -0,0 +1,89 @@
...
+  infoBg: '#eff6ff',
+  infoBorder: '#bfdbfe',
+  infoText: '#1e3a8a',
+
+  // Feature gate UI
+  featureOnBg: '#f0fdf4',
+  featureOnBorder: '#86efac',
+  featureOnIconBg: '#dcfce7',
+  featureOnIcon: '#16a34a',
+  featureBOnBg: '#f3e8ff',
```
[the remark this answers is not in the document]
mei-artificerinnovations: Fixed in r2. Renamed `tailwind.config.js` → `tailwind.config.ts` and deleted the old `.js` file. jiti (Tailwind v3's TS loader) resolves `.ts` configs and their transitive `import` graph, eliminating the ESM→TS resolution gap. The old `.js` file is removed in this commit.

apps/web/tailwind.config.ts
```diff
@@ -1,4 +1,5 @@
 import typography from '@tailwindcss/typography';
+import { colors } from '../../packages/shared/src/theme/colors';
```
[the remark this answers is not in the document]
mei-artificerinnovations: Intentional — kept as-is. The original SVG used `#8b5cf6` (violet-500) for a warmer, slightly purple-tinted stroke against the cooler indigo fill. Added a clarifying comment in r2: `// violet-500 — warmer stroke against indigo fill`. Happy to revisit if the design direction changes, but aligning it to indigo would make the flask icon feel flatter.

packages/shared/src/theme/colors.ts
mei-artificerinnovations: Fixed in r2. Added `iconOffBg: gray[200]` as an explicit token and `BooleanFeatureTiles` `iconOff` now uses `colors.iconOffBg`. Using `colors.border` here was a semantic mismatch — borders and icon badge backgrounds shouldn't share a token.

packages/shared/src/theme/colors.ts
mei-artificerinnovations: Fixed in r2. `codeBox.backgroundColor` now uses `colors.codeBg` (`#0f172a`, slate-900).

packages/shared/src/theme/colors.ts
mei-artificerinnovations: Fixed in r2. `codeText.color` now uses `colors.codeText` (`#cbd5e1`, slate-300).

apps/web/tailwind.config.ts
mei-artificerinnovations: Fixed in r2. `codeValOn` now uses `colors.codeValTrue` (`#4ade80`, green-400).

apps/web/tailwind.config.ts
mei-artificerinnovations: Fixed in r2. `codeValOff` now uses `colors.codeValFalse` (`#64748b`, slate-500).

apps/mobile/src/components/dashboard/BooleanFeatureTiles.tsx
mei-artificerinnovations: Fixed in r2. `variantColors` now returns `borderWidth: number` alongside `bg`, `text`, and `border`. All variant shape logic lives in one function — no more inline ternary. `TouchableOpacity` style reads `borderWidth: vc.borderWidth`.

apps/mobile/src/screens/DashboardScreen.tsx
mei-artificerinnovations: Fixed in r2. `badge.backgroundColor` now uses `colors.warnBg` (`#fffbeb`, amber-50).

packages/shared/src/components/primitives/Button.native.tsx
mei-artificerinnovations: Agreed on the naming. `billingColors.indigo` is a pre-existing field name with callsites outside this PR — renaming it here would expand the diff scope. I'll file a follow-up to rename it to `billingColors.brand` or `billingColors.primary` once this lands, so it can get proper cross-file review coverage.

apps/mobile/src/screens/billing/styles.ts
mei-artificerinnovations: Left as `'#ffffff'` — there's no gray scale equivalent for pure white (gray-50 is `#f9fafb`, already used as `pageBg`). Added a comment in r2: `// pure white — no gray scale equivalent`.

packages/shared/src/theme/colors.ts
mei-artificerinnovations: Agreed this belongs in `package.json` exports long-term. Keeping it out of scope for this PR — adding the `exports` field touches the package boundary and should get its own review. Added a module doc comment to `colors.ts` in r2 that documents both import paths explicitly until that follow-up lands.

packages/shared/src/theme/colors.ts
mei-artificerinnovations: Added in r2. Top of `colors.ts` now has a doc comment explaining the role of the file, both import paths (mobile tsconfig alias vs. Tailwind direct path), and the relationship between `colors.indigo` and the Tailwind `primary` scale.

apps/mobile/src/screens/billing/styles.ts
mei-artificerinnovations: The comment there (`// Emitted from layoutWidth.ts...`) is existing — not added in this PR. Happy to strip it in a cleanup pass. The safelist entries themselves are load-bearing for JIT; the comment is just explanatory context for whoever asks why those width tokens are pinned.

apps/web/tailwind.config.ts
mei-artificerinnovations: Added in r2 as an inline comment: `// primary-* → indigo, aligning web with mobile colors.brand (#4f46e5)`. The intent is to make it explicit that `primary-600` in Tailwind classes resolves to the same `#4f46e5` as `colors.brand` on mobile — the brand is unified.

packages/shared/src/components/primitives/Button.native.tsx
ZappoMan: @mei-artificerinnovations please create a ticket for any future work like this.

packages/shared/src/theme/colors.ts
ZappoMan: @mei-artificerinnovations please add to appropriate follow up ticket

apps/mobile/src/screens/billing/styles.ts
ZappoMan: @mei-artificerinnovations - I agree with Rahul that this reads strangely, I think we should either definitely fix this or at a minimum add a task to the followup ticket.

packages/shared/src/theme/colors.ts
```diff
@@ -0,0 +1,109 @@
...
+  900: '#312e81',
+} as const;
+
+export const colors = {
+  gray,
+  indigo,
+
+  // Surfaces
+  pageBg: gray[50],
+  cardBg: '#ffffff', // pure white — no gray scale equivalent
```
ZappoMan: **Before merge:** Add `white: '#ffffff'` and set `cardBg: white` so components don't keep sprinkling `'#fff'` / `'#ffffff'`. Makes the next stray-hex grep much easier.

packages/shared/src/theme/colors.ts
```diff
@@ -0,0 +1,109 @@
...
+  // Status: error
+  errorBg: '#fef2f2',       // red-50 — card/banner background
+  errorBadgeBg: '#fee2e2',  // red-100 — badge/pill background (slightly stronger)
+  errorBorder: '#fecaca',   // red-200
+  errorText: '#991b1b',     // red-800
+  errorTextAlt: '#b91c1c',  // red-700 — lighter variant, e.g. inline message body
+  errorIcon: '#dc2626',     // red-600 — icon and err text
+
+  // Status: success
+  successBg: '#f0fdf4',     // green-50
```
ZappoMan: **DRY:** `successBg` and `featureOnBg` are the same green-50 hex — consider `featureOnBg: successBg` (or one shared const) so a future tweak can't update one and miss the other.

packages/shared/src/theme/colors.ts
```diff
@@ -0,0 +1,109 @@
...
+  300: '#a5b4fc',
+  400: '#818cf8',
+  500: '#6366f1',
+  600: '#4f46e5',
+  700: '#4338ca',
+  800: '#3730a3',
+  900: '#312e81',
+} as const;
+
+export const colors = {
```
ZappoMan, line 47: **Follow-up (template forks):** Split into `palette` + `semantic` and `palette.brand: indigo` so forks can set `brand: teal` without teal hex under a key named `indigo`. Not blocking this PR — happy as a follow-up issue.

packages/shared/src/theme/colors.ts
```diff
@@ -0,0 +1,109 @@
...
+  textMuted: gray[500],
+  textFaint: gray[400],
+
+  // Brand
+  brand: indigo[600],
+  brandDark: indigo[700],
+  brandLight: indigo[100],
+
+  // Status: error
+  errorBg: '#fef2f2',       // red-50 — card/banner background
```
ZappoMan, line 72: **Optional:** Error/warn/info/code hex are still free-floating while gray/indigo use scales. Sparse `palette.red` / `amber` / `green` / `slate` would let `errorBadgeBg` be `red[100]`, etc. Defer if you want to ship #243 now.

packages/shared/src/components/primitives/Button.native.tsx
```diff
@@ -34,17 +35,18 @@ function variantColors(variant: ButtonVariant): {
   bg: string;
   text: string;
   border: string;
+  borderWidth: number;
 } {
   switch (variant) {
     case 'primary':
-      return { bg: '#4F46E5', text: '#FFFFFF', border: 'transparent' };
+      return { bg: colors.brand, text: '#ffffff', border: 'transparent', borderWidth: 0 };
```
ZappoMan: You migrated bg to tokens — please use `colors.cardBg` or `colors.white` for `text` here (and destructive below) instead of `'#ffffff'` so this file is fully on the token module.

apps/mobile/src/components/dashboard/UsageStrip.tsx
```diff
@@ -201,53 +202,53 @@ const styles = StyleSheet.create({
...
   },
   actions: { marginTop: 16, flexDirection: 'row', flexWrap: 'wrap', gap: 12 },
   btnPrimary: {
-    backgroundColor: '#4f46e5',
+    backgroundColor: colors.brand,
     paddingHorizontal: 16,
     paddingVertical: 10,
     borderRadius: 8,
   },
   btnPrimaryText: { color: '#fff', fontWeight: '600', fontSize: 14 },
```
ZappoMan: `btnPrimaryText` still uses `'#fff'` — switch to shared white/`cardBg` token for consistency with the rest of this PR.

apps/web/tailwind.config.ts
```diff
@@ -21,18 +22,12 @@ export default {
...
-          700: '#1d4ed8',
-          800: '#1e40af',
-          900: '#1e3a8a',
-        },
+        // Pin gray and indigo Tailwind scales to shared tokens so future
+        // changes to colors.ts propagate to gray-* / indigo-* classes automatically.
+        gray: colors.gray,
+        indigo: colors.indigo,
+        // primary-* → indigo, aligning web with mobile colors.brand (#4f46e5)
+        primary: colors.indigo,
```
ZappoMan, line 30: Looks correct. **Follow-up:** when you add `palette.brand`, wire `primary: palette.brand` so forks only swap the alias; `colors.indigo` is fine until then.

ZappoMan (commented):
## Review (r3) — looks good visually; a few consistency fixes

Thanks for r2 — manual web/mobile passes look good; the indigo migration and the new tokens (`errorBadgeBg`, `featureOffBg`, code block, etc.) land well.

### Before merge (this PR)
- Add a shared **white** token and replace remaining `'#fff'` / `'#ffffff'` in files this branch already touches (`Button.native`, `UsageStrip`, auth screens, `billing/styles.ts`).
- Optional DRY: `featureOnBg` / `successBg` share the same hex.

### Follow-up (non-blocking)
- `palette` + `semantic` + `palette.brand` alias for forkable templates ([#243](https://github.com/Artificer-Innovations/BeakerStack/issues/243) follow-up).
- Add `theme/**` to Tailwind `content` if we start emitting classes from shared theme files.
- Migrate `CollectionsGrid`, `FeatureGateCard`, `AnnotatedPrimitive` (still hardcoded hex).

Inline notes on specific lines below.

ZappoMan:
### Additional inline-style notes (files without diff anchors)

GitHub wouldn't attach these to unchanged lines, but they're the same **before-merge** white-token pass:

| File | Location |
|------|----------|
| `HomeScreen.tsx` | `primaryButtonText` (`#ffffff`) |
| `LoginScreen.tsx` | `ActivityIndicator` + `primaryButtonText` |
| `SignupScreen.tsx` | same as Login |
| `billing/styles.ts` | `primaryBtnText` (`#fff`) |
| `tailwind.config.ts` | add `../../packages/shared/src/theme/**/*.{js,ts}` to `content` (JIT safety) |

Once `colors.white` exists, these are quick one-liners.

packages/shared/src/theme/colors.ts
mei-artificerinnovations: Done in r3. Added `const white = '#ffffff' as const` before the `colors` object (same pattern as `gray`/`indigo`), exported as `colors.white`, and switched `cardBg: white`. See below for `Button.native` and `UsageStrip` followthrough.

packages/shared/src/theme/colors.ts
mei-artificerinnovations: Fixed in r3. Extracted `const green50 = '#f0fdf4' as const` and both `successBg` and `featureOnBg` now reference it. Added a short comment explaining the intent so a future reader knows the alias is deliberate.

packages/shared/src/theme/colors.ts
mei-artificerinnovations: Tracked in #250 — "colors.ts: split into palette + semantic layers with palette.brand alias". Covers the `palette.brand = palette.indigo` alias so forks swap one line, and the Tailwind `primary: palette.brand` wiring (which @ZappoMan also mentioned on `tailwind.config.ts:30`). Not blocking this PR.

packages/shared/src/theme/colors.ts
mei-artificerinnovations: Tracked in #251 — "colors.ts: add sparse red/amber/green/slate sub-scales for status tokens". Agreed this would make the token definitions read as clean named steps (e.g. `errorBadgeBg: red[100]`) rather than free-floating hex. Deferring to keep #243 shippable now — pure refactor with zero callsite impact.

packages/shared/src/components/primitives/Button.native.tsx
mei-artificerinnovations: Fixed in r3. Both `primary` and `destructive` variants now use `colors.white` for `text`. `Button.native.tsx` is now fully on the token module.

apps/mobile/src/components/dashboard/UsageStrip.tsx
mei-artificerinnovations: Fixed in r3. `btnPrimaryText.color` is now `colors.white`.

apps/web/tailwind.config.ts
mei-artificerinnovations: Captured in #250 (palette+semantic split). Once `palette.brand` lands, the Tailwind config update to `primary: palette.brand` is a one-liner — noted in the issue so it doesn't get lost.

PR #253, opened 2026-05-17
feat(release): CalVer notes from promotion PR template and workflow

ZappoMan:
## Summary

Automates CalVer GitHub Release bodies from the merged `develop`→`main` promotion PR, fixes git-cliff tag anchoring, and adds a Cursor project skill to draft high-quality promotion PRs.

### Changes

**Release automation**
- **Promotion PR template** (`.github/PULL_REQUEST_TEMPLATE/promote-develop-to-main.md`) — Option C: `## CalVer release notes` above `---`, maintainer checklist below (published vs maintainer-only split)
- **Release Template workflow** — resolves promotion PR (`develop`→`main`), extracts CalVer section, appends git-cliff changelog since latest prior `YYYY.NNN` tag (any year, strictly before `NEW_TAG` for `override_tag`)
- **`scripts/extract-calver-release-notes.sh`** — parses PR body for the publishable CalVer block

**PR templates**
- Restored **`.github/pull_request_template.md`** as the auto-loaded default for normal PRs (multi-template `default.md` removed — GitHub does not treat it as default)
- Promotion template remains in `.github/PULL_REQUEST_TEMPLATE/` for `?template=promote-develop-to-main.md`

**Agent skill**
- **`.cursor/skills/promote-develop-to-main/SKILL.md`** — drafts Summary / Highlights / Adopter notes from `git log` + diff, opens PR via `gh pr create --body-file`
- **`.claude/skills/`** stays gitignored (local Stripe mirror only); BeakerStack skills live under `.cursor/skills/`

**Docs**
- `CONTRIBUTING.md`, `docs/VERSIONING.md`, `docs/DEVELOPMENT.md`, `.github/RELEASE_TEMPLATE.md`

### Review fixes (Copilot)

- `PREV_TAG` uses latest CalVer tag **strictly less than** `NEW_TAG` (safe `override_tag` backfills)
- Promotion PR auto-detect filters `commits/{sha}/pulls` for `develop`→`main` before falling back to `gh pr list`

### Test plan

- [ ] Normal PR to `develop` still gets default template body automatically
- [ ] Open promotion PR with `promote-develop-to-main` template (or skill); confirm `---` splits published vs maintainer sections
- [ ] `scripts/extract-calver-release-notes.sh` on a body with `## CalVer release notes`
- [ ] After merge to `develop`: ask agent to “promote develop to main” — skill drafts CalVer section and opens PR
- [ ] After promotion merge: Release Template on `main` publishes PR CalVer section + git-cliff changelog


.github/PULL_REQUEST_TEMPLATE/default.md
```diff
@@ -0,0 +1,42 @@
...
+- [ ] Tests added/updated (if needed)
+- [ ] All tests pass locally
+
+## Testing
+
+<!-- Describe how you tested your changes -->
+
+## Related Issues
+
+<!-- Link to related issues, if any -->
```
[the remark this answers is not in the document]
ZappoMan: Good catch — restored `.github/pull_request_template.md` as the auto-loaded default for normal PRs. The promotion template stays in `.github/PULL_REQUEST_TEMPLATE/promote-develop-to-main.md` for `?template=promote-develop-to-main.md`. Removed `default.md` from the multi-template folder since it has no special meaning to GitHub.

.github/workflows/release-template.yml
```diff
@@ -67,18 +75,97 @@ jobs:
...
+          gh pr view "${{ steps.promotion.outputs.number }}" --json body --jq .body > pr-body.md
+          bash scripts/extract-calver-release-notes.sh pr-body.md > calver-release-notes.md
+          echo "Extracted $(wc -l < calver-release-notes.md) lines from PR #${{ steps.promotion.outputs.number }}"
+
+      - name: Generate git-cliff changelog
+        run: |
+          set -euo pipefail
+          NEW_TAG="${{ steps.tag.outputs.tag }}"
+          # CalVer template tags only (excludes @beakerstack/* npm tags)
+          PREV_TAG="$(git tag -l | grep -E '^[0-9]{4}\.[0-9]{3}$' | sort -V | tail -1 || true)"
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 1e5bff7: `PREV_TAG` is now the latest `YYYY.NNN` tag **strictly less than** `NEW_TAG` via `sort -V | awk '$0 < t' | tail -1`, so `override_tag` backfills (e.g. cutting `2026.002` when `2026.003` exists) anchor on `2026.001..HEAD` instead of a newer tag.

.github/workflows/release-template.yml
```diff
@@ -67,18 +75,97 @@ jobs:
...
+          if [[ -z "$PR" ]]; then
+            echo "::error::Could not find a merged develop→main promotion PR. Pass promotion_pr or merge promotion first."
+            exit 1
+          fi
+          BASE="$(gh pr view "$PR" --json baseRefName --jq .baseRefName)"
+          HEAD="$(gh pr view "$PR" --json headRefName --jq .headRefName)"
+          if [[ "$BASE" != "main" || "$HEAD" != "develop" ]]; then
+            echo "::error::PR #${PR} is ${HEAD}→${BASE}, expected develop→main"
+            exit 1
+          fi
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 1e5bff7: the commits API response is filtered with `select(.base.ref == "main" and .head.ref == "develop")` before picking a PR number, and we still fall back to `gh pr list --state merged --base main --head develop` if nothing matches. Explicit `promotion_pr` input is also validated with the same develop→main check.

.github/workflows/release-template.yml
```diff
@@ -67,18 +75,111 @@ jobs:
...
+            echo "No prior CalVer tag — first template release"
+            git-cliff \
+              --config cliff.toml \
+              --tag-pattern '^[0-9]{4}\.' \
+              --tag "$NEW_TAG" \
+              --unreleased \
+              --output cliff-changelog.md
+          fi
+
+          if ! grep -q "## ${NEW_TAG}" cliff-changelog.md; then
```
mei-artificerinnovations: **Bug (low probability but real):** `grep -q "## ${NEW_TAG}"` treats the pattern as a regex, and `.` in a CalVer tag like `2026.003` matches any character. `## 2026X003` would pass this check silently.

Fix: use the `-F` (fixed-string) flag:
```bash
if ! grep -qF "## ${NEW_TAG}" cliff-changelog.md; then
```
One-character change, no other adjustment needed.

scripts/extract-calver-release-notes.sh
```diff
@@ -0,0 +1,56 @@
...
+# Content after the heading line
+after_heading = body[idx + len(heading) :]
+line_end = after_heading.find("\n")
+content = after_heading[line_end + 1 :] if line_end != -1 else ""
+
+# Stop at maintainer separator (--- on its own line)
+lines = content.splitlines()
+published: list[str] = []
+for line in lines:
+    if line.strip() == "---":
```
mei-artificerinnovations, line 43: **Informational:** `line.strip() == "---"` will terminate extraction if `---` appears inside a fenced code block in the CalVer section. Low risk today since the template does not put code blocks in the CalVer prose, but worth noting if the format ever grows.

Not blocking this PR — just something to be aware of if the extraction ever cuts short unexpectedly.

.github/workflows/release-template.yml
```diff
@@ -67,18 +75,111 @@ jobs:
...
+          PR="${{ inputs.promotion_pr }}"
+          if [[ -n "$PR" ]] && ! is_promotion_pr "$PR"; then
+            echo "::error::PR #${PR} is not a develop→main promotion PR"
+            exit 1
+          fi
+
+          if [[ -z "$PR" ]]; then
+            PR="$(gh api "repos/${GITHUB_REPOSITORY}/commits/${GITHUB_SHA}/pulls" \
+              --jq '[.[] | select(.base.ref == "main" and .head.ref == "develop") | .number][0] // empty' 2>/dev/null || true)"
+            if [[ -n "$PR" ]] && ! is_promotion_pr "$PR"; then
```
mei-artificerinnovations: **Nit:** The `gh api commits/{sha}/pulls` query already filters `base.ref == "main" and head.ref == "develop"`, so the `is_promotion_pr` re-validation here will always pass. Harmless double-check — just noting it is redundant if you are counting API calls.

.github/workflows/release-template.yml
```diff
@@ -67,18 +75,111 @@ jobs:
...
+
+          PR="${{ inputs.promotion_pr }}"
+          if [[ -n "$PR" ]] && ! is_promotion_pr "$PR"; then
+            echo "::error::PR #${PR} is not a develop→main promotion PR"
+            exit 1
+          fi
+
+          if [[ -z "$PR" ]]; then
+            PR="$(gh api "repos/${GITHUB_REPOSITORY}/commits/${GITHUB_SHA}/pulls" \
+              --jq '[.[] | select(.base.ref == "main" and .head.ref == "develop") | .number][0] // empty' 2>/dev/null || true)"
```
sarah-artificerinnovations, line 105: Two separate `gh pr view` calls to get `baseRefName` and `headRefName` can be one:
```bash
read base head < <(gh pr view "$num" --json baseRefName,headRefName --jq '[.baseRefName,.headRefName]|@tsv')
```
Minor — the function runs at most 3 times per workflow. No change required.

mei-artificerinnovations (commented):
Clean, well-thought-out PR. The three-layer design (template → extraction script → workflow assembly) is easy to follow and the escape hatches (`skip_pr_body`, `promotion_pr` override) handle the bootstrapping and backfill cases well.

One genuine bug (`grep -q` vs `grep -qF`) flagged inline — easy one-character fix. The rest are informational notes, not blockers.

Approving after the grep flag fix.

.github/workflows/release-template.yml
```diff
@@ -67,18 +75,111 @@ jobs:
...
+          fi
+
+          if [[ -z "$PR" ]] || ! is_promotion_pr "$PR"; then
+            echo "::error::Could not find a merged develop→main promotion PR. Pass promotion_pr or merge promotion first."
+            exit 1
+          fi
+
+          echo "number=${PR}" >> "$GITHUB_OUTPUT"
+          echo "Using promotion PR #${PR} (develop→main)"
+
```
sarah-artificerinnovations, line 119: The `awk` string comparison is correct here **only** because the `grep -E '^[0-9]{4}\.[0-9]{3}$'` filter above guarantees a fixed 3-digit suffix. Consider a brief inline comment so a future editor doesn't assume numeric comparison or remove the filter:
```bash
# String compare is safe: grep above guarantees YYYY.NNN fixed-width format
```

scripts/extract-calver-release-notes.sh
sarah-artificerinnovations, line 43: `line.strip() == '---'` will miss alternate Markdown thematic break forms (`- - -`, `* * *`, `***`). Since the PR body is always produced by the Cursor skill or the template — neither of which uses those forms — this is fine in practice. No change needed, just noting the assumption.

sarah-artificerinnovations (approved):
Review: PR #253 — CalVer release automation

**Verdict: APPROVE**

Well-designed end-to-end automation for the develop→main promotion + CalVer release flow. The three-stage PR resolution, tag anchoring fix, and skip_pr_body escape hatch are all thoughtful. A few nits inline; none are blockers.

### extract-calver-release-notes.sh
Embedding Python in a bash heredoc is a reasonable choice for parsing multiline Markdown — avoids external deps and works on any GH Actions runner. Error messages are clear and actionable.

### Tag anchoring
Replacing `--latest` with an explicit `PREV_TAG..HEAD` range is the right fix. The `grep -E '^[0-9]{4}\.[0-9]{3}$'` filter is load-bearing for the awk string comparison — without the 3-digit guarantee the comparison would be wrong. The `|| true` on the PREV_TAG assignment correctly handles the first-release case.

### Promotion PR resolution
Three-stage fallback (manual input → commit-associated PR → last merged develop→main) is robust. The `is_promotion_pr` guard prevents the fallback from accidentally latching onto a non-promotion PR.

### Cursor skill
The "Hard rules" table at the top is the right information architecture — a maintainer running this in a hurry needs the constraints up front before the procedure. The quality checklist at the bottom is a good self-verification gate.

### PR template split (published / maintainer)
Clean and correct. The `---` separator contract is enforced in both the extractor script and documented in the template itself. The default `pull_request_template.md` restoration is the right call — GitHub only auto-loads the root-level file, not `PULL_REQUEST_TEMPLATE/default.md`.

mei-artificerinnovations (approved):
Approved. Fix `grep -q` → `grep -qF` on the changelog header check (line 157) before merging — one-character change. Everything else is informational.

scripts/extract-calver-release-notes.sh
rahul-artificerinnovations, line 43: The `line.strip() == "---"` guard is correct given the template convention where `---` is reserved as the published/maintainer separator. One edge case: a maintainer who writes a bare `---` thematic break inside their CalVer prose (e.g. in Highlights) will silently truncate extraction at that point rather than failing loudly. Worth adding to the SKILL.md quality checklist or a doc-comment here: _avoid bare `---` inside the CalVer release notes section_.

.github/workflows/release-template.yml
```diff
@@ -67,18 +75,111 @@ jobs:
...
+            if [[ -n "$PR" ]] && ! is_promotion_pr "$PR"; then
+              PR=""
+            fi
+          fi
+
+          if [[ -z "$PR" ]]; then
+            PR="$(gh pr list --state merged --base main --head develop --limit 1 --json number --jq '.[0].number // empty')"
+          fi
+
+          if [[ -z "$PR" ]] || ! is_promotion_pr "$PR"; then
```
rahul-artificerinnovations: The `! is_promotion_pr "$PR"` re-validation here is redundant for the path where `$PR` just came from `gh pr list --base main --head develop` (line 111), which already guarantees the correct base/head. It also doesn't protect against the empty-string case — `is_promotion_pr ""` would error before returning false, but `[[ -z "$PR" ]]` already handles that. Very minor: could simplify the last guard to just `if [[ -z "$PR" ]]; then`.

.github/workflows/release-template.yml
```diff
@@ -67,18 +75,111 @@ jobs:
...
+          }
+
+          PR="${{ inputs.promotion_pr }}"
+          if [[ -n "$PR" ]] && ! is_promotion_pr "$PR"; then
+            echo "::error::PR #${PR} is not a develop→main promotion PR"
+            exit 1
+          fi
+
+          if [[ -z "$PR" ]]; then
+            PR="$(gh api "repos/${GITHUB_REPOSITORY}/commits/${GITHUB_SHA}/pulls" \
```
rahul-artificerinnovations, line 104: In a `workflow_dispatch` event, `GITHUB_SHA` resolves to the branch HEAD at trigger time — not the promotion merge commit. If Release Template runs days after the merge with hotfixes on top of `main`, this SHA maps to a newer commit whose `/commits/{sha}/pulls` response won't include the promotion PR. The `gh pr list` fallback on line 111 handles this correctly, but a short comment here would prevent confusion: e.g. `# GITHUB_SHA is HEAD at trigger time; for deferred runs this SHA may not map to the promotion merge — fallback below handles it`.

rahul-artificerinnovations (approved):
Solid automation. The three-layer design (PR template → extract script → workflow) is clean, the Copilot issues are all properly fixed, and the escape hatches (`skip_pr_body`, explicit `promotion_pr`) cover the edge cases. A few inline observations — none are blockers.

CONTRIBUTING.md
```diff
@@ -78,6 +78,13 @@ Beaker Stack uses a two-stage branch flow:
 
 Feature branches should always target `develop`, never `main` directly. The `develop` → `main` promotion is done deliberately when the integrated state is ready for production.
 
+### Promoting `develop` → `main`
+
+1. Open a PR: `develop` → `main` using the [**Promote develop to main** template](.github/PULL_REQUEST_TEMPLATE/promote-develop-to-main.md) (GitHub UI: choose template when opening the PR, or `gh pr create --base main --head develop --template promote-develop-to-main.md`). **Agents:** follow [`.cursor/skills/promote-develop-to-main/SKILL.md`](.cursor/skills/promote-develop-to-main/SKILL.md) to draft CalVer release notes and open the PR with `gh`.
```
diego-artificerinnovations: test comment

diego-artificerinnovations (commented):
test review

diego-artificerinnovations (commented):
test

scripts/extract-calver-release-notes.sh
diego-artificerinnovations, line 43: test

CONTRIBUTING.md
diego-artificerinnovations: blocking: `gh pr create` has no `--template <filename>` flag — this command will error for any maintainer who tries it.

The SKILL.md correctly uses `--body-file`. CONTRIBUTING.md should match:

```bash
gh pr create --base main --head develop \
  --title "release: promote develop to main (...)" \
  --body-file /tmp/promotion-pr-body.md
```

Alternatively, drop the `gh` snippet from CONTRIBUTING.md and let the skill be the canonical CLI path — most maintainers will use the GitHub UI template picker anyway.

scripts/extract-calver-release-notes.sh
diego-artificerinnovations, line 43: nit: a bare `---` line anywhere inside the CalVer section terminates extraction early — the parser treats the first `---` it sees as the maintainer separator regardless of position. Probably fine given the template structure, but a short comment here and a note in the PR template would help authors know not to use `---` as a thematic break within the CalVer prose:

```python
# Any '---' line ends extraction — do not use '---' as a divider inside CalVer notes.
if line.strip() == "---":
    break
```

diego-artificerinnovations (changes requested):
Really clean architecture overall. The CalVer section / `---` split as the published/maintainer boundary is elegant, the SHA-verified git-cliff install is good hygiene, and the SKILL.md hard-rules table is exactly the right level of prescription for agents. One blocking bug, two non-blocking observations.

**Non-blocking (workflow):** In the `Compute next CalVer tag` step, `git tag -l "${YEAR}.*"` uses a glob that also matches semver-shaped tags starting with the year (e.g. `2026.1.0`). `sort -n` on `1.0` parses as `1`, which could skew the sequence counter. The PREV_TAG step guards correctly with `grep -E '^[0-9]{4}\.[0-9]{3}$'`. Applying the same filter here keeps both steps consistent:

```bash
LAST_N=$(git tag -l | grep -E ^${YEAR}\.[0-9]{3} | sed s/${YEAR}\.// | sort -n | tail -1)
```

Low-risk given current tag conventions, but worth closing the gap.

.github/workflows/release-template.yml
ZappoMan: Fixed in 869cf55: `grep -qF` so the CalVer header is matched literally (dots are not regex wildcards).

scripts/extract-calver-release-notes.sh
ZappoMan: Agreed — added a comment in the extractor and a note in the promotion PR template + skill checklist (869cf55).

.github/workflows/release-template.yml
ZappoMan: Removed the redundant `is_promotion_pr` re-check after the jq filter in 869cf55. Kept validation for explicit `promotion_pr` input.

.github/workflows/release-template.yml
ZappoMan: Done in 869cf55 — single `gh pr view` with tsv output in `is_promotion_pr`.

.github/workflows/release-template.yml
ZappoMan: Added the inline awk comment in 869cf55.

scripts/extract-calver-release-notes.sh
ZappoMan: Acknowledged — we rely on the template/skill convention (`---` on its own line only). Documented in template + extractor (869cf55).

ZappoMan: Added to the skill quality checklist and promotion template in 869cf55.

.github/workflows/release-template.yml
ZappoMan: Simplified the final guard to `[[ -z "$PR" ]]` in 869cf55; explicit `promotion_pr` is still validated up front.

.github/workflows/release-template.yml
ZappoMan: Added the workflow_dispatch / GITHUB_SHA comment in 869cf55. The `gh pr list` fallback remains the reliable path for deferred runs.

CONTRIBUTING.md
ZappoMan: Fixed in 869cf55: removed invalid `gh pr create --template`. CONTRIBUTING now points to the UI template picker / `?template=` compare URL; agents use `--body-file` per the skill.

scripts/extract-calver-release-notes.sh
ZappoMan: Done in 869cf55 — comment in extractor, note in promotion template, and skill checklist item.

ZappoMan: Also fixed in 869cf55: `Compute next CalVer tag` now filters `^YEAR.[0-9]{3}$` so semver-shaped tags like `2026.1.0` cannot skew the sequence counter.

mei-artificerinnovations (approved):
All three review items addressed:

- `grep -qF` ✓ — fixed-string flag applied, `.` in CalVer tags no longer a regex hazard
- `is_promotion_pr` refactored to a single `gh pr view` call with both fields; redundant double-check removed ✓
- Extraction script documents the `---` separator contract inline ✓

Bonus: `LAST_N` tag filter tightened to `grep -E "^${YEAR}\.[0-9]{3}$"` — correctly excludes semver-shaped tags like `2026.1.0` that the old `git tag -l "${YEAR}.*"` glob would have included.

Approved. Good to merge.

diego-artificerinnovations (approved):
All three concerns addressed — approving.

- **CONTRIBUTING.md**: invalid `--template` flag removed; example now uses `--body-file` / UI picker. Blocking issue resolved.
- **`---` edge case**: comment added to extractor and a note in the PR template warning authors not to use `---` as a thematic break. Good.
- **Tag glob filter**: `git tag -l | grep -E "^${YEAR}\.[0-9]{3}$"` now applied consistently in the sequence-counter step. Matches the `PREV_TAG` guard.

Ships clean.

PR #256, opened 2026-05-17
feat(setup): Stripe phase and safer fork rename defaults

ZappoMan:
### Stripe billing phase (`setup:full`)

- Adds a dedicated **`stripe`** phase (after `google`, before `write` / `github`) so adopters collect billing keys with context instead of hitting `STAGING_STRIPE_SECRET_KEY` cold at GitHub sync.
- **Preflight briefing** — test vs live mode, one webhook per Supabase project, and what stays manual (`billing:sync-stripe`, `supabase secrets set`, Stripe CLI).
- **Per-tier prompts** (preview → staging → production): prints webhook URL when `*_SUPABASE_URL` exists, then masked `sk_*` and `whsec_*` keys.
- **`--skip-stripe`** — skips the phase and omits Stripe secrets from required GitHub sync.

### Fork-safe rename defaults (`rename-project` + identity phase)

- **Default rename mode is now `--preserve-upstream`**: display name, slugs, and bundle IDs change; **`@beakerstack/*` npm scope**, `docs/UPGRADING.md`, `docs/VERSIONING.md`, and related upstream merge docs are **left unchanged** so forks can keep merging from Artificer-Innovations/BeakerStack and consuming published packages.
- **`--full-rebrand`** restores legacy behavior (rename scope + all docs).
- **`setup:full` identity phase** asks: “Fully rebrand npm scope and upgrade docs? (y/N)” — default **no** passes `--preserve-upstream`.
- Expanded tests and [docs/renaming.md](docs/renaming.md) table explaining default vs full rebrand.

## Docs

- [docs/setup-prep-checklist.md](docs/setup-prep-checklist.md) — `stripe` phase
- [docs/stripe-billing-setup.md](docs/stripe-billing-setup.md) — Before the Stripe wizard phase
- [docs/renaming.md](docs/renaming.md) — preserve upstream vs full rebrand

## Test plan

**Stripe**

- [ ] `npm run setup:full -- --dry-run --from=stripe` — briefing + dry-run collection message
- [ ] `npm run setup:full -- --from=stripe` — staging keys land in `.env.cloud.generated.local` after `write`
- [ ] `npm run setup:full -- --skip-stripe --from=github` — Stripe keys not in missing-required list
- [ ] `node --test scripts/__tests__/setup-stripe.test.mjs scripts/__tests__/setup-manifest-github-sync.test.mjs`

**Rename**

- [ ] `npm run rename -- --from "Beaker Stack" --to "Acme App" --dry-run` — skips `@beakerstack`, UPGRADING, VERSIONING
- [ ] Same with `--full-rebrand` — scope and upgrade docs renamed
- [ ] `node --test scripts/__tests__/rename-project.test.mjs`
- [ ] `setup:full` identity: answer **N** to full rebrand → log shows `preserve upstream`

### DIFF
--- a/scripts/__tests__/setup-stripe.test.mjs
+++ b/scripts/__tests__/setup-stripe.test.mjs
@@ -1,0 +1,33 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+
+import {
+  supabaseStripeWebhookUrl,
+  tierStripeKeysPresent,
+  STRIPE_SETUP_TIERS,
+} from '../lib/setup-stripe.mjs';
+
+test('supabaseStripeWebhookUrl', () => {
+  assert.equal(
+    supabaseStripeWebhookUrl('https://abcd1234.supabase.co'),
+    'https://abcd1234.supabase.co/functions/v1/stripe-webhook'
+  );
+  assert.equal(supabaseStripeWebhookUrl('http://localhost:54321'), '');
+});
+
+test('tierStripeKeysPresent', () => {
+  const tier = STRIPE_SETUP_TIERS.find(t => t.id === 'staging');
+  assert.ok(tier);
+  assert.ok(
+    !tierStripeKeysPresent({ STAGING_STRIPE_SECRET_KEY: 'sk_test_x' }, tier)
+  );
+  assert.ok(
+    tierStripeKeysPresent(
+      {
+        STAGING_STRIPE_SECRET_KEY: 'sk_test_x',
+        STAGING_STRIPE_WEBHOOK_SECRET: 'whsec_x',
+      },
+      tier
+    )
+  );
+});
--- a/scripts/__tests__/setup-stripe.test.mjs
+++ b/scripts/__tests__/setup-stripe.test.mjs
@@ -2,6 +2,9 @@ import assert from 'node:assert/strict';
 import test from 'node:test';
 
 import {
+  isStripeGithubSecretDef,
+  resolveSupabaseUrlForStripeTier,
+  setupStripeKeysDeferred,
   supabaseStripeWebhookUrl,
   tierStripeKeysPresent,
   STRIPE_SETUP_TIERS,
@@ -13,6 +16,13 @@ test('supabaseStripeWebhookUrl', () => {
     'https://abcd1234.supabase.co/functions/v1/stripe-webhook'
   );
   assert.equal(supabaseStripeWebhookUrl('http://localhost:54321'), '');
+});
+
+test('supabaseStripeWebhookUrl strips trailing slash', () => {
+  assert.equal(
+    supabaseStripeWebhookUrl('https://abcd1234.supabase.co/'),
+    'https://abcd1234.supabase.co/functions/v1/stripe-webhook'
+  );
 });
 
 test('tierStripeKeysPresent', () => {
@@ -30,4 +40,39 @@ test('tierStripeKeysPresent', () => {
       tier
     )
   );
+});
+
+test('isStripeGithubSecretDef matches STRIPE keys', () => {
+  assert.ok(isStripeGithubSecretDef({ name: 'STAGING_STRIPE_SECRET_KEY' }));
+  assert.ok(
+    isStripeGithubSecretDef({ name: 'PRODUCTION_STRIPE_WEBHOOK_SECRET' })
+  );
+  assert.ok(!isStripeGithubSecretDef({ name: 'AWS_ACCESS_KEY_ID' }));
+  assert.ok(!isStripeGithubSecretDef({}));
+});
+
+test('setupStripeKeysDeferred', () => {
+  assert.ok(setupStripeKeysDeferred({ SETUP_STRIPE_SKIPPED: 'true' }));
+  assert.ok(!setupStripeKeysDeferred({ SETUP_STRIPE_SKIPPED: 'false' }));
+  assert.ok(!setupStripeKeysDeferred({}));
+});
+
+test('resolveSupabaseUrlForStripeTier uses first non-empty key', () => {
+  const tier = STRIPE_SETUP_TIERS.find(t => t.id === 'preview');
+  assert.ok(tier);
+  assert.equal(
+    resolveSupabaseUrlForStripeTier(
+      { PREVIEW_SUPABASE_URL: 'https://a.supabase.co' },
+      tier
+    ),
+    'https://a.supabase.co'
+  );
+  assert.equal(
+    resolveSupabaseUrlForStripeTier(
+      { PR_TESTING_SUPABASE_URL: 'https://b.supabase.co' },
+      tier
+    ),
+    'https://b.supabase.co'
+  );
+  assert.equal(resolveSupabaseUrlForStripeTier({}, tier), '');
 });
--- a/scripts/__tests__/setup-stripe.test.mjs
+++ b/scripts/__tests__/setup-stripe.test.mjs
@@ -15,13 +15,24 @@ test('supabaseStripeWebhookUrl', () => {
     supabaseStripeWebhookUrl('https://abcd1234.supabase.co'),
     'https://abcd1234.supabase.co/functions/v1/stripe-webhook'
   );
-  assert.equal(supabaseStripeWebhookUrl('http://localhost:54321'), '');
+  assert.equal(supabaseStripeWebhookUrl('not-a-url'), '');
 });
 
 test('supabaseStripeWebhookUrl strips trailing slash', () => {
   assert.equal(
     supabaseStripeWebhookUrl('https://abcd1234.supabase.co/'),
     'https://abcd1234.supabase.co/functions/v1/stripe-webhook'
+  );
+});
+
+test('supabaseStripeWebhookUrl supports local and custom base URLs', () => {
+  assert.equal(
+    supabaseStripeWebhookUrl('http://127.0.0.1:54321'),
+    'http://127.0.0.1:54321/functions/v1/stripe-webhook'
+  );
+  assert.equal(
+    supabaseStripeWebhookUrl('https://api.example.com/custom/path/'),
+    'https://api.example.com/custom/path/functions/v1/stripe-webhook'
   );
 });
 
--- a/scripts/lib/setup-manual-instructions.mjs
+++ b/scripts/lib/setup-manual-instructions.mjs
@@ -511,7 +511,7 @@ export function printStripePhaseReadinessBriefing(ctx) {
   logInfo('');
   logInfo('4) Complete supabase phase first');
   logInfo(
-    '   • The wizard prints each webhook URL when STAGING_/PREVIEW_/PRODUCTION_SUPABASE_URL is already in acc.'
+    '   • The wizard prints each webhook URL when STAGING_/PREVIEW_/PRODUCTION_SUPABASE_URL is already set from the supabase phase.'
   );
   logInfo('');
   logInfo('5) Not run in this phase (do later)');
@@ -594,7 +594,7 @@ export function printGithubPhaseReadinessBriefing(ctx) {
     '   • Skipped **AWS** or **Supabase**? Add secrets manually later or re-run `--from=aws` / `--from=supabase`.'
   );
   logInfo(
-    '   • Skipped **stripe**? Re-run `--from=stripe` or enter keys here; use `--skip-stripe` to omit Stripe from CI sync.'
+    '   • Skipped **stripe** (menu **N** or `--skip-stripe`)? Stripe keys are omitted from github prompts; re-run `--from=stripe` without `--skip-stripe` to collect them.'
   );
   logInfo('');
   logInfo('5) During this phase');
--- a/scripts/lib/setup-manual-instructions.mjs
+++ b/scripts/lib/setup-manual-instructions.mjs
@@ -250,6 +250,13 @@ const PHASE_INTROS = {
       'If you choose Yes, a prerequisites checklist prints next — optional; press Enter to skip.',
     ],
   },
+  stripe: {
+    title: 'Stripe billing (CI keys)',
+    body: [
+      'Collects PREVIEW_*, STAGING_*, and PRODUCTION_* Stripe secret + webhook signing keys for GitHub Actions.',
+      'If you choose Yes, a prerequisites checklist prints next — Stripe Dashboard test/live mode, webhook URLs per Supabase project.',
+    ],
+  },
   write: {
     title: 'Write env files',
     body: [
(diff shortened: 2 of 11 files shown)


sarah-artificerinnovations (approved):
## Review: Stripe wizard phase

**Overall: LGTM — approve with two minor nits (neither blocking).**

The Stripe phase slots cleanly into the existing wizard pattern: phase intro → readiness briefing → per-tier collect loop → `SETUP_STRIPE_SKIPPED_ENV` bookkeeping. The `--skip-stripe` flag correctly propagates into both the phase runner and `listMissingRequiredGithubForCi`, so skipping at the wizard won't nag again at GitHub sync. Docs updated in both the prep checklist and the stripe guide. Tests cover the pure helpers consistently with the rest of the suite.

Two nits below.

sarah-artificerinnovations:
Two nits (neither blocking):

**1. Duplicate JSDoc above `phaseStripe`** (`scripts/setup-full.mjs`) — there are two consecutive JSDoc blocks before the function. The first (3-param version, missing `promptInput`) is a leftover draft that tooling ignores. The second is the correct one. Just delete the first block.

**2. Stripe missing-keys hint at GitHub sync** — the hint in `collectMissingGithubCiEnvIntoAcc` only suggests `--from=stripe` for guided re-collection, but doesn't mention `--skip-stripe` for adopters who have no billing integration. Worth adding "or use `--skip-stripe` to omit Stripe from CI requirements" so non-billing repos don't feel stuck.

scripts/setup-full.mjs
```diff
@@ -2191,6 +2201,41 @@ async function phaseExpo(flags, rl, acc, promptInput) {
   await promptExpoTokenForGithub(rl, promptInput, acc, flags);
 }
 
+/**
+ * @param {CliFlags} flags
+ * @param {import('node:readline/promises').ReadLine} rl
+ * @param {Record<string, string>} acc
+ */
```
diego-artificerinnovations: nit: there are two JSDoc blocks before `phaseStripe` — lines 2204–2208 are an orphaned 3-param draft (no `promptInput`) left over from before the signature was finalised. The correct block is lines 2209–2214. Remove the first one:

```js
// remove this block:
/**
 * @param {CliFlags} flags
 * @param {import('node:readline/promises').ReadLine} rl
 * @param {Record<string, string>} acc
 */
```

scripts/__tests__/setup-stripe.test.mjs
```diff
@@ -0,0 +1,33 @@
...
+  assert.ok(
+    tierStripeKeysPresent(
+      {
+        STAGING_STRIPE_SECRET_KEY: 'sk_test_x',
+        STAGING_STRIPE_WEBHOOK_SECRET: 'whsec_x',
+      },
+      tier
+    )
+  );
+});
```
diego-artificerinnovations, line 54: nit: the two existing tests cover the happy paths well, but a few cheap additions would close the gap on the functions that the manifest filtering depends on:

```js
test('isStripeGithubSecretDef matches STRIPE keys', () => {
  assert.ok(isStripeGithubSecretDef({ name: 'STAGING_STRIPE_SECRET_KEY' }));
  assert.ok(isStripeGithubSecretDef({ name: 'PRODUCTION_STRIPE_WEBHOOK_SECRET' }));
  assert.ok(!isStripeGithubSecretDef({ name: 'AWS_ACCESS_KEY_ID' }));
  assert.ok(!isStripeGithubSecretDef({}));
});

test('setupStripeKeysDeferred', () => {
  assert.ok(setupStripeKeysDeferred({ SETUP_STRIPE_SKIPPED: 'true' }));
  assert.ok(!setupStripeKeysDeferred({ SETUP_STRIPE_SKIPPED: 'false' }));
  assert.ok(!setupStripeKeysDeferred({}));
});

test('resolveSupabaseUrlForStripeTier uses first non-empty key', () => {
  const tier = STRIPE_SETUP_TIERS.find(t => t.id === 'preview');
  assert.equal(resolveSupabaseUrlForStripeTier({ PREVIEW_SUPABASE_URL: 'https://a.supabase.co' }, tier), 'https://a.supabase.co');
  assert.equal(resolveSupabaseUrlForStripeTier({ PR_TESTING_SUPABASE_URL: 'https://b.supabase.co' }, tier), 'https://b.supabase.co');
  assert.equal(resolveSupabaseUrlForStripeTier({}, tier), '');
});

test('supabaseStripeWebhookUrl strips trailing slash', () => {
  assert.equal(
    supabaseStripeWebhookUrl('https://abcd1234.supabase.co/'),
    'https://abcd1234.supabase.co/functions/v1/stripe-webhook'
  );
});
```

`isStripeGithubSecretDef` and `setupStripeKeysDeferred` are worth testing specifically because they gate the manifest filter — if either regresses, `--skip-stripe` silently stops working.

diego-artificerinnovations (approved):
Clean addition — the tier structure (preview/staging/production with separate modes and webhook URL derivation), the `SETUP_STRIPE_SKIPPED` env flag threading through the manifest filter, and the github-phase hint are all the right design choices. The briefing is concise and actionable.

Two nits inline, none blocking.

scripts/setup-full.mjs
rahul-artificerinnovations: There are two JSDoc blocks on `phaseStripe` — the first one (lines 2204–2208) is stale and wrong: it lists only three params (`flags`, `rl`, `acc`) and omits `promptInput`. The second block is correct. The first should be deleted.

scripts/setup-full.mjs
```diff
@@ -2191,6 +2201,41 @@ async function phaseExpo(flags, rl, acc, promptInput) {
...
+/**
+ * @param {CliFlags} flags
+ * @param {import('node:readline/promises').ReadLine} rl
+ * @param {import('stream').Readable & { isTTY?: boolean; setRawMode?: (flag: boolean) => void }} promptInput
+ * @param {Record<string, string>} acc
+ */
+async function phaseStripe(flags, rl, promptInput, acc) {
+  if (flags.skipStripe) {
```
rahul-artificerinnovations: This `if (flags.skipStripe)` guard is dead code. The main loop's early-continue at line ~2762 (`if (phase === 'stripe' && flags.skipStripe) { ... continue; }`) handles `--skip-stripe` before `phaseStripe` is ever called. The branch here can never execute; remove it to avoid misleading readers about the actual control flow.

scripts/lib/setup-manual-instructions.mjs
```diff
@@ -464,6 +471,64 @@ export function printGooglePhaseReadinessBriefing(ctx) {
...
+  logInfo(
+    '   • Webhook signing secret — Dashboard → Developers → Webhooks → your endpoint → Reveal'
+  );
+  logInfo(
+    '   • One webhook URL per Supabase project: https://<PROJECT_REF>.supabase.co/functions/v1/stripe-webhook'
+  );
+  logInfo('');
+  logInfo('4) Complete supabase phase first');
+  logInfo(
+    '   • The wizard prints each webhook URL when STAGING_/PREVIEW_/PRODUCTION_SUPABASE_URL is already in acc.'
```
rahul-artificerinnovations: "already in acc" leaks the internal accumulator name into user-facing terminal output. Same pattern appears one more time in `setup-stripe.mjs` (`No Supabase URL in acc yet`). Suggest "already set from the supabase phase" / "not yet set from the supabase phase" for both.

rahul-artificerinnovations (approved):
Clean addition. Pulling Stripe key collection out of the github-phase catch-all and giving it its own phase is exactly the right call — the briefing, per-tier prompts, and `--skip-stripe` deferral path are all well-designed. `setup-stripe.mjs` is pleasantly self-contained. Two minor code issues and one user-facing wording nit below, none are blockers.

scripts/setup-full.mjs
mei-artificerinnovations: There are two consecutive JSDoc blocks here — only the second one (starting with `@param {CliFlags} flags`, `@param {import('node:readline/promises')...`, `@param {import('stream')...`, `@param acc`) is attached to `phaseStripe`. The first block (ending here) is dead and should be removed.

scripts/lib/setup-manual-instructions.mjs
mei-artificerinnovations: "in acc" is an internal implementation detail (the accumulator object) — users won’t know what it means. Suggest: `is already collected from the supabase phase` or similar.

scripts/setup-full.mjs
```diff
@@ -2299,6 +2344,13 @@ async function collectMissingGithubCiEnvIntoAcc(flags, rl, promptInput, acc) {
   let details = listMissingRequiredGithubCiDetails(acc);
   if (!details.length) return 'none';
 
+  const stripeMissing = details.filter(d => isStripeGithubSecretDef(d.def));
```
mei-artificerinnovations, line 2346: This accesses `d.def` from the objects returned by `listMissingRequiredGithubCiDetails`. Can you confirm that function includes `def` in its return objects? If not, `stripeMissing` will always be empty and the hint below will never print. A quick sanity check: `console.log(details[0])` in a dry-run to verify the shape.

mei-artificerinnovations (commented):
Solid addition — the tiered prompts, `--skip-stripe` deferral, and the github-phase hint all hang together well. Three items before this merges.

**Required**
1. Dead JSDoc block on `phaseStripe`
2. User-facing "in acc" wording leaks an internal variable name

**Verify**
3. `d.def` access in `collectMissingGithubCiEnvIntoAcc` — depends on whether `listMissingRequiredGithubCiDetails` includes `def` in its return objects; if not, the github-phase Stripe hint never fires.

scripts/setup-full.mjs
ZappoMan: Fixed in 00a2b84 — removed the stale 3-param JSDoc; only the block with `promptInput` remains.

scripts/__tests__/setup-stripe.test.mjs
ZappoMan: Added the suggested tests (`isStripeGithubSecretDef`, `setupStripeKeysDeferred`, `resolveSupabaseUrlForStripeTier`, trailing-slash webhook URL) in 00a2b84.

scripts/setup-full.mjs
```diff
@@ -2580,6 +2632,11 @@ async function main() {
         logInfo('Skipping identity/rename (--skip-rename).');
         continue;
       }
+      if (phase === 'stripe' && flags.skipStripe) {
+        logInfo('Skipping Stripe phase (--skip-stripe).');
+        acc[SETUP_STRIPE_SKIPPED_ENV] = 'true';
+        continue;
```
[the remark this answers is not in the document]
ZappoMan: Fixed — `acc[SETUP_STRIPE_SKIPPED_ENV] = 'true'` is set right after env files load when `--skip-stripe` is passed, so `--from=write` / `--from=github` still omits Stripe from required github sync.

scripts/setup-full.mjs
```diff
@@ -2191,6 +2201,41 @@ async function phaseExpo(flags, rl, acc, promptInput) {
...
 }
 
+/**
+ * @param {CliFlags} flags
+ * @param {import('node:readline/promises').ReadLine} rl
+ * @param {Record<string, string>} acc
+ */
+/**
+ * @param {CliFlags} flags
+ * @param {import('node:readline/promises').ReadLine} rl
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 00a2b84 (duplicate JSDoc removed).

scripts/lib/setup-manual-instructions.mjs
```diff
@@ -528,6 +593,9 @@ export function printGithubPhaseReadinessBriefing(ctx) {
   logInfo(
     '   • Skipped **AWS** or **Supabase**? Add secrets manually later or re-run `--from=aws` / `--from=supabase`.'
   );
+  logInfo(
+    '   • Skipped **stripe**? Re-run `--from=stripe` or enter keys here; use `--skip-stripe` to omit Stripe from CI sync.'
+  );
```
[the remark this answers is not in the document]
ZappoMan: Updated github briefing: skipping stripe (menu N or `--skip-stripe`) omits Stripe from github prompts; re-run `--from=stripe` without the flag to collect keys.

scripts/setup-full.mjs
ZappoMan: Fixed in 00a2b84 — removed the orphaned JSDoc block.

scripts/setup-full.mjs
ZappoMan: Fixed — removed the unreachable `if (flags.skipStripe)` inside `phaseStripe`; the loop early-continue handles it, and `--skip-stripe` now sets the defer flag before resume.

scripts/lib/setup-manual-instructions.mjs
ZappoMan: Replaced "in acc" with "set from the supabase phase" in the stripe briefing and setup-stripe warnings (00a2b84).

scripts/setup-full.mjs
ZappoMan: Fixed in 00a2b84 — duplicate JSDoc removed.

scripts/lib/setup-manual-instructions.mjs
ZappoMan: Same fix as Rahul's thread — user-facing copy no longer says "in acc" (00a2b84).

scripts/setup-full.mjs
ZappoMan: Confirmed: `listMissingRequiredGithubCiDetails` returns `def` on each entry (see `setup-manifest.mjs` and new test `listMissingRequiredGithubCiDetails includes def for github-phase filtering`). The github hint filter is correct.

mei-artificerinnovations (approved):
All three items addressed, and the rename changes are solid.

- Dead JSDoc removed; `phaseStripe` has one clean block now.
- "in acc" → "already set from the supabase phase" — clear.
- `d.def` kept; Brad confirmed the shape.

The `(?<!@)\b…\b` lookbehind in `replaceFlatLowerPreservingScope` is the right call — preserves `@beakerstack/billing` (lookbehind catches the `@`), leaves `beakerstackBillingConfig` alone (trailing `B` breaks the word boundary), and still rewrites `com.anonymous.beakerstack` and bare `productId` values. Tests cover all four cases. `resolvePreserveUpstream` precedence and the `process.argv.includes` guard for the conflict warning are both correct. Approved.

diego-artificerinnovations (approved):
All Stripe nits resolved — re-approving.

**Stripe fixes (00a2b84):** duplicate JSDoc gone, dead skip-branch gone (and `SETUP_STRIPE_SKIPPED` now set early in `main()` before the phase loop, which is better — covers `--from=github` resumes without having run the stripe phase at all). All four suggested tests added plus a bonus manifest-details test. Clean.

**Rename feature (07c113634):** The preserve-upstream default is the right call for forks. The mechanism is sound: annotating the `flatlower` pair with `preserveScope` routes it through `replaceFlatLowerPreservingScope`, which uses `(?<!@)\b…\b` — correctly sparing `@beakerstack/*` imports and `beakerstackBillingConfig` (no word boundary between `k` and `B` in camelCase), while still renaming standalone occurrences like `com.anonymous.beakerstack`. Tests cover the key cases well.

One observation for a follow-up (not blocking): `--strict` in preserve mode will still flag `@beakerstack/*` imports as remaining occurrences, because `collectRemainingMatches` does a raw pattern match on `"beakerstack"` without scope awareness. A fork keeping the npm scope that runs `--strict` will always get a non-zero exit. Worth either noting in the `--strict` docs ("not recommended with `--preserve-upstream` when `@scope` imports remain") or filtering scope-prefixed lines from the remaining-match scan in preserve mode.

rahul-artificerinnovations (approved):
All three Stripe observations addressed cleanly, and the rename additions are solid.

On the Stripe fixes: stale JSDoc removed, dead `flags.skipStripe` guard removed, and the early `acc[SETUP_STRIPE_SKIPPED_ENV] = 'true'` at the top of `main()` is a bonus improvement — it ensures `--from=github --skip-stripe` resumes correctly without needing to hit the stripe phase at all. Both user-facing "in acc" strings replaced.

On the rename changes: the `--preserve-upstream` default is exactly the right call for a template repo — forks should be able to keep pulling upstream updates and consuming `@beakerstack/*` packages without fighting a renamed scope. The `(?<!@)\b` approach in `replaceFlatLowerPreservingScope` is elegant: `@beakerstack/billing` and `beakerstackBillingConfig` (no word boundary between `k` and `B`) are both preserved, while `com.anonymous.beakerstack` and bare quoted values are renamed correctly. Verified the regex against all the interesting cases. Test coverage is thorough. Good to merge.

sarah-artificerinnovations:
Both nits from my first review are addressed. Happy with the state of this PR.

The fixes from the full round of reviews are solid:

- **Duplicate JSDoc removed** — single correct block above `phaseStripe`. ✓
- **`--skip-stripe` + `--from=write` resume bug fixed** — `SETUP_STRIPE_SKIPPED_ENV` is now set early in `main()` before the phase loop, so `--skip-stripe --from=github` correctly omits Stripe from the required list even without ever running the stripe phase. Good catch from Copilot. ✓
- **Dead `if (flags.skipStripe)` inside `phaseStripe` removed.** ✓
- **Additional tests** for `isStripeGithubSecretDef`, `setupStripeKeysDeferred`, `resolveSupabaseUrlForStripeTier`, trailing-slash URL. ✓

The new `--preserve-upstream` rename default is a good addition. The `(?<!@)\bbeakerstack\b` pattern correctly keeps `@beakerstack/pkg` imports intact while still rewriting `com.anonymous.beakerstack` and bare `'beakerstack'` values — tests verify all four cases. Making this the default (with `--full-rebrand` for legacy behavior) means forks can safely rename display names without accidentally breaking their npm workspace imports.

scripts/lib/setup-stripe.mjs
```diff
@@ -0,0 +1,174 @@
...
+ * @param {string} supabaseUrl
+ * @returns {string}
+ */
+export function supabaseStripeWebhookUrl(supabaseUrl) {
+  const raw = String(supabaseUrl || '')
+    .trim()
+    .replace(/\/+$/, '');
+  const m = raw.match(/^https:\/\/([a-z0-9-]+)\.supabase\.co/i);
+  if (!m) return '';
+  return `https://${m[1]}.supabase.co/functions/v1/stripe-webhook`;
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 933f224: `supabaseStripeWebhookUrl` now uses `URL` parsing — standard `*.supabase.co` hosts still get the canonical webhook URL; any other valid `http(s)` base (local `127.0.0.1:54321`, custom domains, paths) appends `/functions/v1/stripe-webhook` without duplicating if already present. Tests cover local + custom path bases.

scripts/lib/setup-stripe.mjs
```diff
@@ -0,0 +1,174 @@
...
+      `  Secret key → ${tier.secretKey}  |  Signing secret → ${tier.webhookSecretKey}`
+    );
+    logInfo(
+      '  (Enter to skip this tier — github phase will ask again if still empty.)'
+    );
+    logInfo('');
+
+    const skipTier = (
+      await question(`Collect Stripe keys for ${tier.label} now? (Y/n): `)
+    )
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 933f224: copy now matches `(Y/n)` — “Enter or y = collect this tier; n = skip” (github phase may still prompt if empty).

scripts/rename-project.mjs
```diff
@@ -140,29 +150,128 @@ export function buildReplacementPairs(fromName, toName) {
...
+  preserveUpstream
+) {
+  if (!preserveUpstream) {
+    return replacements;
+  }
+
+  return replacements.map(pair =>
+    pair.description === 'flatlower' ? { ...pair, preserveScope: true } : pair
+  );
+}
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 933f224: `uniquePairs` prefers the `flatlower` entry when two variants share the same `from` (e.g. single-token `Beaker` → `beaker`), so `preserveScope` annotation still applies. Added test `uniquePairs keeps flatlower when camelCase shares the same from token`.

PR #258, opened 2026-05-18
Harden stripe-webhook for multiple BeakerStack apps on one Stripe account

ZappoMan:
## Summary

- Add ingress classification so foreign Stripe events (from another BeakerStack app sharing the same Stripe account) are logged with a redacted payload and return `200` with `ignored: true` without mutating billing tables.
- Enforce `billing_deploy_target` on `customer.subscription.*` and require a local `billing_subscriptions` row by `stripe_subscription_id` before subscription or invoice updates.
- Remove invoice sync via `stripe_customer_id` alone, which could attach another app’s invoices when customers are shared across apps.

## Test plan

- [x] `node --test scripts/__tests__/billing-webhook-guards.test.mjs`
- [ ] Deploy `billing-stripe` then `stripe-webhook` on each Supabase project that shares the Stripe account
- [ ] Trigger billing on one app; confirm the other app’s `billing_webhook_events` shows `ignored: …` with `payload.redacted === true` and no `billing_invoices` / `billing_subscriptions` changes


sarah-artificerinnovations (changes requested):
## Review: Harden stripe-webhook for multi-app Stripe account

**Overall: Request changes — three medium-priority issues before merging; two are easy fixes.**

The design is correct and the motivation is sound — ingress classification running before the DB insert, with a redacted payload for foreign events, is the right approach. The pure-core / Deno-wrapper split makes the logic testable. The removal of `resolveUserIdByStripeCustomer` is a genuine correctness improvement: it closes the cross-app invoice mirroring hole when Stripe customers are shared.

That said, I found issues significant enough to warrant changes before this lands. Detailed below.

---

### [M1] Missing test — owned subscription update happy path

The test suite only has **negative** cases for `customer.subscription.*`. There's no test for:

> A subscription update event that matches deploy target + has an owned row → `{ action: 'process' }`.

Every other event type has a passing-case test. The classifier branches for subscription events include `deployTargetMismatch` + `ownedOk`; without a green-path test, a regression in either branch could silently make the classifier reject all subscription updates.

Missing test to add:

```js
test('classify: owned subscription update without mismatch processes', async () => {
  const stub = stubSupabase(
    new Map([['sub_1', { user_id: 'u', product_id: 'app_a' }]])
  );
  const decision = await classifyStripeEventCore(
    {
      type: 'customer.subscription.updated',
      data: {
        object: { id: 'sub_1', metadata: { billing_deploy_target: EXPECTED } },
      },
    },
    {
      expectedTarget: EXPECTED,
      findOwnedSubscription: makeFindOwned(stub),
      loadAllowedProductIds: makeLoadAllowed(stub),
    }
  );
  assert.deepEqual(decision, { action: 'process' });
});
```

---

### [M2] `ownedOk` mixes an ignore-decision and a data-row as the return value — fragile discriminant

```js
const ownedOk = async subId => {
  ...
  return row;  // either { action: 'ignore', reason: '...' }  or the billing_subscriptions row
};

const decision = await ownedOk(stripeSub.id);
if (decision && 'action' in decision) return decision;   // relies on no 'action' column
return { action: 'process' };
```

The discriminant `'action' in decision` assumes `billing_subscriptions` rows will never have an `action` column. That's true today, but it's a silent footgun: if anyone adds `action` to `OWNED_SUBSCRIPTION_SELECT`, the classifier will silently drop all previously-owned events.

A safer return shape:

```js
// On ignore:
return { _ignore: { action: 'ignore', reason: '...' } };
// On success:
return { _ignore: null, row };

// Call site:
const result = await ownedOk(subId);
if (result._ignore) return result._ignore;
// result.row is now the typed subscription row
```

Or, simpler: return `null` for "not owned" and let the caller handle that with `if (!result) return { action: 'ignore', ... }`. The current approach merges two distinct concerns (is-there-an-ignore-decision, and what's-the-row) into one return path in a way that's hard to type correctly.

The same concern applies to `requireOwnedSubscription` in `index.ts`, though `'status' in owned` is a better discriminant since `OwnedSubscriptionRow` deliberately omits the `status` column.

---

### [M3] Non-subscription invoices are now silently dropped — behavioral change needs a comment

Old behavior: `syncInvoiceRow` was called unconditionally for `invoice.created / .finalized / .voided / .paid / .payment_failed / .payment_succeeded`. If the invoice had no subscription, `resolveUserIdByStripeCustomer` was used as a fallback (which is now correctly removed).

New behavior: **all six invoice event types require a `stripe_subscription_id`** or they return `ignored: invoice_missing_subscription`. This means one-time Stripe invoices (no subscription) are now silently dropped for all event types, not just logged.

This is probably intentional for a subscription-only billing model. But:
1. There's no comment in the code or the doc update explaining this intentional scope restriction.
2. If an adopter adds a one-time charge workflow, they'll see `ignored` in `billing_webhook_events` with no explanation that this is by design.

Add a comment on the `invoice_missing_subscription` returns:
```ts
// BeakerStack only syncs subscription-backed invoices; one-time invoices are logged as ignored.
if (!subId) {
  return { status: 'ignored', reason: 'invoice_missing_subscription' };
}
```

---

### [m4] `allowedProductIdsCache` is a module-level singleton that never invalidates

```ts
let allowedProductIdsCache: Set<string> | null = null;

async function loadAllowedProductIds(supabase: SupabaseClient): Promise<Set<string>> {
  if (allowedProductIdsCache) return allowedProductIdsCache;
  ...
}
```

Two notes:

1. **Staleness:** The cache lives for the entire isolate lifetime. If `billing_products` is updated without a deployment (e.g. a direct DB row insert), the next webhook event will use stale product IDs until the isolate cold-starts. This is a minor practical concern since `billing_products` is effectively config, but it should be documented.

2. **Concurrent init:** Deno is single-threaded but async. Two events arriving before the first query resolves will both find `null` and both issue a `billing_products` query. Harmless (both write the same data) but worth knowing.

Add a comment explaining why this is acceptable:
```ts
// Cache persists for the isolate lifetime; billing_products is updated only on deployment,
// which causes a cold start. If you add a product mid-deployment, manually restart the function.
let allowedProductIdsCache: Set<string> | null = null;
```

---

### [m5] Duplicate invoice classification logic across three case groups

The body of each invoice case in `classifyStripeEventCore` is identical:
```js
const subId = stripeSubscriptionIdFromRef(invoice.subscription);
if (!subId) return { action: 'ignore', reason: 'invoice_missing_subscription' };
const decision = await ownedOk(subId);
if (decision && 'action' in decision) return decision;
return { action: 'process' };
```

This 5-line block appears three times (for payment_failed; paid+payment_succeeded; created+finalized+voided). Extract to a `classifyInvoiceEvent(invoice, ownedOk)` helper before the switch. Minor but lowers future diff noise.

---

### [m6] `deployTargetMismatch` in `processStripeEvent` checkout branch is now dead code, and the warn log lost context

Since `classifyStripeEvent` runs at ingress before `processStripeEvent`, a deploy-target-mismatched checkout is already caught and returns early at line ~180. The `deployTargetMismatch` check inside `case 'checkout.session.completed'` in `processStripeEvent` is unreachable under normal conditions (defense-in-depth is fine) but the warn log now only includes `sessionId` — it lost the `expected` and `got` values that made it debuggable. Since `deployTargetMismatch` is now a one-arg wrapper that calls `getBillingDeployTarget()` internally, those values aren't easily available at the call site. At minimum, add a comment noting this is defense-in-depth and why the log is less detailed.

---

### Minor observations (not requiring changes)

- `customer.subscription.created` falls to `default: process` in the classifier. Fine today since `processStripeEvent` also doesn't handle it. If it's ever added to `processStripeEvent`, remember to add it to the classifier.
- `stripeSubscriptionIdFromRef` is exported from both `billing-webhook-guards-core.mjs` and re-exported from `billing-webhook-guards.ts` (via `export { stripeSubscriptionIdFromRef }`). Index imports it from the wrapper — ✓.
- The `OWNED_SUBSCRIPTION_SELECT` constant deliberately omits the `status` column, which is what makes the `'status' in owned` discriminant in `requireOwnedSubscription` correct. That's a subtle invariant — worth a comment on the constant.

supabase/functions/_shared/billing-webhook-guards-core.mjs
```diff
@@ -0,0 +1,147 @@
...
+ *   findOwnedSubscription: (stripeSubscriptionId: string) => Promise<unknown | null>,
+ *   loadAllowedProductIds?: () => Promise<Set<string> | null>,
+ * }} deps
+ * @returns {Promise<ClassifyDecision>}
+ */
+export async function classifyStripeEventCore(event, deps) {
+  const { expectedTarget, findOwnedSubscription, loadAllowedProductIds } = deps;
+
+  const ownedOk = async subId => {
+    if (!subId) return null;
```
diego-artificerinnovations: nit (silent logic gap): `ownedOk` returns `null` when `subId` is falsy, but the callers only short-circuit on `if (decision && 'action' in decision)` — so a null falls through to `{ action: 'process' }`. A malformed subscription event with no `id` would be classified as processable.

Fix: return an ignore decision instead of null, which lets you drop the `&& decision` guard from every call site:

```js
// before
if (!subId) return null;

// after
if (!subId) return { action: 'ignore', reason: 'unknown_stripe_subscription' };
```

And callers become:

```js
if ('action' in decision) return decision;
```

In practice Stripe always sets `id` on subscription objects, so this won't trigger in production — but the null-as-sentinel pattern is unnecessarily tricky.

supabase/functions/stripe-webhook/index.ts
```diff
@@ -203,23 +226,31 @@ Deno.serve(async req => {
...
     case 'checkout.session.completed': {
       const session = event.data.object as Stripe.Checkout.Session;
-      const expectedTarget = getBillingDeployTarget();
-      const metaTarget = session.metadata?.billing_deploy_target?.trim();
-      if (
-        metaTarget !== undefined &&
-        metaTarget !== '' &&
-        metaTarget !== expectedTarget
-      ) {
+      if (deployTargetMismatch(session.metadata ?? undefined)) {
```
diego-artificerinnovations: nit: the `deployTargetMismatch` check here is dead code. `classifyStripeEvent` already handles `checkout.session.completed` deploy-target filtering at the ingress (lines 148–169 above) and would have set `ingressIgnored = true` before `processStripeEvent` is ever called with this event. If the ingress passed it through, `deployTargetMismatch` will return `false` here by definition.

The check (and the `console.warn` below it) can be removed. If you want belt-and-suspenders, a comment is clearer than duplicate live code.

scripts/__tests__/billing-webhook-guards.test.mjs
```diff
@@ -0,0 +1,232 @@
...
+test('classify: unhandled event type processes', async () => {
+  const decision = await classifyStripeEventCore(
+    { type: 'payment_intent.succeeded', data: {} },
+    {
+      expectedTarget: EXPECTED,
+      findOwnedSubscription: makeFindOwned(stubSupabase()),
+    }
+  );
+  assert.deepEqual(decision, { action: 'process' });
+});
```
diego-artificerinnovations, line 369: nit: `customer.subscription.trial_will_end` shares the same classifier branch as `subscription.updated/deleted` (all three fall into one `case` block in `classifyStripeEventCore`) and calls `ownedOk` the same way, but it has no tests here. Worth adding the foreign-subscription path — it's one cheap case:

```js
test('classify: trial_will_end for foreign subscription ignores', async () => {
  const stub = stubSupabase(new Map()); // no owned rows
  const decision = await classifyStripeEventCore(
    {
      type: 'customer.subscription.trial_will_end',
      data: { object: { id: 'sub_foreign' } },
    },
    {
      expectedTarget: EXPECTED,
      findOwnedSubscription: makeFindOwned(stub),
      loadAllowedProductIds: makeLoadAllowed(stub),
    }
  );
  assert.deepEqual(decision, { action: 'ignore', reason: 'unknown_stripe_subscription' });
});
```

The owned path (processes) is implicitly exercised by the subscription tests, but the ignore path is untested.

diego-artificerinnovations (approved):
Strong PR. The core design choices are all right: `billing_deploy_target` enforcement, `requireOwnedSubscription` gating every mutation, removal of the `stripe_customer_id` fallback lookup, redacted-payload logging for ignored events, and the DI pattern that makes the classifier purely testable in Node. Three nits below — one is a silent logic gap worth fixing.

ZappoMan:
Thanks for the thorough review, @sarah-artificerinnovations — addressed in 8005fab:

- **M1:** Added owned `customer.subscription.updated` happy-path test (plus `trial_will_end` ignore test from Diego).
- **M2:** Replaced `ownedOk` with `resolveOwnedSubscription` returning `{ type: 'ignore' | 'owned' }` so we no longer overload row vs decision via `'action' in decision`.
- **M3:** Documented subscription-only invoice scope in `classifyInvoiceEvent`, handler branches, and setup docs.
- **m4:** Dropped the isolate-level product-id cache; load once per webhook request instead.
- **m5:** Consolidated invoice classification into `classifyInvoiceEvent`.
- **m6:** Removed dead checkout `deployTargetMismatch` handler check (ingress already filters).

Also incorporated Copilot/Diego inline feedback: Postgrest error handling, ingress ignore update failure → 500, `SupabaseClient` type import, and subscription events without `id` → ignore.

supabase/functions/_shared/billing-webhook-guards.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export async function findOwnedSubscription(
+  supabase: SupabaseClient,
+  stripeSubscriptionId: string
+): Promise<OwnedSubscriptionRow | null> {
+  const { data } = await supabase
+    .from('billing_subscriptions')
+    .select(OWNED_SUBSCRIPTION_SELECT)
+    .eq('stripe_subscription_id', stripeSubscriptionId)
+    .maybeSingle();
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 8005fab: `findOwnedSubscription` now checks Postgrest `error` and throws so the webhook returns 500 and Stripe retries instead of treating a DB failure as `unknown_stripe_subscription`.

supabase/functions/_shared/billing-webhook-guards.ts
```diff
@@ -0,0 +1,77 @@
...
+
+async function loadAllowedProductIds(
+  supabase: SupabaseClient
+): Promise<Set<string>> {
+  if (allowedProductIdsCache) return allowedProductIdsCache;
+  const { data } = await supabase.from('billing_products').select('id');
+  allowedProductIdsCache = new Set(
+    (data ?? []).map((p: { id: string }) => p.id)
+  );
+  return allowedProductIdsCache;
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 8005fab: `loadAllowedProductIds` logs and returns `null` on query failure, which disables the optional `product_id` allowlist for that request rather than treating every owned row as `unknown_product_id`.

supabase/functions/_shared/billing-webhook-guards.ts
```diff
@@ -0,0 +1,77 @@
...
+let allowedProductIdsCache: Set<string> | null = null;
+
+async function loadAllowedProductIds(
+  supabase: SupabaseClient
+): Promise<Set<string>> {
+  if (allowedProductIdsCache) return allowedProductIdsCache;
+  const { data } = await supabase.from('billing_products').select('id');
+  allowedProductIdsCache = new Set(
+    (data ?? []).map((p: { id: string }) => p.id)
+  );
```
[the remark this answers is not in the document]
ZappoMan: Addressed in 8005fab: removed the module-level cache; allowed product ids are loaded once per webhook request inside `classifyStripeEvent`, so new `billing_products` rows are visible on the next event without waiting for an isolate restart.

supabase/functions/stripe-webhook/index.ts
```diff
@@ -169,6 +180,18 @@ Deno.serve(async req => {
...
+  if (ingressIgnored) {
+    await supabase
+      .from('billing_webhook_events')
+      .update({
+        processed: true,
+        processed_at: new Date().toISOString(),
+        error: `ignored: ${decision.reason}`,
+      })
+      .eq('stripe_event_id', event.id);
+    return jsonResponse({ received: true, ignored: true }, 200, req);
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 8005fab: the ingress ignored path now checks the update `{ error }` and returns 500 with `log_failed` when marking the row processed fails, so Stripe can retry instead of leaving `processed=false` while returning 200.

docs/stripe-billing-setup.md
```diff
@@ -81,6 +81,24 @@ Stripe **test mode** can deliver the same events to every registered webhook URL
...
+
+1. **Separate Supabase project per app** — each deployment gets its own `auth.users` and `billing_*` tables.
+2. **Separate Stripe webhook endpoint per project** — unique URL and `STRIPE_WEBHOOK_SECRET` (`whsec_…`); never reuse secrets across projects.
+3. **App-scoped `productId`** in each app’s `billing-sync.json` / seeds (e.g. `myapp`, `myotherapp`), not a bare name like `pro`. Required if two apps ever share one Supabase database (`UNIQUE (user_id, product_id)`).
+4. Deploy **`billing-stripe` before `stripe-webhook`** on every project that shares the Stripe account so subscription metadata includes `billing_deploy_target`.
+
+**Webhook hardening (`stripe-webhook`):**
+
+- **Ingress classification** runs after signature verification and **before** inserting into `billing_webhook_events`. Events classified as foreign are logged with a **redacted** payload (`redacted: true`, no customer/invoice bodies) and return HTTP **200** `{ "received": true, "ignored": true }` without mutating subscriptions or invoices.
```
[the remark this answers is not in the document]
ZappoMan: Clarified in 8005fab: the doc now states classification runs before processing billing tables and before choosing full vs redacted `payload`, while still inserting a row for idempotency.

supabase/functions/_shared/billing-webhook-guards.ts
```diff
@@ -0,0 +1,77 @@
+import type Stripe from 'npm:stripe@14.21.0';
+import type { createClient } from 'npm:@supabase/supabase-js@2.45.0';
+import { getBillingDeployTarget } from './billing-deploy-target.ts';
+import {
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 8005fab: `SupabaseClient` is imported as a proper type from `@supabase/supabase-js` instead of `ReturnType<typeof createClient>` with `import type` on the value.

supabase/functions/_shared/billing-webhook-guards-core.mjs
ZappoMan: Fixed in 8005fab: `resolveOwnedSubscription` returns an explicit `{ type: "ignore", decision }` when `subId` is missing (or not owned), so callers never fall through to `process` on a null sentinel.

supabase/functions/stripe-webhook/index.ts
ZappoMan: Removed in 8005fab: the duplicate `deployTargetMismatch` check in `checkout.session.completed` was dead after ingress classification; left a short comment that deploy-target filtering happens at ingress.

scripts/__tests__/billing-webhook-guards.test.mjs
ZappoMan: Added in 8005fab: `classify: trial_will_end for foreign subscription ignores` plus the owned subscription update happy-path test.

supabase/functions/stripe-webhook/index.ts
```diff
@@ -169,6 +180,18 @@ Deno.serve(async req => {
...
 
+  if (ingressIgnored) {
+    await supabase
+      .from('billing_webhook_events')
+      .update({
+        processed: true,
+        processed_at: new Date().toISOString(),
+        error: `ignored: ${decision.reason}`,
+      })
+      .eq('stripe_event_id', event.id);
```
rahul-artificerinnovations: **Bug: `.update()` result is not checked.** If the Supabase update fails here, the event row stays as `processed: false` with no `error` field — it looks like a pending unprocessed event. Stripe already gets HTTP 200 and will not retry, so the row is permanently stuck in that misleading state.

Fix:
```ts
const { error: markErr } = await supabase
  .from('billing_webhook_events')
  .update({ processed: true, processed_at: new Date().toISOString(), error: `ignored: ${decision.reason}` })
  .eq('stripe_event_id', event.id);
if (markErr) throw asErrorFromSupabase(markErr);
```
Throwing causes the handler to return 500, Stripe retries, hits the 23505 duplicate-key path, finds `processed: false`, and re-attempts the update — correct retry behaviour. Alternatively log-and-continue is acceptable if you'd rather not risk a retry loop on a persistent DB issue, but at minimum the error must surface somewhere.

supabase/functions/stripe-webhook/index.ts
```diff
@@ -203,23 +226,31 @@ Deno.serve(async req => {
   return jsonResponse({ received: true }, 200, req);
 });
 
+async function requireOwnedSubscription(
```
rahul-artificerinnovations, line 258: **Double subscription lookup for every owned event.** `classifyStripeEvent` → `classifyStripeEventCore` → `findOwnedSubscription` already queries `billing_subscriptions` during ingress; `requireOwnedSubscription` queries the same row again inside the handler for every subscription and invoice event type.

For the volume typical of a SaaS webhook this is fine, but if you want to eliminate it: extend `ClassifyDecision`'s `process` variant to carry the fetched row:
```ts
| { action: 'process'; subscription?: OwnedSubscriptionRow }
```
and pass it through from `ownedOk` → `classifyStripeEventCore` → `classifyStripeEvent`. `requireOwnedSubscription` can then short-circuit when the row is already present. Not a blocker — noting it for the next billing pass.

supabase/functions/_shared/billing-webhook-guards.ts
```diff
@@ -0,0 +1,77 @@
...
+): Promise<OwnedSubscriptionRow | null> {
+  const { data } = await supabase
+    .from('billing_subscriptions')
+    .select(OWNED_SUBSCRIPTION_SELECT)
+    .eq('stripe_subscription_id', stripeSubscriptionId)
+    .maybeSingle();
+  return (data as OwnedSubscriptionRow | null) ?? null;
+}
+
+let allowedProductIdsCache: Set<string> | null = null;
```
rahul-artificerinnovations: `allowedProductIdsCache` is module-level with no TTL or invalidation. In Deno Deploy, a warm worker can live for hours; a new `billing_products` row added after deploy won't be visible until the next cold start. This is almost certainly acceptable (product catalogue changes are rare and a function redeploy resets it), but it should be documented so a future engineer doesn't chase a "why isn't my new product recognised" incident:
```ts
// Cached for worker lifetime — no TTL. A billing_products change takes effect
// on the next cold start or function redeploy.
let allowedProductIdsCache: Set<string> | null = null;
```

scripts/__tests__/billing-webhook-guards.test.mjs
```diff
@@ -0,0 +1,232 @@
...
+    {
+      expectedTarget: EXPECTED,
+      findOwnedSubscription: makeFindOwned(stub),
+      loadAllowedProductIds: makeLoadAllowed(stub),
+    }
+  );
+  assert.deepEqual(decision, { action: 'process' });
+});
+
+test('classify: subscription deploy target mismatch ignores', async () => {
```
rahul-artificerinnovations, line 166: This test covers `customer.subscription.updated` for the deploy-target-mismatch path. `customer.subscription.deleted` and `customer.subscription.trial_will_end` share the same switch arm in the core but are never used as the event `type` in any test. Given they have meaningfully different real-world semantics (deletion triggers cancel, trial_will_end is advisory), a quick copy of this test for each missing type would give explicit confidence that the classify path is entered correctly for those event names — especially useful if the switch arms are ever split.

rahul-artificerinnovations (changes requested):
The deploy-target isolation architecture is solid: classify at ingress, redact foreign payloads, thread `userId` through `syncInvoiceRow` to eliminate the `resolveUserIdByStripeCustomer` fallback, and drop the dangerous `billing_subscriptions`-by-customer-id lookup. The `.mjs` core module + Deno wrapper split makes the logic fully testable in Node. One genuine bug requires a fix before merge; three smaller observations inline.

sarah-artificerinnovations (approved):
All three required fixes are in and correct.

**M1 — happy-path test:** `classify: owned subscription update without mismatch processes` asserts `{ action: 'process' }` for an owned sub with a matching deploy target. Covers the positive path through `customer.subscription.updated`.

**M2 — `resolveOwnedSubscription` return type:** Clean discriminated union — `{ type: 'ignore', decision }` | `{ type: 'owned', row }`. Callers branch on `result.type === 'ignore'`, which TypeScript can check exhaustively. No mixed shapes.

**M3 — one-time invoice documentation:** JSDoc on `classifyInvoiceEvent` plus inline comment on every `invoice.*` handler: "BeakerStack only syncs subscription-backed invoices; one-time invoices are ignored." Explicit at both the logic layer and the handler layer.

**Bonus:** `resolveUserIdByStripeCustomer` is gone entirely — `syncInvoiceRow` now takes `userId: string` sourced directly from the owned subscription row. The old customer-lookup race (REQ-046 workaround) is eliminated structurally.

Minor nits all addressed: per-request `loadAllowedProductIds` with scope comment, `classifyInvoiceEvent` consolidates all invoice logic, stale comments removed.

This is the right architecture for multi-app Stripe account safety. Approving.

supabase/functions/stripe-webhook/index.ts
ZappoMan: Fixed in fb84da6: introduced `markWebhookEventProcessed` (throws on Postgrest error) and use it for ingress ignored, handler ignored, and handler success paths. Failures now return 500 so Stripe retries and the duplicate-key reprocess path can recover.

supabase/functions/stripe-webhook/index.ts
ZappoMan: Agreed this is a worthwhile optimization. Deferring to a follow-up: carrying `OwnedSubscriptionRow` on `{ action: 'process' }` would remove the second lookup without changing behavior. Not blocking merge for typical webhook volume.

supabase/functions/_shared/billing-webhook-guards.ts
ZappoMan: This was removed in 8005fab: there is no module-level `allowedProductIdsCache` anymore. `loadAllowedProductIds` runs once per webhook request inside `classifyStripeEvent` (see current `billing-webhook-guards.ts`).

ZappoMan:
Babysit update (fb84da6): fixed Rahul's handler-path `markWebhookEventProcessed` bug and extended subscription deploy-target tests. Rahul's cache comment referred to code removed in 8005fab; double-lookup optimization deferred. CI was green on 8005fab; re-running on latest push.

scripts/__tests__/billing-webhook-guards.test.mjs
ZappoMan: Addressed in fb84da6: the deploy-target mismatch test now loops over `customer.subscription.updated`, `.deleted`, and `.trial_will_end` so all three switch arms are covered explicitly.

supabase/functions/stripe-webhook/index.ts
ZappoMan: Implemented in the latest push: `ClassifyDecision` process results now include `ownedSubscription` when ingress already resolved the row. `processStripeEvent` passes it to `requireOwnedSubscription`, which skips the second `billing_subscriptions` query when preloaded.

rahul-artificerinnovations (approved):
All four observations addressed, and then some.

On the bug fix: `markWebhookEventProcessed` is the right abstraction — it consolidates the three inline `.update()` blocks, throws on error, and the `ingressIgnored` path is now inside the try/catch, so Stripe gets 500 (will retry) if the mark fails rather than a 200 that hides the DB failure.

On the double-lookup: the `ownedSubscription` threading is clean. `ProcessDecision` carries the fetched row, `ownedSubscriptionFromDecision` extracts it safely at the call site, and `requireOwnedSubscription` short-circuits via `preloaded`. The lookup-count test (`lookups === 1`) pins the invariant explicitly — good discipline.

On `allowedProductIdsCache`: removing the module-level cache and loading per-request is the right call over adding a TTL. One extra `billing_products` query per webhook is negligible, and `loadAllowedProductIds` now returns `null` on failure and skips the product guard gracefully rather than crashing the classification.

On test coverage: the parameterized loop over `updated`/`deleted`/`trial_will_end` for the mismatch path, plus the new `trial_will_end` foreign-subscription and no-id tests, cover the cases cleanly. Good to merge.

supabase/functions/stripe-webhook/index.ts
```diff
@@ -142,18 +149,23 @@ Deno.serve(async req => {
     return jsonResponse({ error: 'invalid_signature' }, 400, req);
   }
 
+  const decision = await classifyStripeEvent(supabase, event);
+  const ingressIgnored = decision.action === 'ignore';
+  const payloadForLog = ingressIgnored
+    ? redactedWebhookPayload(event)
+    : (event as unknown as Record<string, unknown>);
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 3875845: `classifyStripeEvent` is wrapped in try/catch before any `billing_webhook_events` insert; failures return `{ error: processing_failed }` with HTTP 500 so Stripe can retry.

supabase/functions/stripe-webhook/index.ts
```diff
@@ -142,18 +149,23 @@ Deno.serve(async req => {
...
+
   const { error: logErr } = await supabase
     .from('billing_webhook_events')
     .insert({
       stripe_event_id: event.id,
       event_type: event.type,
-      payload: event as unknown as Record<string, unknown>,
+      payload: payloadForLog,
       processed: false,
     });
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 3875845: `webhookPayloadForLog` redacts payloads for ignored events and for event types outside `BILLING_WEBHOOK_HANDLED_EVENT_TYPES` (e.g. `payment_intent.*`), so foreign fan-out no longer stores full bodies for unhandled types.

supabase/functions/_shared/billing-webhook-guards.ts
```diff
@@ -0,0 +1,89 @@
...
+export async function classifyStripeEvent(
+  supabase: SupabaseClient,
+  event: Stripe.Event
+): Promise<ClassifyDecision> {
+  const allowedProductIds = await loadAllowedProductIds(supabase);
+  return classifyStripeEventCore(event, {
+    expectedTarget: getBillingDeployTarget(),
+    findOwnedSubscription: id => findOwnedSubscription(supabase, id),
+    allowedProductIds,
+  }) as Promise<ClassifyDecision>;
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 3875845: `billing_products` is loaded lazily via `loadAllowedProductIds()` only inside `resolveOwnedSubscription` (subscription/invoice paths). Checkout and default/unhandled types skip that query.

mei-artificerinnovations (commented):
Good architecture overall. The core/.mjs + Deno wrapper split is the right call for testability — tests run clean in Node without a Deno shim. Three things worth addressing before merge:

supabase/functions/stripe-webhook/index.ts
```diff
@@ -170,26 +189,27 @@ Deno.serve(async req => {
...
+    });
     if (result.status === 'ignored') {
-      await supabase
-        .from('billing_webhook_events')
-        .update({
-          processed: true,
-          processed_at: new Date().toISOString(),
-          error: `ignored: ${result.reason}`,
-        })
-        .eq('stripe_event_id', event.id);
```
mei-artificerinnovations, line 182: This deploy-target re-check is dead code in the live path — `classifyStripeEvent` already rejects mismatches before `processStripeEvent` is ever called. Compare `checkout.session.completed` above which has a comment and no re-check. Either add `// defense-in-depth — classifier already enforces this` here, or remove it to match the checkout pattern. As written it looks like the primary guard, which could mislead someone who reads this function in isolation and assumes it's safe to call without pre-classification.

supabase/functions/_shared/billing-webhook-guards.ts
```diff
@@ -0,0 +1,99 @@
...
+  metadata: Record<string, string> | null | undefined
+): boolean {
+  return deployTargetMismatchCore(metadata, getBillingDeployTarget());
+}
+
+export function redactedWebhookPayload(
+  event: Stripe.Event
+): Record<string, unknown> {
+  return redactedWebhookPayloadCore(event);
+}
```
mei-artificerinnovations, line 38: `loadAllowedProductIds` returns `null` on error (guard skipped); `findOwnedSubscription` throws on error (request 500s). The asymmetry is intentional per the comment, but the contract is invisible at the call site in core.mjs — `if (allowed && !allowed.has(...))` silently bypasses the product guard when `allowed` is null, with nothing in the core JSDoc explaining that null means skip. During a `billing_products` outage every webhook passes the product guard. For multi-app isolation that's the primary purpose of this PR — worth either a JSDoc `@param` note in core or a stronger comment here explaining the deliberate fail-open trade-off.

supabase/functions/stripe-webhook/index.ts
```diff
@@ -203,29 +223,52 @@ Deno.serve(async req => {
   return jsonResponse({ received: true }, 200, req);
 });
```
mei-artificerinnovations, line 241: Soft data-integrity issue on cancellation retries: this update sets `stripe_subscription_id = null`, then `markWebhookEventProcessed` is called. If the mark-processed write fails (transient), the event stays `processed = false`. On retry, `classifyStripeEvent` calls `findOwnedSubscription(stripeSub.id)` → no row (already nulled) → classifies as `ignore: unknown_stripe_subscription`. Billing row is correct but the audit log shows "ignored" for a webhook that was actually processed, which makes cancellation debugging misleading. One fix: do the `stripe_subscription_id = null` update last, after `markWebhookEventProcessed` succeeds. Or query the deletion by `user_id + product_id + stripe_subscription_id` (not just sub ID) so a partial retry is still findable.

supabase/functions/stripe-webhook/index.ts
ZappoMan: Addressed in 2167845: removed the handler deploy-target re-check and added the same ingress comment pattern as checkout (classifier already enforces deploy target).

supabase/functions/_shared/billing-webhook-guards.ts
ZappoMan: Documented in 2167845: JSDoc on `loadAllowedProductIds` in core and an explicit fail-open comment in the Deno wrapper — `null` skips the product_id allowlist while `findOwnedSubscription` still throws on DB errors.

supabase/functions/stripe-webhook/index.ts
ZappoMan: Fixed in 2167845: cancellation keeps `stripe_subscription_id` through the billing update; `clearStripeIdsFor` nulls Stripe ids only after `markWebhookEventProcessed` succeeds so Stripe retries still classify as owned.

Issue #257, opened 2026-05-18
Improve unit test coverage


ZappoMan:
## Summary

Repo-wide statement coverage is **~97.3%** (`npm run test:coverage`), but **28 of 194** instrumented files are still **below 95% statement coverage**. This issue tracks that catalog and the highest-impact gaps.

The original scope — **`packages/billing/src/presentation/formatters.ts`** (9.68% statements, no co-located unit tests) — remains the top billing outlier. Behavior is only exercised from the web app today (`apps/web/src/billing/__tests__/formatters.test.ts`). `docs/ARCHITECTURE.md` references `packages/billing/.../formatters.test.ts`, but that file does not exist.

## Statement coverage catalog (< 95%)

Measured from integrated coverage (`coverage-final.json` across web, mobile, shared-tests, billing). Sorted by coverage % ascending within each package.

### `apps/web` (3 files — package ~99.2% statements)

| Stmt % | Covered | File |
|--------|---------|------|
| 78.38% | 29/37 | `apps/web/src/lib/supabase.ts` |
| 91.94% | 57/62 | `apps/web/src/components/ScrollToTop.tsx` |
| 93.75% | 30/32 | `apps/web/src/lib/authStorageHint.ts` |

### `apps/mobile` (12 files — package ~89.5% statements)

| Stmt % | Covered | File |
|--------|---------|------|
| 0.00% | 0/23 | `apps/mobile/src/billing/useDemoCollectionCount.ts` |
| 0.00% | 0/1 | `apps/mobile/src/components/dashboard/LucideLockIcon.tsx` |
| 0.00% | 0/2 | `apps/mobile/src/navigation/BillingNavigator.tsx` |
| 18.75% | 3/16 | `apps/mobile/src/lib/randomUuid.ts` |
| 50.53% | 48/95 | `apps/mobile/src/components/dashboard/CollectionDetail.tsx` |
| 60.00% | 3/5 | `apps/mobile/src/components/dashboard/utils.ts` |
| 63.33% | 19/30 | `apps/mobile/src/screens/billing/BillingOverviewScreen.tsx` |
| 80.00% | 4/5 | `apps/mobile/src/billing/mobileBillingUrls.ts` |
| 82.76% | 24/29 | `apps/mobile/src/screens/ProfileScreen.tsx` |
| 90.00% | 9/10 | `apps/mobile/src/screens/billing/BillingTabBar.tsx` |
[...]

PR #259, opened 2026-05-18
test(#257): coverage gaps across billing, web, and mobile

mei-artificerinnovations:
Fixes #257

## Summary

Raises test coverage toward the ≥95% target from #257 by adding targeted unit tests across `packages/billing`, `apps/web`, `apps/mobile`, and `packages/shared-tests`. Also co-locates billing presentation formatter tests in the package (with a slim web entrypoint smoke test) and pins date formatters to UTC so assertions are stable in every timezone.

### Billing package (`packages/billing`)
- **Formatters:** Co-located `formatters.test.ts` with stable en-US assertions; web duplicate reduced to entrypoint smoke test.
- **Production fix:** `formatDate` / `formatMonthYear` now pass `timeZone: 'UTC'` so locale formatting does not shift by runner timezone.
- **Components & hooks:** Additional cases for `FeatureGate` (null/empty key early exit), `UpgradePrompt.native`, `CustomerPortalLink.native`, `useUsage` (RPC error + realtime non-match branches).
- **Utils:** `parseBillingFunctionError` (no-hint + null fallback), `openExternalUrl.native` (unparseable URL + redacted host in error).

### Web (`apps/web`)
- `authStorageHint`: non-string and empty `access_token` branches (merged into main test file).
- `supabase.ts`: Node-env `Logger.warn` / `debug` paths when env vars missing or realtime URL is `ws://`.
- `ScrollToTop`: hash scroll false/true paths, 60-frame RAF cap; lint fixes for non-null assertions and DEV `console.warn`.

### Mobile (`apps/mobile`)
- `getMobileBillingProviderUrls`: deep-link default, custom scheme, http/https path segments.
- `useDemoCollectionCount`: loading, null data, error wrapping, refresh.
- `randomUuid`: all three crypto fallbacks (with guarded `globalThis.crypto` restore).
- `BillingTabBar`, `BillingNavigator`, `limLabel`, `LucideLockIcon` render smoke tests.

### Shared (`packages/shared-tests`)
- `ProfileAvatar.web`: `onError` initials fallback (with React-tree cleanup workaround).
- `ProtectedRoute.native`: authenticated children render path.

## Test plan

- [x] CI `tests` workflow passes
- [x] `npm test -w packages/billing`
- [x] Mobile/web coverage suites for touched packages

## Notes

- Coverage-only cases were consolidated into existing `*.test.ts(x)` files rather than leaving parallel `*.coverage.test.*` files.
- Follow-up from review: timezone pinning in formatters, crypto mutation guards, BillingTabBar/Navigator mock fixes.


packages/billing/src/presentation/formatters.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  });
+});
+
+describe('formatMonthYear', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('formats a valid ISO string as en-US month + year', () => {
+    expect(formatMonthYear('2024-03-01T00:00:00.000Z')).toBe('March 2024');
```
ZappoMan, line 58: **Timezone flake (reproduced locally):** `2024-03-01T00:00:00.000Z` is still Feb 29 in US timezones west of UTC, so this assertion receives `'February 2024'` instead of `'March 2024'`. CI likely passes because runners use UTC.

**Suggestion:** use a midday UTC anchor (same pattern as `formatDate` on L35), e.g. `'2024-03-15T12:00:00.000Z'`, or set `process.env.TZ = 'UTC'` in a `beforeAll` for this file.

packages/billing/src/presentation/formatters.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('formats a valid ISO string as en-US short date', () => {
+    expect(formatDate('2024-06-15T12:00:00.000Z')).toBe('Jun 15, 2024');
+  });
+
+  it('returns a string for invalid date input without throwing', () => {
+    expect(formatDate('not-a-date')).toBeTypeOf('string');
```
ZappoMan, line 39: These invalid-input cases only assert `toBeTypeOf('string')`, which any non-throwing return satisfies (including the raw input). That's weaker than the old web tests' `toBeTruthy()`. If the product expectation for `'not-a-date'` / `'bad'` is documented anywhere, consider pinning it (e.g. exact fallback string) so regressions in `Date` parsing don't slip through silently.

apps/web/src/billing/__tests__/formatters.test.ts
```diff
@@ -1,78 +1,16 @@
...
-
-  it('returns original string when toLocaleDateString throws', () => {
-    vi.spyOn(Date.prototype, 'toLocaleDateString').mockImplementation(() => {
-      throw new Error('locale');
-    });
-    expect(formatMonthYear('2024-03-01T00:00:00.000Z')).toBe(
-      '2024-03-01T00:00:00.000Z'
-    );
+// Behavior is covered by packages/billing/src/presentation/formatters.test.ts.
+// This file confirms the re-export resolves through the package entrypoint.
```
ZappoMan, line 9: 👍 Nice split: behavior lives in the package, this file only guards the `@beakerstack/billing/presentation` entrypoint wiring. Clear comment explains the intent.

apps/web/src/__tests__/authStorageHint.coverage.test.ts
```diff
@@ -0,0 +1,27 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import { getLikelyAuthenticated } from '../lib/authStorageHint';
+
+describe('authStorageHint (coverage)', () => {
```
ZappoMan: The numeric `access_token` branch is a good gap-fill, but the empty-string case duplicates the spirit of `authStorageHint.test.ts` L58–60 (`token missing`). Consider folding both cases into `authStorageHint.test.ts` instead of a separate `.coverage.test.ts` file — keeps related assertions in one place and matches how `parseBillingFunctionError` already has a main test file.

ZappoMan (commented):
## Summary

Solid direction on **co-locating formatter tests** in `packages/billing` and slimming the web re-export smoke test — that's the right layering.

The follow-up coverage commit is thorough and CI is green, but I have a few concerns below (especially a **timezone flake** in `formatMonthYear` that fails locally on my machine in US/Pacific).

**Verdict:** Approve with minor changes — fix the `formatMonthYear` assertion (or pin TZ), and consider folding several `.coverage.test.*` cases into existing co-located suites to reduce file sprawl.

Thanks for the BillingTabBar / BillingNavigator mock fixes and the ProfileAvatar cleanup workaround.

packages/billing/src/utils/parseBillingFunctionError.coverage.test.ts
```diff
@@ -0,0 +1,16 @@
+import { describe, it, expect } from 'vitest';
+import { parseBillingFunctionError } from './parseBillingFunctionError.js';
+
+describe('parseBillingFunctionError (coverage)', () => {
```
ZappoMan: Same consolidation thought: `parseBillingFunctionError.test.ts` already exists with only 2 cases. These two branches fit naturally there as additional `it(...)` blocks rather than a parallel `*.coverage.test.ts` file.

apps/mobile/src/lib/__tests__/randomUuid.test.ts
```diff
@@ -0,0 +1,31 @@
...
+const { randomUuid } = require('../randomUuid') as typeof import('../randomUuid');
+
+const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/;
+
+describe('randomUuid', () => {
+  it('returns a valid v4 UUID via crypto.randomUUID (path 1)', () => {
+    expect(randomUuid()).toMatch(UUID_RE);
+  });
+
+  it('returns a valid v4 UUID via getRandomValues when randomUUID is absent (path 2)', () => {
```
ZappoMan, line 15: Path 2/3 mutate `globalThis.crypto` inline without `afterEach` restoration. If an assertion fails between delete and restore, later tests (or other files in the same worker) can see a broken `crypto`. Recommend `afterEach` that always restores `saved` / `savedCrypto`, or `try/finally` inside each test.

apps/mobile/src/components/dashboard/__tests__/LucideLockIcon.test.tsx
```diff
@@ -0,0 +1,14 @@
+import { describe, it, expect } from '@jest/globals';
+import React from 'react';
+import { render } from '@testing-library/react-native';
+import { LucideLockIcon } from '../LucideLockIcon';
+
+describe('LucideLockIcon', () => {
+  it('renders without crashing with default props', () => {
```
ZappoMan, line 7: `expect(() => render(...)).not.toThrow()` only proves the component mounts; it won't catch wrong `size`/`color` props or accessibility regressions. If the goal is coverage % only, this is fine — otherwise a snapshot or `getByRole`/testID assertion on the rendered SVG would add more signal per line.

apps/mobile/src/navigation/__tests__/BillingNavigator.test.tsx
```diff
@@ -0,0 +1,30 @@
+import { describe, it, expect, jest } from '@jest/globals';
+import React from 'react';
+import { render } from '@testing-library/react-native';
+
+// jest.mock factories are hoisted before imports, so React is not in scope inside them.
+// Use plain functions that return children/null without needing React.
+jest.mock('@react-navigation/native-stack', () => ({
```
ZappoMan, line 7: The hoisting-safe mocks (`children` passthrough, `Screen: () => null`) are a good fix. The `require()` after mocks is the right pattern. Smoke-only coverage is acceptable here given navigation wiring is mostly integration-tested elsewhere.

apps/mobile/src/screens/billing/__tests__/BillingTabBar.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+    expect(getByText('Overview')).toBeTruthy();
+    expect(getByText('Usage')).toBeTruthy();
+  });
+
+  it('marks Overview as selected and Usage as unselected when route is BillingOverview', () => {
+    const { getAllByRole } = render(<BillingTabBar />);
+    const tabs = getAllByRole('tab');
+    expect(tabs).toHaveLength(2);
+    // Check accessibilityState directly on the element props (no custom matchers needed)
+    expect(tabs[0].props.accessibilityState).toEqual({ selected: true });
```
ZappoMan, line 29: 👍 Direct `props.accessibilityState` check is clearer than depending on `jest-native` matchers that aren't in the preset. Tab press + `navigate` assertions cover the meaningful behavior.

apps/mobile/src/billing/__tests__/useDemoCollectionCount.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    await waitFor(() => expect(result.current.loading).toBe(false));
+    expect(result.current.maxItemsInAnyCollection).toBe(0);
+  });
+
+  it('refresh re-fetches and updates count', async () => {
+    mockRpc
+      .mockResolvedValueOnce({ data: [{ id: 'a', item_count: 1 }], error: null })
+      .mockResolvedValueOnce({
+        data: [{ id: 'a', item_count: 1 }, { id: 'b', item_count: 2 }],
+        error: null,
```
ZappoMan, line 76: 👍 Strong hook coverage: loading stall, null data, non-Error RPC error wrapping, Error preservation, null `item_count`, and `refresh`. This is one of the higher-value additions in the coverage commit.

apps/mobile/src/billing/__tests__/mobileBillingUrls.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+  });
+
+  it('uses deep-link scheme by default (beaker-stack://billing)', () => {
+    delete process.env[ENV_KEY];
+    const urls = getMobileBillingProviderUrls();
+    expect(urls.checkoutSuccessUrl).toBe('beaker-stack://billing?checkout=success');
+    expect(urls.checkoutCancelUrl).toBe('beaker-stack://billing?checkout=cancel');
+    expect(urls.portalReturnUrl).toBe('beaker-stack://billing');
+  });
```
ZappoMan, line 22: 👍 Env restore in `afterEach` is correct. Deep-link vs http/https matrix matches `mobileBillingUrls.ts` branches and documents cancel URL path difference (`/billing/plans` for http).

packages/billing/src/hooks/useUsage.coverage.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+  it('does not refetch when realtime product_id does not match', async () => {
+    rpc.mockResolvedValue({
+      data: { used: 1, limit: 5, remaining: 4, periodEnd: '', periodStart: '' },
+      error: null,
+    });
+    renderHook(() => useUsage('ai'));
+    await waitFor(() => expect(rpc).toHaveBeenCalledTimes(1));
+    rpc.mockClear();
+    usageRealtimeCb.current?.({ new: { product_id: 'other_product', event_type: 'ai' }, old: null });
+    await Promise.resolve();
```
ZappoMan: The realtime negative-path tests are valuable, but the double `await Promise.resolve()` flush is brittle — any async refetch scheduling change could make this flaky. Prefer `await waitFor(() => expect(rpc).not.toHaveBeenCalled())` with a short timeout, or `vi.waitFor` if you need zero calls after the callback.

packages/billing/src/components/FeatureGate.web.coverage.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+vi.mock('../hooks/useFeature.js', () => ({ useFeature: vi.fn() }));
+
+describe('FeatureGate (web) — null/empty key early exit', () => {
+  beforeEach(() => {
+    vi.mocked(useFeature).mockReset();
+  });
+
+  it('renders fallback without calling useFeature when no feature key is provided', () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const props = { fallback: <span>fallback-no-key</span> } as any;
```
ZappoMan: `as any` props are understandable for hitting `rawKey == null || rawKey === ''` early exit, but they bypass the type system entirely. Alternative: pass `feature={undefined}` / omit `feature` with a typed partial if the props type allows it, so the test documents the real public API surface.

apps/web/src/components/__tests__/ScrollToTop.coverage.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  it('stops after HASH_SCROLL_MAX_FRAMES (60) without finding the element', () => {
+    const rafQueue: FrameRequestCallback[] = [];
+    vi.spyOn(window, 'requestAnimationFrame').mockImplementation(cb => {
+      rafQueue.push(cb);
+      return rafQueue.length;
+    });
+    vi.spyOn(console, 'warn').mockImplementation(() => {});
+
+    startHashScroll('#never-mounts-xyz');
```
ZappoMan: This exercises the 60-frame RAF cap well, but it never asserts the DEV `console.warn` when the anchor is missing (the branch you added the eslint-disable for in `ScrollToTop.tsx`). Stubbing `import.meta.env.DEV` + spying `console.warn` would lock that behavior in and distinguish "stopped at frame 60" from "found element".

packages/shared-tests/__tests__/components/profile/ProfileAvatar.web.coverage.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+  it('hides the img and renders initials in parent when image fails to load', () => {
+    const { getByAltText } = render(<ProfileAvatar profile={baseProfile} />);
+    const img = getByAltText('Test User');
+    const parent = img.parentElement!;
+    fireEvent.error(img);
+    expect(img.style.display).toBe('none');
+    expect(parent.textContent).toContain('TU');
+    // onError replaces img via parent.innerHTML, detaching img from React's tree.
+    // Re-attach so React can removeChild(img) during cleanup without throwing.
+    parent.innerHTML = '';
```
ZappoMan, line 29: The `innerHTML` + re-attach workaround is well-documented and fixes the React cleanup crash — good detective work. Longer term, the component's imperative `parent.innerHTML` in `onError` is hard to test cleanly; extracting initials fallback into a pure helper (or using React state instead of DOM mutation) would make this testable without DOM surgery.

packages/billing/src/components/CustomerPortalLink.native.coverage.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+      openPortal: vi.fn().mockRejectedValue(new Error('portal down')),
+      pending: false,
+      error: null,
+    });
+    render(<CustomerPortalLink>Portal</CustomerPortalLink>);
+    fireEvent.click(screen.getByText('Portal'));
+    await waitFor(() => expect(screen.getByText('portal down')).toBeInTheDocument());
+  });
+
+  it('suppresses openError when portalError is also set (openError && !portalError is false)', async () => {
```
ZappoMan: 👍 The `openError && !portalError` suppression case is exactly the kind of branch that's easy to regress — good targeted coverage.

packages/shared-tests/__tests__/components/auth/ProtectedRoute.native.coverage.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+}));
+
+jest.mock('@beakerstack/shared/contexts/AuthContext', () => ({
+  useAuthContext: () => ({
+    user: { id: 'user-1', email: 'test@example.com' },
+    loading: false,
+  }),
+}));
+
+describe('ProtectedRoute (native) — authenticated render coverage', () => {
```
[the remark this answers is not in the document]
mei-artificerinnovations: Confirmed intentional, and only runs under web config (not twice). The native Jest config (`jest.native.cjs`) uses `testMatch: ["**/__tests__/**/*.native.test.{ts,tsx}"]` — my `.native.coverage.test.tsx` suffix doesn't match that pattern, so it's excluded from the native run. It runs only under the web config, where `react-native → react-native-web` mapping means `<View>` renders as `<div>`. Since the test mocks `useAuthContext` entirely, it's testing the conditional render logic rather than any native-specific behavior.

packages/shared-tests/__tests__/components/profile/ProfileAvatar.web.coverage.test.tsx
```diff
@@ -0,0 +1,38 @@
+import '@testing-library/jest-dom';
```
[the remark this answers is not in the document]
mei-artificerinnovations: Issue #257 explicitly lists `packages/shared/src/components/profile/ProfileAvatar.web.tsx` in the files needing coverage improvement. The `onError` handler branch (image load failure → initials fallback) had zero coverage. Included here to satisfy the issue requirement. Happy to move to a separate PR if scope is a concern.

packages/billing/src/presentation/formatters.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  });
+});
+
+describe('formatDate', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('formats a valid ISO string as en-US short date', () => {
+    expect(formatDate('2024-06-15T12:00:00.000Z')).toBe('Jun 15, 2024');
```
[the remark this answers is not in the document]
mei-artificerinnovations: Fixed in the same commit `ea8bccf4` — `timeZone: "UTC"` added to `formatDate` as well. While the test input (`12:00:00.000Z`) is safe for most timezones, the underlying inconsistency was the same and warranted fixing.

apps/mobile/src/navigation/__tests__/BillingNavigator.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+  createNativeStackNavigator: () => ({
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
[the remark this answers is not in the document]
mei-artificerinnovations: Fixed in commit `c638e614`. `jest.mock()` factories are hoisted before all imports, so `React` (imported on line 2) is not yet defined when the factory executes. Removed all `React.createElement`/`React.Fragment` references from the factory:

```typescript
createNativeStackNavigator: () => ({
  Navigator: ({ children }: any) => children,
  Screen: (_props: any) => null,
}),
```

Returning `children` directly avoids needing React in the hoisted scope.

apps/mobile/src/screens/billing/__tests__/BillingTabBar.test.tsx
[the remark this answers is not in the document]
mei-artificerinnovations: Fixed in commit `c638e614`. The mobile Jest preset (`preset: "react-native"`) does not load `@testing-library/jest-native`, so custom matchers like `toHaveAccessibilityState` are not registered. Replaced with a direct property check:

```typescript
expect(tabs[0].props.accessibilityState).toEqual({ selected: true });
expect(tabs[1].props.accessibilityState).toEqual({ selected: false });
```

This is equivalent and doesn't require extending the matcher set.

packages/billing/src/presentation/formatters.test.ts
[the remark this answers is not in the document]
mei-artificerinnovations: Fixed in commit `ea8bccf4`. Added `timeZone: "UTC"` to `toLocaleDateString` in `formatMonthYear` (and `formatDate` for consistency). Billing period dates are stored in UTC and should render as-stored regardless of host timezone:

```typescript
return new Date(iso).toLocaleDateString("en-US", {
  year: "numeric",
  month: "long",
  timeZone: "UTC",
});
```

`2024-03-01T00:00:00.000Z` is now consistently "March 2024" in all environments.

mei-artificerinnovations: Fixed in commit `ea8bccf4` — `timeZone: "UTC"` added to `formatMonthYear` (and `formatDate`). Reproduced and confirmed: `2024-03-01T00:00:00.000Z` in UTC-8 is Feb 29 local. Passing `timeZone: "UTC"` pins interpretation to the stored value.

packages/billing/src/presentation/formatters.test.ts
mei-artificerinnovations: The `toBeTypeOf("string")` assertion mirrors the spirit of the catch branch — "returns something that won't crash the caller." Agreed these could be tighter (e.g., assert the raw input is returned for `bad` since it can't be formatted). Left minimal to match the existing test file's pattern, but happy to strengthen if you'd prefer explicit return-value checks on the error paths.

packages/billing/src/hooks/useUsage.coverage.test.tsx
mei-artificerinnovations: The double `await Promise.resolve()` matches how the existing `useUsage.test.tsx` flushes the realtime subscription channel handler — one tick to let the channel callback run, one tick to let any queued state update settle. It's fragile, agreed. The alternative would be `act()` wrapping, but the hook uses `supabase.channel()` which doesn't integrate with React's `act` queue. Left it matching the existing test pattern. If this becomes a source of flake, extracting the flush into a helper that retries with a timeout would be more robust.

apps/mobile/src/lib/__tests__/randomUuid.test.ts
mei-artificerinnovations: Fixed in commit `ea8bccf4`. Wrapped the mutations in `try/finally` so the restore runs unconditionally even if the assertion throws:

```typescript
delete (globalThis.crypto as any).randomUUID;
try {
  expect(randomUuid()).toMatch(UUID_RE);
} finally {
  if (saved) (globalThis.crypto as any).randomUUID = saved;
}
```

This is safer than inline restore which is skipped on assertion failure.

apps/web/src/components/__tests__/ScrollToTop.coverage.test.tsx
mei-artificerinnovations: The DEV `console.warn` branch requires `import.meta.env.DEV` to be truthy, which in the Vitest jsdom environment is `false` by default. Testing it would need `vi.stubEnv("DEV", "true")` plus `vi.spyOn(console, "warn")`. Skipped here to avoid env-stubbing complexity, but the branch does run (it just exits early at the `if (import.meta.env.DEV)` guard). The `eslint-disable-next-line` was needed to unblock CI — the lint rule fires regardless of the `if` condition.

packages/billing/src/components/FeatureGate.web.coverage.test.tsx
mei-artificerinnovations: The `as any` casts are intentional to reach the early-exit guards (`rawKey == null`, `rawKey === ""`) that the TypeScript types would prevent at compile time. An alternative would be `vi.mocked(useFeature).mockReturnValue(false)` and `feature=""` with a separate mock, but the null/undefined case requires bypassing types regardless. Happy to restructure if you prefer a cast-free approach.

apps/mobile/src/components/dashboard/__tests__/LucideLockIcon.test.tsx
mei-artificerinnovations: Agreed it's smoke-only. The main risk being guarded here is the SVG mock incompatibility crashing the import — `react-native-svg` needs the mock or the entire test suite errors before running. Since the source has no branches to cover (`size`/`color` are just props), a render-doesn't-throw check is the coverage floor. Would add a snapshot or prop assertion if there's a specific regression risk you're thinking about.

PR #260, opened 2026-05-18
feat(setup): auto-provision Stripe webhooks in setup:full

ZappoMan:
## Summary

- Extract `ensureStripeWebhook()` into `scripts/lib/ensure-stripe-webhook.mjs` for shared use by CI (`stripe:ensure-webhook`) and `setup:full`.
- In the stripe wizard phase, after collecting `sk_*` for a hosted Supabase tier, automatically create or update the Stripe webhook and store `*_STRIPE_WEBHOOK_SECRET` when Stripe returns a new signing secret.
- Fall back to a single masked `whsec` prompt when the endpoint already exists and the secret was never saved (Dashboard Reveal).
- Update setup prep checklist and stripe billing docs; add unit tests for the ensure lib and auto-ensure eligibility.

## Test plan

- [x] `node --test scripts/__tests__/ensure-stripe-webhook.test.mjs scripts/__tests__/setup-stripe.test.mjs`
- [ ] Run `npm run setup:full -- --from=stripe` against a fork with hosted Supabase URLs and verify webhook + `whsec` are filled for a greenfield tier
- [ ] Re-run on an existing Stripe webhook without stored `whsec` and confirm the Reveal fallback prompt appears
- [ ] Confirm CI `stripe:ensure-webhook` behavior unchanged on staging/preview deploy

### DIFF
--- a/docs/stripe-billing-setup.md
+++ b/docs/stripe-billing-setup.md
@@ -163,7 +163,7 @@ When **Stripe’s servers** must call your project (staging/production, or a sta
    `https://<PROJECT_REF>.supabase.co/functions/v1/stripe-webhook`
    - `PROJECT_REF` is in the Supabase project URL (Dashboard → Project Settings → API → Project URL).
 
-3. **Events to send** — at minimum select the types the template handler implements (you can also use “Receive all events” while testing):
+3. **Events to send** — select the same types as [`STRIPE_WEBHOOK_ENABLED_EVENTS`](../scripts/lib/ensure-stripe-webhook.mjs) (auto-ensure and CI use this list; you can also use “Receive all events” while testing):
    - `checkout.session.completed`
    - `customer.subscription.updated`
    - `customer.subscription.deleted`
@@ -171,6 +171,9 @@ When **Stripe’s servers** must call your project (staging/production, or a sta
    - `invoice.payment_failed`
    - `invoice.paid`
    - `invoice.payment_succeeded`
+   - `invoice.created`
+   - `invoice.finalized`
+   - `invoice.voided`
 
 4. After saving, open the webhook details and **Reveal** the **Signing secret** (`whsec_…`). That value is **`STRIPE_WEBHOOK_SECRET`** for the **same** Stripe mode (test vs live) as your API keys.
 
--- a/docs/stripe-billing-setup.md
+++ b/docs/stripe-billing-setup.md
@@ -12,16 +12,16 @@ This guide walks you from **zero** to a working **test-mode** Stripe integration
 
 ## Before the Stripe wizard phase
 
-If you use `npm run setup:full`, the **stripe** phase (after **supabase**, before **write** / **github**) collects GitHub Actions keys so you are not surprised by `STAGING_STRIPE_SECRET_KEY` at github sync.
+If you use `npm run setup:full`, the **stripe** phase (after **supabase**, before **write** / **github**) collects GitHub Actions keys so you are not surprised by `STAGING_STRIPE_SECRET_KEY` at github sync. For each **hosted** Supabase project (`https://<ref>.supabase.co`), the wizard uses the same logic as CI ([`scripts/lib/ensure-stripe-webhook.mjs`](../scripts/lib/ensure-stripe-webhook.mjs)) to create or update the Stripe webhook and store `*_STRIPE_WEBHOOK_SECRET` when Stripe returns a new signing secret.
 
 ### Checklist
 
-| #   | Requirement                      | Details                                                                                                                                                                  |
-| --- | -------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| 1   | **Stripe account**               | [dashboard.stripe.com](https://dashboard.stripe.com) — **Test mode** for preview + staging.                                                                              |
-| 2   | **API secret keys**              | Developers → API keys → `sk_test_…` (preview/staging) and `sk_live_…` (production when ready).                                                                           |
-| 3   | **Webhook per Supabase project** | URL `https://<PROJECT_REF>.supabase.co/functions/v1/stripe-webhook` — one endpoint per preview/staging/production project; **Reveal** signing secret `whsec_…` for each. |
-| 4   | **Supabase URLs from setup**     | Complete the **supabase** phase first so the wizard can print each webhook URL.                                                                                          |
+| #   | Requirement                      | Details                                                                                                                                                                                                                                                                 |
+| --- | -------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| 1   | **Stripe account**               | [dashboard.stripe.com](https://dashboard.stripe.com) — **Test mode** for preview + staging.                                                                                                                                                                             |
+| 2   | **API secret keys**              | Developers → API keys → `sk_test_…` (preview/staging) and `sk_live_…` (production when ready).                                                                                                                                                                          |
+| 3   | **Webhook per Supabase project** | One endpoint per preview/staging/production at `https://<PROJECT_REF>.supabase.co/functions/v1/stripe-webhook`. **Greenfield:** wizard/CI create it from `sk_*` only. **Existing endpoint:** paste `whsec_…` from Dashboard → **Reveal** if the secret was never saved. |
+| 4   | **Supabase URLs from setup**     | Complete the **supabase** phase first so the wizard can ensure webhooks against the correct URL.                                                                                                                                                                        |
 
 Skip billing in CI: answer **N** at the stripe phase or `npm run setup:full -- --skip-stripe`.
 
@@ -155,7 +155,7 @@ Seed data (when using repo `supabase/seed.sql`) includes the template product **
 
 ## 5. Stripe Dashboard — webhook endpoint (hosted Supabase)
 
-When **Stripe’s servers** must call your project (staging/production, or a stable tunnel), register the webhook in Stripe:
+When **Stripe’s servers** must call your project (staging/production, or a stable tunnel), register the webhook in Stripe. **`npm run setup:full`** (stripe phase) and deploy CI (`npm run stripe:ensure-webhook`) can create or update this endpoint for `https://<ref>.supabase.co` projects automatically; use the manual steps below if you prefer the Dashboard or need to **Reveal** an existing signing secret.
 
 1. **Developers → Webhooks → Add endpoint**.
 2. **Endpoint URL** (replace placeholders):
--- a/scripts/__tests__/ensure-stripe-webhook.test.mjs
+++ b/scripts/__tests__/ensure-stripe-webhook.test.mjs
@@ -1,0 +1,142 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+
+import {
+  ensureStripeWebhook,
+  isAutoEnsureStripeWebhookUrl,
+  MissingWebhookSecretError,
+  normalizeStripeWebhookUrl,
+  stripeSecretKeyMatchesMode,
+  STRIPE_WEBHOOK_ENABLED_EVENTS,
+} from '../lib/ensure-stripe-webhook.mjs';
+
+test('normalizeStripeWebhookUrl trims trailing slash', () => {
+  assert.equal(
+    normalizeStripeWebhookUrl(
+      'https://abcd.supabase.co/functions/v1/stripe-webhook/'
+    ),
+    'https://abcd.supabase.co/functions/v1/stripe-webhook'
+  );
+});
+
+test('isAutoEnsureStripeWebhookUrl accepts hosted supabase https webhook', () => {
+  assert.ok(
+    isAutoEnsureStripeWebhookUrl(
+      'https://abcd1234.supabase.co/functions/v1/stripe-webhook'
+    )
+  );
+});
+
+test('isAutoEnsureStripeWebhookUrl rejects local docker', () => {
+  assert.ok(
+    !isAutoEnsureStripeWebhookUrl(
+      'http://127.0.0.1:54321/functions/v1/stripe-webhook'
+    )
+  );
+});
+
+test('isAutoEnsureStripeWebhookUrl rejects custom hosts', () => {
+  assert.ok(
+    !isAutoEnsureStripeWebhookUrl(
+      'https://api.example.com/functions/v1/stripe-webhook'
+    )
+  );
+});
+
+test('stripeSecretKeyMatchesMode', () => {
+  assert.ok(stripeSecretKeyMatchesMode('sk_test_abc', 'test'));
+  assert.ok(stripeSecretKeyMatchesMode('sk_live_abc', 'live'));
+  assert.ok(!stripeSecretKeyMatchesMode('sk_live_abc', 'test'));
+});
+
+test('ensureStripeWebhook creates endpoint when missing', async () => {
+  const calls = [];
+  const mockStripe = {
+    webhookEndpoints: {
+      list: async () => ({ data: [] }),
+      create: async params => {
+        calls.push(['create', params]);
+        return { id: 'we_123', secret: 'whsec_new' };
+      },
+      update: async () => {
+        throw new Error('should not update');
+      },
+    },
+  };
+
+  const result = await ensureStripeWebhook({
+    secretKey: 'sk_test_x',
+    webhookUrl: 'https://proj.supabase.co/functions/v1/stripe-webhook',
+    description: 'test',
+    stripe: mockStripe,
+  });
+
+  assert.equal(result.signingSecret, 'whsec_new');
+  assert.equal(result.created, true);
+  assert.equal(result.endpointId, 'we_123');
+  assert.deepEqual(calls[0][1].enabled_events, STRIPE_WEBHOOK_ENABLED_EVENTS);
+});
+
+test('ensureStripeWebhook reuses existing secret', async () => {
+  const url = 'https://proj.supabase.co/functions/v1/stripe-webhook';
+  const mockStripe = {
+    webhookEndpoints: {
+      list: async () => ({
+        data: [
+          {
+            id: 'we_existing',
+            url,
+            enabled_events: [...STRIPE_WEBHOOK_ENABLED_EVENTS],
+          },
+        ],
+      }),
+      create: async () => {
+        throw new Error('should not create');
+      },
+      update: async () => {
+        throw new Error('should not update');
+      },
+    },
+  };
+
+  const result = await ensureStripeWebhook({
+    secretKey: 'sk_test_x',
+    webhookUrl: url,
+    existingWebhookSecret: 'whsec_saved',
+    stripe: mockStripe,
+  });
+
+  assert.equal(result.signingSecret, 'whsec_saved');
+  assert.equal(result.created, false);
+});
+
+test('ensureStripeWebhook throws MissingWebhookSecretError when endpoint exists without secret', async () => {
+  const url = 'https://proj.supabase.co/functions/v1/stripe-webhook';
+  const mockStripe = {
+    webhookEndpoints: {
+      list: async () => ({
+        data: [
+          {
+            id: 'we_existing',
+            url,
+            enabled_events: ['checkout.session.completed'],
+          },
+        ],
+      }),
+      update: async () => ({ id: 'we_existing' }),
+      create: async () => {
+        throw new Error('should not create');
+      },
+    },
+  };
+
+  await assert.rejects(
+    () =>
+      ensureStripeWebhook({
+        secretKey: 'sk_test_x',
+        webhookUrl: url,
+        stripe: mockStripe,
+      }),
+    MissingWebhookSecretError
+  );
+});
--- a/scripts/__tests__/ensure-stripe-webhook.test.mjs
+++ b/scripts/__tests__/ensure-stripe-webhook.test.mjs
@@ -139,6 +139,45 @@ test('ensureStripeWebhook reuses existing secret', async () => {
 
   assert.equal(result.signingSecret, 'whsec_saved');
   assert.equal(result.created, false);
+  assert.equal(result.reenabled, false);
+});
+
+test('ensureStripeWebhook re-enables disabled endpoint', async () => {
+  const url = 'https://proj.supabase.co/functions/v1/stripe-webhook';
+  const updateCalls = [];
+  const mockStripe = {
+    webhookEndpoints: {
+      list: async () => ({
+        data: [
+          {
+            id: 'we_disabled',
+            url,
+            status: 'disabled',
+            enabled_events: [...STRIPE_WEBHOOK_ENABLED_EVENTS],
+          },
+        ],
+      }),
+      update: async (id, params) => {
+        updateCalls.push([id, params]);
+        return { id };
+      },
+      create: async () => {
+        throw new Error('should not create');
+      },
+    },
+  };
+
+  const result = await ensureStripeWebhook({
+    secretKey: 'sk_test_x',
+    webhookUrl: url,
+    existingWebhookSecret: 'whsec_saved',
+    stripe: mockStripe,
+  });
+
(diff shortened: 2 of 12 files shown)


sarah-artificerinnovations (approved):
Triage: Standard / extraction + new wizard behavior; multi-file, non-runtime.

Clean PR — the extraction is the right call, the dependency injection pattern makes `ensureStripeWebhook` properly testable, and the `isAutoEnsureStripeWebhookUrl` allowlist being tight (only `https://*.supabase.co/functions/v1/stripe-webhook`) is the correct default. Docs accurately reflect the new flow.

Three nits:

---

**n1 — dead `readSecret` parameter in `ensureTierStripeWebhookSecret`**

In `scripts/lib/setup-stripe.mjs`, `readSecret` appears in the JSDoc and the call site in `collectStripeEnvKeys`, but the function body only destructures `{ acc, tier, webhookUrl, logInfo, logWarn, applySecret }` — `readSecret` is never used inside. Looks like a vestige of an earlier design where the function would handle the fallback prompt itself. Remove it from the JSDoc type and the call site; the function's job is purely auto-ensure, and the prompt lives in the caller, which is correct.

---

**n2 — missing test for the `eventsUpdated: true` path**

`ensure-stripe-webhook.test.mjs` covers: create (no existing endpoint), reuse-existing-secret, and `MissingWebhookSecretError`. The missing case: endpoint exists, events are stale, signing secret IS provided — should update events and return `{ eventsUpdated: true, created: false }`. Worth a test since it's a distinct code path and the update side-effect before a throw (n3 below) is where bugs tend to hide.

---

**n3 — `ensureTierStripeWebhookSecret` has no unit tests**

It's the most complex new function — six distinct branches (key already set, no secretKey, non-auto URL, mode mismatch, success create/update/reuse, `MissingWebhookSecretError`, generic error). The `applySecret` callback + `acc` mutation pattern is subtler than it looks. `ensureStripeWebhook` itself is well tested, so the risk is low, but a targeted test for the wrapper would make the mode-mismatch warn and the error-fallback-to-manual paths explicit.

---

n1 is cosmetic. n2 and n3 are worth addressing before merge, but won't block given the blast radius (a bug here means the developer pastes `whsec_` manually — annoying, not catastrophic).

Approve with those.

scripts/__tests__/ensure-stripe-webhook.test.mjs
```diff
@@ -0,0 +1,173 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+
+import {
+  ensureStripeWebhook,
+  findStripeWebhookEndpointByUrl,
+  isAutoEnsureStripeWebhookUrl,
+  MissingWebhookSecretError,
+  normalizeStripeWebhookUrl,
```
mei-artificerinnovations, line 9: `findStripeWebhookEndpointByUrl` is imported here but is not exported from `../lib/ensure-stripe-webhook.mjs`. This will throw at import time and fail all tests in the file. The pagination logic it tests also does not exist in the lib — `ensureStripeWebhook` only fetches the first page (limit: 100).

scripts/lib/ensure-stripe-webhook.mjs
```diff
@@ -0,0 +1,183 @@
...
+export async function findStripeWebhookEndpointByUrl(stripe, webhookUrl) {
+  const normalized = normalizeStripeWebhookUrl(webhookUrl);
+  if (!normalized) return null;
+
+  let startingAfter;
+  for (;;) {
+    const list = await stripe.webhookEndpoints.list({
+      limit: 100,
+      ...(startingAfter ? { starting_after: startingAfter } : {}),
+    });
```
mei-artificerinnovations, line 93: Single `list({ limit: 100 })` without pagination. If the account has >100 webhook endpoints, `found` will be `undefined` even when the endpoint exists, and `create` will produce a duplicate. Use a `findStripeWebhookEndpointByUrl` paginator here.

mei-artificerinnovations (changes requested):
## Review

Overall this is well-structured — extracting `ensureStripeWebhook` into a shared lib is the right call, and the fallback-to-manual-prompt flow in `collectStripeEnvKeys` handles the edge cases cleanly. One blocking bug, one design note.

### Blocking — `findStripeWebhookEndpointByUrl` is imported but not exported

`scripts/__tests__/ensure-stripe-webhook.test.mjs` imports `findStripeWebhookEndpointByUrl` from the lib, but that function is not exported (or defined) in `scripts/lib/ensure-stripe-webhook.mjs`. The test will fail at import time and every test in the file will error.

This also points to a real pagination gap: `ensureStripeWebhook` calls `stripe.webhookEndpoints.list({ limit: 100 })` once and stops. Accounts with >100 endpoints would get a duplicate created instead of finding the existing one. The pagination test was written for a helper that was never added.

Fix: export a `findStripeWebhookEndpointByUrl(stripe, url)` helper that paginates (`has_more` loop), and have `ensureStripeWebhook` call it.

### Design note — `ensureTierStripeWebhookSecret` error suppression

The general catch in `ensureTierStripeWebhookSecret` logs a warn and returns `false`, which causes the caller to fall back to a manual `whsec_...` prompt. That's the right UX, but it means Stripe API errors (rate limits, network failures, invalid key format) are silent beyond the single warn line. If you expect this to run in CI without a human watching, consider logging `e.stack` or a structured error to help debug later.

scripts/lib/ensure-stripe-webhook.mjs
```diff
@@ -0,0 +1,159 @@
...
+  if (!secretKey || !url) {
+    throw new Error('ensureStripeWebhook requires secretKey and webhookUrl');
+  }
+
+  const stripe =
+    opts.stripe ?? new Stripe(secretKey, { apiVersion: '2023-10-16' });
+  const description =
+    opts.description?.trim() || 'BeakerStack (stripe-webhook)';
+
+  const list = await stripe.webhookEndpoints.list({ limit: 100 });
```
diego-artificerinnovations: nit: `list({ limit: 100 })` fetches at most one page — if the Stripe account has more than 100 webhook endpoints the target URL might be absent and `ensureStripeWebhook` would create a duplicate instead of updating the existing one.

Stripe's own Dashboard caps accounts at ~16 webhook endpoints, so this won't trigger in practice. But since this function is now shared (CI + setup:full), a one-line comment documents the assumption:

```js
// Stripe caps webhook endpoints per account well below 100; a single page is sufficient.
const list = await stripe.webhookEndpoints.list({ limit: 100 });
```

Alternatively, `stripe.webhookEndpoints.list({ limit: 100 }).autoPagingEach(...)` if you ever want to be safe against the cap changing.

scripts/__tests__/ensure-stripe-webhook.test.mjs
```diff
@@ -0,0 +1,142 @@
...
+  await assert.rejects(
+    () =>
+      ensureStripeWebhook({
+        secretKey: 'sk_test_x',
+        webhookUrl: url,
+        stripe: mockStripe,
+      }),
+    MissingWebhookSecretError
+  );
+});
```
diego-artificerinnovations, line 248: nit: `ensureTierStripeWebhookSecret` (the orchestration layer in `setup-stripe.mjs`) has no unit tests. Its helpers are individually covered, but the glue logic — mode check, `MissingWebhookSecretError` fallback, non-auto-ensure early exit — is untested. Since the function takes all its deps via a context object, stubs are trivial:

```js
// in setup-stripe.test.mjs (or a new ensure-tier.test.mjs)
import { ensureTierStripeWebhookSecret, STRIPE_SETUP_TIERS } from '../lib/setup-stripe.mjs';

function makeCtx(overrides = {}) {
  const tier = STRIPE_SETUP_TIERS.find(t => t.id === 'preview');
  return {
    acc: { PREVIEW_STRIPE_SECRET_KEY: 'sk_test_abc' },
    tier,
    webhookUrl: 'https://proj.supabase.co/functions/v1/stripe-webhook',
    logInfo: () => {},
    logWarn: () => {},
    applySecret: async () => {},
    readSecret: async () => '',
    ...overrides,
  };
}

test('ensureTierStripeWebhookSecret: wrong Stripe mode → returns false', async () => {
  // sk_live_ key on a test-mode tier should warn and short-circuit
  const ctx = makeCtx({ acc: { PREVIEW_STRIPE_SECRET_KEY: 'sk_live_bad' } });
  const warned = [];
  ctx.logWarn = s => warned.push(s);
  const result = await ensureTierStripeWebhookSecret(ctx);
  assert.equal(result, false);
  assert.ok(warned.some(w => /sk_test/.test(w)));
});

test('ensureTierStripeWebhookSecret: non-auto-ensure URL → returns false', async () => {
  const ctx = makeCtx({ webhookUrl: 'http://127.0.0.1:54321/functions/v1/stripe-webhook' });
  const result = await ensureTierStripeWebhookSecret(ctx);
  assert.equal(result, false);
});
```

The `MissingWebhookSecretError` path and the happy-path (create → `applySecret` called) are also worth a case each, but the above two are the cheapest ones to miss.

diego-artificerinnovations (approved):
Clean extraction and a solid UX improvement. The `ensureStripeWebhook` interface is well-factored — injectable `stripe` client for testing, `MissingWebhookSecretError` for the Reveal fallback path, `normalizeStripeWebhookUrl` separating URL normalization concerns cleanly. Removing the duplicated event list and Stripe SDK usage from the CI script is a real win. Two nits below.

scripts/__tests__/ensure-stripe-webhook.test.mjs
```diff
@@ -0,0 +1,173 @@
...
+    stripe: mockStripe,
+  });
+
+  assert.equal(result.signingSecret, 'whsec_new');
+  assert.equal(result.created, true);
+  assert.equal(result.endpointId, 'we_123');
+  assert.deepEqual(calls[0][1].enabled_events, STRIPE_WEBHOOK_ENABLED_EVENTS);
+});
+
+test('ensureStripeWebhook reuses existing secret', async () => {
```
rahul-artificerinnovations, line 111: The "reuses existing secret" test covers events-match + secret-provided → `eventsUpdated: false`, and the `MissingWebhookSecretError` test covers events-mismatch + no-secret. The missing case is **events-mismatch + secret-provided → `update` called + `eventsUpdated: true`**. This is the "webhook already exists, but event list grew since it was created" path — the most likely real-world update scenario. Worth a quick test:
```js
test('ensureStripeWebhook updates events and returns eventsUpdated: true', async () => {
  const url = 'https://proj.supabase.co/functions/v1/stripe-webhook';
  const updateCalls = [];
  const mockStripe = {
    webhookEndpoints: {
      list: async () => ({
        data: [{ id: 'we_existing', url, enabled_events: ['checkout.session.completed'] }],
      }),
      update: async (id, params) => { updateCalls.push([id, params]); return { id }; },
      create: async () => { throw new Error('should not create'); },
    },
  };
  const result = await ensureStripeWebhook({
    secretKey: 'sk_test_x', webhookUrl: url,
    existingWebhookSecret: 'whsec_saved', stripe: mockStripe,
  });
  assert.equal(result.eventsUpdated, true);
  assert.equal(result.signingSecret, 'whsec_saved');
  assert.equal(updateCalls.length, 1);
});
```

scripts/lib/setup-stripe.mjs
```diff
@@ -100,6 +107,93 @@ export function tierStripeKeysPresent(acc, tier) {
...
+ *   acc: Record<string, string>;
+ *   tier: StripeSetupTier;
+ *   webhookUrl: string;
+ *   logInfo: (s: string) => void;
+ *   logWarn: (s: string) => void;
+ *   applySecret: (raw: string, primaryKey: string) => Promise<void>;
+ * }} ctx
+ * @returns {Promise<boolean>} true if webhook secret is set when done
+ */
+export async function ensureTierStripeWebhookSecret(ctx) {
```
rahul-artificerinnovations, line 138: `ensureTierStripeWebhookSecret` has zero unit tests. It has at least six distinct code paths: already-set early-return, no-secretKey, non-hosted URL, mode-mismatch warn, `MissingWebhookSecretError` fallback, and the happy path. The individual building blocks (`ensureStripeWebhook`, `stripeSecretKeyMatchesMode`, `isAutoEnsureStripeWebhookUrl`) are all tested, so the risk is low — but the orchestration logic (which path is taken, what gets logged, what `applySecret` receives) isn't covered at all. A few cases would add confidence: at minimum the already-set early-return, the mode-mismatch warn-and-return, and the happy-path `applySecret` call.

rahul-artificerinnovations (approved):
Clean extraction. Moving the webhook-ensure logic into a shared lib was the right call — CI and the setup wizard now stay in sync on event lists and matching semantics without duplication. The `findStripeWebhookEndpointByUrl` pagination fix (caught by the previous reviewer) is solid, the injection point for `stripe` keeps tests fully synchronous, and `MissingWebhookSecretError` as a named class makes the catch branches in both callers clean and type-safe. Two minor test coverage gaps below, neither is a blocker.

scripts/__tests__/ensure-stripe-webhook.test.mjs
```diff
@@ -0,0 +1,142 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 88ad1eb: added both stripe test files to `test:unit:scripts` in package.json (82 script tests pass).

scripts/__tests__/setup-stripe.test.mjs
```diff
@@ -68,6 +70,27 @@ test('setupStripeKeysDeferred', () => {
...
 });
 
+test('hosted supabase webhook URL is eligible for auto-ensure', () => {
+  const url = supabaseStripeWebhookUrl('https://abcd1234.supabase.co');
+  assert.ok(isAutoEnsureStripeWebhookUrl(url));
+});
+
+test('local supabase webhook URL is not auto-ensured', () => {
+  const url = supabaseStripeWebhookUrl('http://127.0.0.1:54321');
+  assert.ok(!isAutoEnsureStripeWebhookUrl(url));
```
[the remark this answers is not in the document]
ZappoMan: Same fix — `setup-stripe.test.mjs` is now included in `test:unit:scripts`.

scripts/lib/ensure-stripe-webhook.mjs
```diff
@@ -0,0 +1,159 @@
...
+  const found = list.data.find(e => normalizeStripeWebhookUrl(e.url) === url);
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 88ad1eb: exported `findStripeWebhookEndpointByUrl` with pagination; `ensureStripeWebhook` uses it instead of a single `list({ limit: 100 })`.

scripts/lib/setup-stripe.mjs
```diff
@@ -100,6 +107,94 @@ export function tierStripeKeysPresent(acc, tier) {
...
+
+/**
+ * @param {{
+ *   acc: Record<string, string>;
+ *   tier: StripeSetupTier;
+ *   webhookUrl: string;
+ *   logInfo: (s: string) => void;
+ *   logWarn: (s: string) => void;
+ *   applySecret: (raw: string, primaryKey: string) => Promise<void>;
+ *   readSecret: (prompt: string) => Promise<string>;
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 88ad1eb: removed `readSecret` from JSDoc and call site — manual whsec prompt remains in `collectStripeEnvKeys`.

scripts/__tests__/ensure-stripe-webhook.test.mjs
ZappoMan: Addressed in 88ad1eb: `findStripeWebhookEndpointByUrl` is exported and tested (pagination mock). `ensureStripeWebhook` calls the same helper.

scripts/lib/ensure-stripe-webhook.mjs
ZappoMan: Same — paginate all pages before create to avoid duplicate endpoints.

scripts/lib/ensure-stripe-webhook.mjs
ZappoMan: Implemented full pagination via `findStripeWebhookEndpointByUrl` rather than documenting the 100-endpoint cap.

scripts/__tests__/ensure-stripe-webhook.test.mjs
ZappoMan: Added in a98839b: `ensureTierStripeWebhookSecret` tests (already-set, mode mismatch, non-hosted URL, MissingWebhookSecretError, happy-path applySecret) with injectable `ctx.ensureStripeWebhook`.

scripts/__tests__/ensure-stripe-webhook.test.mjs
ZappoMan: Added in a98839b — `ensureStripeWebhook updates events and returns eventsUpdated: true` per your suggested test.

scripts/lib/setup-stripe.mjs
ZappoMan: Added in a98839b — same `ensureTierStripeWebhookSecret` orchestration coverage (five cases).

ZappoMan:
Thanks for the reviews — pushed two follow-up commits addressing all inline feedback:

- **88ad1eb** — Wire stripe tests into `test:unit:scripts`; export `findStripeWebhookEndpointByUrl` with pagination; remove dead `readSecret` from `ensureTierStripeWebhookSecret`.
- **a98839b** — Add `eventsUpdated: true` test; add `ensureTierStripeWebhookSecret` orchestration tests (injectable `ctx.ensureStripeWebhook`); log stack on unexpected Stripe errors in the wizard path.

All 10 review threads are replied to and resolved. `npm run test:unit:scripts` — 82 passed.

@mei-artificerinnovations — the blocking import/pagination issue should be cleared; please take another look when CI finishes.

@sarah-artificerinnovations @diego-artificerinnovations @rahul-artificerinnovations — n1/n2/n3 and the test gaps should be covered in a98839b.

scripts/lib/ensure-stripe-webhook.mjs
```diff
@@ -0,0 +1,183 @@
...
+      sorted.length !== current.length ||
+      sorted.some((ev, i) => ev !== current[i]);
+
+    let eventsUpdated = false;
+    if (needsUpdate) {
+      await stripe.webhookEndpoints.update(found.id, {
+        enabled_events: STRIPE_WEBHOOK_ENABLED_EVENTS,
+        description,
+      });
+      eventsUpdated = true;
```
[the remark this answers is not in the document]
ZappoMan: Fixed: when `found.status === "disabled"`, `ensureStripeWebhook` calls `webhookEndpoints.update` with `disabled: false` (alongside event sync when needed). Added test `ensureStripeWebhook re-enables disabled endpoint`.

docs/stripe-billing-setup.md
```diff
@@ -155,7 +155,7 @@ Seed data (when using repo `supabase/seed.sql`) includes the template product **
 
 ## 5. Stripe Dashboard — webhook endpoint (hosted Supabase)
 
-When **Stripe’s servers** must call your project (staging/production, or a stable tunnel), register the webhook in Stripe:
+When **Stripe’s servers** must call your project (staging/production, or a stable tunnel), register the webhook in Stripe. **`npm run setup:full`** (stripe phase) and deploy CI (`npm run stripe:ensure-webhook`) can create or update this endpoint for `https://<ref>.supabase.co` projects automatically; use the manual steps below if you prefer the Dashboard or need to **Reveal** an existing signing secret.
```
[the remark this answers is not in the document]
ZappoMan: Fixed: manual Dashboard event list in §5 now includes `invoice.created`, `invoice.finalized`, and `invoice.voided`, and points at `STRIPE_WEBHOOK_ENABLED_EVENTS` in `scripts/lib/ensure-stripe-webhook.mjs` so manual and auto-ensure stay aligned.

sarah-artificerinnovations (approved):
All three nits addressed.

**n1:** `readSecret` removed from `ensureTierStripeWebhookSecret`'s JSDoc and call site. The injectable `ensureStripeWebhook` parameter (with module default) is the right replacement — makes the function properly testable without monkey-patching.

**n2:** `'ensureStripeWebhook updates events and returns eventsUpdated: true'` covers the stale-events path, asserts `eventsUpdated: true`, checks that `update()` was called exactly once, and verifies the existing secret is passed through. Good.

**n3:** Five new tests for `ensureTierStripeWebhookSecret` covering early-return, mode mismatch, non-auto URL, `MissingWebhookSecretError` fallback, and happy path. The `makeEnsureTierCtx` helper keeps them tight.

Two things Brad added that weren't asked for and are both correct:

- **Pagination on `findStripeWebhookEndpointByUrl`** — extracting it into its own exported function with `starting_after` cursor pagination fixes a real (if unlikely) edge case. The dedicated pagination test is exactly right.
- **`reenabled` flag + re-enable logic** — handles `status: 'disabled'` endpoints gracefully rather than treating them as "not found" and trying to create a duplicate. Good catch.

Approving.

rahul-artificerinnovations (approved):
Both observations addressed, and the re-enable fix in the final commit is a genuine bonus catch.

On the test coverage: `eventsUpdated: true` test is exactly what I suggested. The `ensureTierStripeWebhookSecret` tests are comprehensive — the `ensureStripeWebhook` injection via `ctx.ensureStripeWebhook ?? ensureStripeWebhook` is the right approach for testing the orchestration layer without hitting the network, and the mode-mismatch warn test confirms the warning text. Stack trace logging on unknown errors is a nice touch.

On the re-enable fix: the disabled-endpoint case was a real gap — a wizard or CI run finding an existing `disabled` endpoint would silently reuse the secret and return success while Stripe wasn't actually delivering events. The combined `needsEventUpdate || needsReenable` path is clean, and the behavior for disabled-without-secret is correct: re-enable fires, then `MissingWebhookSecretError` is thrown, so the endpoint becomes active the moment the user supplies the secret on the next run.

Docs alignment (adding the three missing invoice events to the manual setup table) is a good catch too. Good to merge.

mei-artificerinnovations (approved):
Both issues from my first pass are addressed.

**Blocking (fixed):** `findStripeWebhookEndpointByUrl` is now exported from the lib and uses a proper pagination loop (`has_more` + `starting_after`). `ensureStripeWebhook` delegates to it, so accounts with >100 endpoints will no longer get duplicate webhooks. The test import matches.

**Bonus:** re-enabling disabled endpoints (`needsReenable` + `disabled: false`) is a good addition — exactly the kind of thing that would silently break in prod if not handled. The `reenabled` flag is threaded through to `setup-stripe.mjs` and the log messages cover all combinations. Tests for the new paths are included.

**Design note (acknowledged):** the general catch in `ensureTierStripeWebhookSecret` is still a single warn line. That is acceptable for the interactive setup use case here — I just wanted it on the record for if this ever runs unattended in CI.

diego-artificerinnovations (approved):
All nits addressed — re-approving.

**fix/address-PR-review (e2d1dc7):** Pagination done right: `findStripeWebhookEndpointByUrl` extracted with cursor-based `starting_after` loop, tested with a two-page mock that verifies exactly two `list` calls. Unused `readSecret` dropped from `ensureTierStripeWebhookSecret`. Both test files wired into `test:unit:scripts`. Clean.

**test/ensureTier (5c25c8a):** All four paths covered — already-set early-exit, wrong-mode warning, non-auto-ensure URL, `MissingWebhookSecretError` fallback, and happy-path `applySecret` call. Injectable `ensureStripeWebhook` in ctx is the right pattern here (avoids mocking module imports). `eventsUpdated: true` case added. Unexpected-error path now logs the stack too. Good.

**fix/re-enable (23c011f8):** Bonus and the right call — disabled endpoints returning a 400 from Stripe is a real foot-gun in multi-app setups where someone toggled a webhook off in the Dashboard. The `reenabled` flag threads through both callers consistently, the re-enable test verifies `disabled: false` on the update params, and the docs now reference `STRIPE_WEBHOOK_ENABLED_EVENTS` directly so the manual list and code can't drift apart again.

PR #261, opened 2026-05-18
feat(admin): operator panel, RPCs, and grant CLI

ZappoMan:
## Summary

- Adds **Admin v1**: `admin_users` / `admin_audit_log` tables, `admin_is_admin` and list/get RPCs (with audit writes), and pgTAP coverage.
- Introduces **`@beakerstack/admin`** (route guard, layout primitives, `useIsAdmin`) and wires lazy **`/admin/*`** routes in the web app (dashboard + paginated users list with detail drawer).
- Adds **service-role CLI** (`npm run admin:grant` / `admin:revoke`) and [docs/guides/admin-setup.md](docs/guides/admin-setup.md); fixes PostgREST **405** on list/get RPCs by marking them `VOLATILE`, and fixes `useIsAdmin` false-negative redirect race.

## Test plan

- [ ] `supabase migration up` (includes `20260518120000_admin_rpc_volatile.sql`)
- [ ] `npm run admin:grant -- you@example.com` then sign in and open `/admin` and `/admin/users`
- [ ] Confirm non-admin users land on `/not-authorized`
- [ ] `npm run test:unit:admin` and `npm run test:db` (admin pgTAP)
- [ ] `npm run lint` and `npm run type-check`


scripts/grant-admin.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  let url: string;
+  let serviceKey: string;
+  try {
+    ({ url, serviceKey } = loadSupabaseServiceEnv());
+  } catch (e) {
+    console.error(e instanceof Error ? e.message : e);
+    process.exit(1);
+  }
```
sarah-artificerinnovations, line 55: **M1 — pagination bug: `listUsers()` only fetches page 1 (50 users by default)**

`supabase.auth.admin.listUsers()` returns the first page of users — typically 50. Any project with more than 50 auth users will silently fail to find users on page 2+, and the CLI exits with "No auth user found" even though the account exists.

Fix: paginate until the target email is found (or use a direct user-lookup if Supabase exposes one):

```ts
let page = 1;
let match;
while (!match) {
  const { data, error } = await supabase.auth.admin.listUsers({ page, perPage: 100 });
  if (error || !data.users.length) break;
  match = data.users.find(u => u.email?.toLowerCase() === email.toLowerCase());
  if (!data.nextPage) break;
  page++;
}
```

Same issue exists in `scripts/revoke-admin.ts` at the same pattern.

scripts/grant-admin.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  const { data: users, error: listErr } = await supabase.auth.admin.listUsers();
+  if (listErr) {
+    console.error('Failed to list users:', listErr.message);
+    process.exit(1);
+  }
+
+  const match = users.users.find(
+    u => u.email?.toLowerCase() === email.toLowerCase()
+  );
```
sarah-artificerinnovations: **M2 — `granted_by` is always `null`**

The `admin_users.granted_by` column (FK → `auth.users`) exists precisely so you can later answer "who granted admin to this user?" — but the CLI never sets it, leaving every CLI-generated grant with `granted_by = NULL`.

The service role doesn't have a `user_id` in `auth.users`, so you can't use it as a FK. Options:
- Drop `granted_by` if it will never be populated, or change its type to `text` to hold a label like `'cli:you@example.com'` or an env identifier.
- Add a `granted_by_label text` column alongside the FK, and populate it with something like `'cli:'` + the operator email from the env/arg.

As-is, the audit trail for grants is incomplete. The SECURITY.md should at least note this: "CLI grants record `granted_by = null`; check deployment logs or use a dedicated service account for attribution."

supabase/migrations/20260518000000_admin_v1.sql
```diff
@@ -0,0 +1,340 @@
...
+    FROM public.user_profiles p
+    WHERE p.user_id = p_user_id;
+
+    SELECT to_jsonb(s.*) INTO v_subscription
+    FROM public.billing_subscriptions s
+    WHERE s.user_id = p_user_id AND s.product_id = p_product_id;
+
+    IF v_subscription IS NOT NULL THEN
+        SELECT to_jsonb(pl.*) INTO v_plan
+        FROM public.billing_plans pl
```
sarah-artificerinnovations, line 303: **M3 — unbounded `p_search` can cause a full `auth.users` scan**

`p_search` has no length cap. A search term of a few thousand characters triggers an expensive full ILIKE scan on `auth.users.email` with no index (indexes don't help for leading-`%` ILIKE patterns). Even for an authenticated admin, this is a DoS vector against the DB.

Suggest capping at a reasonable length at the top of the function:

```sql
v_limit integer := LEAST(GREATEST(COALESCE(p_limit, 25), 1), 100);
v_offset integer := GREATEST(COALESCE(p_offset, 0), 0);
-- add:
p_search := CASE WHEN length(p_search) > 200 THEN NULL ELSE p_search END;
```

(Non-blocking if your admin user count is small, but worth fixing before this scales.)

supabase/migrations/20260518000000_admin_v1.sql
```diff
@@ -0,0 +1,340 @@
...
+    v_usage_aggregates jsonb;
+    v_usage_events jsonb;
+    v_invoices jsonb;
+BEGIN
+    IF v_uid IS NULL OR NOT public.admin_is_admin() THEN
+        RETURN jsonb_build_object('error', 'not_found');
+    END IF;
+
+    IF p_user_id IS NULL THEN
+        RETURN jsonb_build_object('error', 'not_found');
```
sarah-artificerinnovations, line 267: **nit — search count query duplicates the WHERE clause**

The count query and the data query have identical `WHERE` clauses but run separately:

```sql
SELECT count(*)::bigint INTO v_total
FROM auth.users u
LEFT JOIN public.user_profiles p ON p.user_id = u.id
WHERE p_search IS NULL
   OR u.email ILIKE '%' || p_search || '%';
```

This scans the table twice. For v1 user counts this is fine, but it's worth noting for when the user list grows. A single CTE with a window `count(*) OVER ()` would do it in one pass.

packages/admin/src/components/AdminRoute.web.tsx
```diff
@@ -0,0 +1,54 @@
...
+  if (authLoading || adminLoading) {
+    return (
+      <div className='min-h-screen flex flex-col items-center justify-center bg-gray-50 gap-4'>
+        <div className='inline-block animate-spin rounded-full h-8 w-8 border-b-2 border-indigo-600' />
+        <p className='text-sm text-gray-600'>Loading…</p>
+      </div>
+    );
+  }
+
+  if (!userId) {
```
sarah-artificerinnovations, line 62: **nit/UX — RPC error redirects admin to `/not-authorized` with no recovery path**

If `checkIsAdmin()` throws (transient network error, DB hiccup), `useIsAdmin` sets `isAdmin = false` with `loading = false`. `AdminRoute` then redirects to `/not-authorized` — a legitimate admin gets kicked out silently.

Fail-closed is the right security default, but the UX is bad. Consider inspecting `error` from `useIsAdmin` in `AdminRoute` and rendering an error + retry button instead of a hard redirect when the check itself failed. The `error` is already returned from the hook.

packages/admin/SECURITY.md
```diff
@@ -0,0 +1,29 @@
+# Admin panel security
+
+## Where admin status is stored
+
+`public.admin_users` maps `user_id` → grant metadata. Active admins have `revoked_at IS NULL`. The table has RLS enabled with **no** policies for `anon` or `authenticated` — clients cannot read or modify it.
+
+## How admin status is checked
+
+- **Client hint:** `admin_is_admin()` RPC (via `useIsAdmin`). Used for UI only.
```
sarah-artificerinnovations, line 9: **M4 — stale `isAdmin` cache not documented**

`useIsAdmin` runs the `admin_is_admin()` RPC once on mount and caches the result. If an admin grant is revoked while the tab is open, the route guard won't redirect until the component unmounts and remounts (e.g., next navigation that re-initialises `AdminRoute`). During this window the UI is accessible, though the underlying data RPCs (`admin_list_users`, `admin_get_user`) will return `not_found` because they re-check server-side on every call.

This means: revocation takes effect immediately on data access, but the panel UI stays visible until the next remount. Worth documenting explicitly here so adopters don't assume revocation has instant UI effect.

Consider adding a note like: "Admin revocation takes effect immediately on all data RPCs. The client-side route guard may display the panel UI until the next page navigation if the tab was open when revocation occurred."

supabase/migrations/20260518000000_admin_v1.sql
```diff
@@ -0,0 +1,340 @@
...
+    SELECT COALESCE(jsonb_agg(to_jsonb(a.*) ORDER BY a.event_type), '[]'::jsonb)
+    INTO v_usage_aggregates
+    FROM public.billing_usage_aggregates a
+    WHERE a.user_id = p_user_id
+      AND a.product_id = p_product_id
+      AND a.period_start <= now()
+      AND a.period_end > now();
+
+    SELECT COALESCE(jsonb_agg(to_jsonb(e.*) ORDER BY e.created_at DESC), '[]'::jsonb)
+    INTO v_usage_events
```
sarah-artificerinnovations, line 316: **nit — `admin_get_user` returns `to_jsonb(s.*)` for subscriptions and invoices**

`to_jsonb(s.*)` returns every column from `billing_subscriptions`, including `stripe_customer_id`, `stripe_subscription_id`, and any other Stripe identifiers. Same for `to_jsonb(i.*)` on invoices. The UI only renders a few of these fields, but the full payload (including Stripe IDs) is in the API response and visible in browser dev tools.

Stripe customer and subscription IDs are sensitive — a compromised admin session exposes them for every viewed user. Consider projecting only the columns the UI actually needs rather than `SELECT *`. If the admin panel may eventually display more fields, at minimum note in SECURITY.md that admin access includes Stripe identifiers.

supabase/tests/admin.test.sql
```diff
@@ -0,0 +1,148 @@
...
+RESET ROLE;
+
+SELECT ok(
+  (SELECT count(*)::int FROM public.admin_audit_log
+   WHERE actor_user_id = 'a1000000-0000-0000-0000-000000000002') >= 1,
+  'admin list action wrote audit log row'
+);
+
+SELECT * FROM finish();
+ROLLBACK;
```
sarah-artificerinnovations, line 231: **nit — pgTAP coverage gaps**

Three missing cases worth adding:
1. **`anon` role is denied**: current tests only verify non-admin `authenticated` is denied; an `anon` caller should also get `not_found` (or at least fail the RPC call).
2. **`admin_record_audit_event` from non-admin is a no-op**: verifies the silent-return path actually inserts nothing.
3. **Revoked admin is denied**: insert a row with `revoked_at = now()`, then verify `admin_is_admin()` returns `false` and RPCs return `not_found`. This is the most important revocation test and it's missing.

sarah-artificerinnovations (changes requested):
Triage: Heavy / auth + admin surface + DB schema.

The architecture is sound: RLS with no client policies, `SECURITY DEFINER` RPCs re-checking admin status on every call, service-role-only grant/revoke. The route guard correctly fails closed. Inline comments cover the issues; summary below.

---

**M1 (blocking) — `grant-admin.ts` / `revoke-admin.ts` only read the first page of users**

`supabase.auth.admin.listUsers()` returns ~50 users per page by default. Any project with more users will silently report "No auth user found" for users past page 1. This is the primary CLI use case and it's broken at scale. See inline comment on `scripts/grant-admin.ts:42` for a fix.

**M2 (blocking) — `granted_by` column always `null` from CLI**

The column exists for audit attribution but is never populated. Either populate it (with a label column if a user FK doesn't fit the service role model) or document the gap. See inline comment on `scripts/grant-admin.ts:56`.

**M3 (non-blocking) — unbounded `p_search` in `admin_list_users`**

No server-side length cap on the search parameter. A long search term forces a full ILIKE table scan. Cap it at ~200 chars at the top of the function. See inline comment on the migration.

**M4 (non-blocking) — stale route guard cache not documented**

Revocation takes effect immediately on the data RPCs but the client-side route guard stays open until remount. The security consequence is minor (data access is blocked server-side), but adopters need to know this. See inline comment on `packages/admin/SECURITY.md`.

---

Non-blocking nits:
- Count + data query scans the table twice — CTE with window count would fix it
- `admin_get_user` returns `to_jsonb(s.*)` which includes Stripe IDs; consider projecting only needed columns
- `AdminRoute` hard-redirects on RPC error with no retry path
- pgTAP is missing: revoked-admin test, anon-role denial, `admin_record_audit_event` no-op test

Requesting changes on M1 + M2. The rest are worth addressing but won't block.

scripts/grant-admin.ts
```diff
@@ -0,0 +1,82 @@
...
+  if (!match) {
+    console.error(`No auth user found with email: ${email}`);
+    process.exit(1);
+  }
+
+  const { error: upsertErr } = await supabase.from('admin_users').upsert(
+    {
+      user_id: match.id,
+      revoked_at: null,
+      granted_at: new Date().toISOString(),
```
diego-artificerinnovations, line 101: security nit: `granted_by` is not set — CLI grants are unattributed in `admin_users`. When a team member runs `admin:grant`, there's no record of *who* ran it.

The `admin_audit_log` only tracks in-app admin actions; the CLI bypasses it entirely. Add a `granted_by` arg or write an audit row directly:

```ts
// Option A — accept operator email/id arg:
// npm run admin:grant -- target@example.com --granted-by=operator@example.com

// Option B — write to admin_audit_log via service role after upsert:
await supabase.from('admin_audit_log').insert({
  actor_user_id: operatorId, // pass as --granted-by or require env
  action: 'admin.grant',
  target_type: 'user',
  target_id: match.id,
  details: { email: match.email },
});
```

Without attribution, a future security investigation can't distinguish "granted by Alice" from "granted by automated infra" from "granted by an attacker who had service role access temporarily."

scripts/revoke-admin.ts
```diff
@@ -0,0 +1,78 @@
...
+    .update({ revoked_at: new Date().toISOString() })
+    .eq('user_id', match.id)
+    .is('revoked_at', null);
+
+  if (updateErr) {
+    console.error('Failed to revoke admin:', updateErr.message);
+    process.exit(1);
+  }
+
+  console.log(`Revoked admin from ${email} (${match.id})`);
```
diego-artificerinnovations, line 83: security nit: `console.log("Revoked admin from …")` fires unconditionally even when 0 rows were updated — i.e. when the user was never an admin, or was already revoked. Supabase `.update()` does not error on 0-row matches.

This means the CLI silently succeeds and logs a false confirmation. An operator running `admin:revoke` against a user who already had their access revoked (or who was never an admin) gets no signal that nothing happened.

Fix: check the `count` header, or re-select after the update:

```ts
const { error: updateErr, count } = await supabase
  .from('admin_users')
  .update({ revoked_at: new Date().toISOString() })
  .eq('user_id', match.id)
  .is('revoked_at', null)
  .select('user_id', { count: 'exact', head: true });

if (updateErr) { ... }
if (!count) {
  console.error(`No active admin record found for ${email} — nothing changed`);
  process.exit(1);
}
```

supabase/migrations/20260518000000_admin_v1.sql
```diff
@@ -0,0 +1,340 @@
...
+
+    PERFORM public._admin_insert_audit(
+        v_uid,
+        'admin.users.view',
+        'user',
+        p_user_id::text,
+        jsonb_build_object('product_id', p_product_id)
+    );
+
+    SELECT to_jsonb(p.*) INTO v_profile
```
diego-artificerinnovations, line 293: security nit: `to_jsonb(p.*)` returns every column of `user_profiles` to admins — including any sensitive columns added in the future. Same pattern for `to_jsonb(s.*)` (billing_subscriptions, line 282), `to_jsonb(a.*)` (usage_aggregates, line 292), and `to_jsonb(e.*)` (usage_events, line 300).

For `user_profiles` and `billing_subscriptions` in particular, this is a schema-coupling risk: a column added for internal use (e.g. a hashed token, an internal state machine flag) will be silently returned to every admin session without any code change.

The list RPC (`admin_list_users`) already uses an explicit `jsonb_build_object` allowlist — worth doing the same in `admin_get_user` for the profile and subscription blocks:

```sql
-- instead of to_jsonb(p.*):
SELECT jsonb_build_object(
    'display_name', p.display_name,
    'username', p.username,
    'avatar_url', p.avatar_url,
    'updated_at', p.updated_at
) INTO v_profile
FROM public.user_profiles p
WHERE p.user_id = p_user_id;
```

Usage events and invoices are lower risk (they're inherently user-visible data), but the subscription row may contain `stripe_customer_id`, `stripe_subscription_id`, or other identifiers that operators don't need displayed in the drawer.

supabase/migrations/20260518000000_admin_v1.sql
```diff
@@ -0,0 +1,340 @@
...
+            'sort_dir', p_sort_dir,
+            'product_id', p_product_id
+        )
+    );
+
+    SELECT count(*)::bigint INTO v_total
+    FROM auth.users u
+    LEFT JOIN public.user_profiles p ON p.user_id = u.id
+    WHERE p_search IS NULL
+       OR u.email ILIKE '%' || p_search || '%';
```
diego-artificerinnovations: nit: `p_search` wildcards are not escaped, so `%`, `_`, and `\` in the search term act as ILIKE metacharacters. Searching for `%` matches all emails; searching for `_` matches any single character in that position.

For an admin-only panel this is acceptable UX (admins can't do anything with the extra results that they couldn't already do), but it can produce surprising pagination totals. A one-liner fix if you want strict prefix/infix matching:

```sql
OR u.email ILIKE '%' || replace(replace(replace(p_search, '!', '!!'), '%', '!%'), '_', '!_') || '%' ESCAPE '!'
```

(Same applies to line 191 in the paginated query — both the count and data queries need the same treatment to stay in sync.)

supabase/tests/admin.test.sql
diego-artificerinnovations, line 231: nit: the pgTAP suite tests `admin_list_users` as an admin (plan → total is not null) but does not test `admin_get_user` as an admin. Worth adding a case:

```sql
-- after setting JWT claims for operator (a1000000-...-0002):
SELECT ok(
    (public.admin_get_user('a1000000-0000-0000-0000-000000000001'::uuid) -> 'auth') IS NOT NULL,
    'admin admin_get_user returns auth block'
);

-- and verify the audit log captured the view event:
SELECT ok(
  (SELECT count(*)::int FROM public.admin_audit_log
   WHERE actor_user_id = 'a1000000-0000-0000-0000-000000000002'
     AND action = 'admin.users.view') >= 1,
  'admin_get_user wrote audit log row'
);
```

The plan count would need bumping from 15 to 17.

diego-artificerinnovations (approved):
## Security review — Admin v1

The DB-layer design is sound: RLS enabled with zero client-role policies (tables are opaque to PostgREST), all entry points are `SECURITY DEFINER` with `SET search_path = public`, auth is fail-closed (`checkIsAdmin` returns false on error, `admin_is_admin()` returns false for null UID), and the `useIsAdmin` race fix is correct — the `state.userId === userId` guard in the `loading` and `isAdmin` derivations prevents stale RPC results from flash-authorizing a different session. Four security nits below, none blocking.

scripts/grant-admin.ts
```diff
@@ -0,0 +1,82 @@
...
+  } catch (e) {
+    console.error(e instanceof Error ? e.message : e);
+    process.exit(1);
+  }
+
+  const supabase = createClient(url, serviceKey, {
+    auth: { persistSession: false, autoRefreshToken: false },
+  });
+
+  const { data: users, error: listErr } = await supabase.auth.admin.listUsers();
```
rahul-artificerinnovations: `supabase.auth.admin.listUsers()` returns at most 50 users by default and does not paginate. In a production app with more than 50 auth accounts, users beyond the first page will not be found — the script will print `No auth user found with email:` even when the account exists, silently failing to grant admin.

Fix by iterating pages until the email is matched, or by using a higher `perPage` with awareness of the limit:
```ts
// Option A: paginate until found
let page = 1;
let match: { id: string } | undefined;
do {
  const { data, error } = await supabase.auth.admin.listUsers({ page, perPage: 100 });
  if (error) { console.error('Failed to list users:', error.message); process.exit(1); }
  match = data.users.find(u => u.email?.toLowerCase() === email.toLowerCase());
  if (data.users.length < 100) break;  // last page
  page++;
} while (!match);
```
Same fix needed in `revoke-admin.ts`.

scripts/revoke-admin.ts
```diff
@@ -0,0 +1,78 @@
...
+  } catch (e) {
+    console.error(e instanceof Error ? e.message : e);
+    process.exit(1);
+  }
+
+  const supabase = createClient(url, serviceKey, {
+    auth: { persistSession: false, autoRefreshToken: false },
+  });
+
+  const { data: users, error: listErr } = await supabase.auth.admin.listUsers();
```
rahul-artificerinnovations: Same `listUsers()` pagination issue as `grant-admin.ts` — only the first 50 users are searched. For any production app with more than 50 accounts, this will print `No auth user found` even when the account exists. See the suggested fix in the `grant-admin.ts` comment.

supabase/migrations/20260518000000_admin_v1.sql
```diff
@@ -0,0 +1,340 @@
...
+            'sort', p_sort,
+            'sort_dir', p_sort_dir,
+            'product_id', p_product_id
+        )
+    );
+
+    SELECT count(*)::bigint INTO v_total
+    FROM auth.users u
+    LEFT JOIN public.user_profiles p ON p.user_id = u.id
+    WHERE p_search IS NULL
```
rahul-artificerinnovations: No SQL injection here — `p_search` is a PL/pgSQL variable used as a parameterized value, not concatenated into a dynamic `EXECUTE` string, so the query tree is fixed. However, `ILIKE '%' || p_search || '%'` passes LIKE metacharacters through unescaped: `p_search = '%'` matches every email, and `p_search = 'a_b'` matches `a` + any single char + `b`. Since only admins reach this code and can enumerate all users via pagination anyway, this is no privilege gain — but it may surprise callers who expect literal substring matching.

If the intent is literal substring search, escape metacharacters before the concatenation:
```sql
u.email ILIKE '%' || regexp_replace(p_search, '([%_\\])', '\\\1', 'g') || '%'
```
This applies to both the `count(*)` query at line ~156 and the main `SELECT` below. Non-blocking, but worth aligning the implementation with the documented intent.

supabase/migrations/20260518000000_admin_v1.sql
```diff
@@ -0,0 +1,340 @@
...
+CREATE OR REPLACE FUNCTION public.admin_is_admin()
+RETURNS boolean
+LANGUAGE sql
+STABLE
+SECURITY DEFINER
+SET search_path = public
+AS $$
+    SELECT EXISTS (
+        SELECT 1
+        FROM public.admin_users a
```
rahul-artificerinnovations, line 54: `admin_record_audit_event` silently returns `void` when called by a non-admin (`v_uid IS NULL OR NOT admin_is_admin()`). This is intentional — it prevents leaking that the admin panel exists. The tradeoff is that a misconfigured caller (e.g., a new admin feature accidentally running under a non-admin session) will silently drop audit events with no error signal. The SECURITY.md doc covers the intent; worth a brief inline comment here explaining the deliberate silent-drop behavior so future maintainers don't "fix" it by adding an error return.

rahul-artificerinnovations (approved):
The security model here is solid. RLS on both admin tables with zero client-accessible policies, SECURITY DEFINER + fixed `search_path` on every RPC, admin re-check server-side inside every sensitive call, `not_found` opaque response for non-admins, grant/revoke gated behind the service-role CLI, and a `useIsAdmin` implementation that holds `loading=true` on the first render so the route guard doesn't false-negative before the RPC completes. The pgTAP tests verify the non-admin → `not_found` invariant at the DB layer. No path for privilege escalation or cross-user data access found.

Inline comments below — the `listUsers()` pagination in the CLI scripts is the only item I'd call a functional bug (not a security vulnerability, but it means `admin:grant` silently fails in any app with >50 users). Everything else is observation-level.

Two additional observations not worth separate inline comments:

**`_admin_insert_audit` in generated types** — The function is `REVOKE ALL FROM PUBLIC`, so PostgREST won't expose it and clients can't call it. But it appears in both generated type files (`apps/web/src/types/database.ts` and `apps/mobile/src/types/database.ts`). The underscore convention is the hint that it's private, but adopters scanning the types may try `supabase.rpc('_admin_insert_audit', ...)` and get a cryptic PostgREST error. Low risk; worth a comment near the function in the migration that it's internal-only and not PostgREST-exposed.

**pgTAP `admin_get_user` positive case missing** — The test suite checks that the function exists and that non-admins get `not_found`, but there's no positive assertion that it returns the target user's data when called as admin. The existing tests are sufficient to block accidental regression of the auth guard; adding a happy-path case would make the test suite complete. Non-blocking.

Good work overall — the defense-in-depth framing (guard is UX, not security), the VOLATILE fix migration, and the audit log automatically populated on list/view are all good calls.

supabase/migrations/20260518000000_admin_v1.sql
```diff
@@ -0,0 +1,340 @@
...
+        'user',
+        p_user_id::text,
+        jsonb_build_object('product_id', p_product_id)
+    );
+
+    SELECT to_jsonb(p.*) INTO v_profile
+    FROM public.user_profiles p
+    WHERE p.user_id = p_user_id;
+
+    SELECT to_jsonb(s.*) INTO v_subscription
```
mei-artificerinnovations, line 297: **HIGH** — to_jsonb(s.*) serializes every column of billing_subscriptions, including any column added in the future. Same issue on lines 278 (user_profiles), 287 (billing_plans), 292 (billing_usage_aggregates), 300 (billing_usage_events), and 311 (billing_invoices).

Use explicit projections like the auth.users block above:
```sql
SELECT jsonb_build_object(
    'id', s.id,
    'plan_id', s.plan_id,
    'status', s.status,
    'current_period_end', s.current_period_end
    -- only fields the UI actually needs
) INTO v_subscription
FROM public.billing_subscriptions s
WHERE s.user_id = p_user_id AND s.product_id = p_product_id;
```

scripts/grant-admin.ts
mei-artificerinnovations: **HIGH** — listUsers() with no pagination args returns at most the first page (default 1000 users). On a workspace with >1000 accounts the find() silently returns undefined and the script exits with "No auth user found" even though the user exists. Paginate with a loop or use a direct DB lookup by email.

scripts/grant-admin.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  const match = users.users.find(
+    u => u.email?.toLowerCase() === email.toLowerCase()
+  );
+  if (!match) {
+    console.error(`No auth user found with email: ${email}`);
+    process.exit(1);
+  }
+
+  const { error: upsertErr } = await supabase.from('admin_users').upsert(
```
mei-artificerinnovations, line 97: **MEDIUM** — granted_by is not set in this upsert. All script-granted admins will have NULL in admin_users.granted_by. Consider passing the operator identity (e.g. a --granted-by flag or ADMIN_GRANTER env var) and writing it here.

scripts/revoke-admin.ts
mei-artificerinnovations: **HIGH** — Same listUsers() pagination issue as grant-admin.ts:48. Users past page 1 will not be found, causing a silent no-op.

scripts/revoke-admin.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  const match = users.users.find(
+    u => u.email?.toLowerCase() === email.toLowerCase()
+  );
+  if (!match) {
+    console.error(`No auth user found with email: ${email}`);
+    process.exit(1);
+  }
+
+  const { error: updateErr } = await supabase
```
mei-artificerinnovations: **MEDIUM** — Supabase .update() does not error when zero rows match. If the target was never an admin or was already revoked, this silently no-ops, and the script still prints "Revoked admin from X". Add a .select() and check data?.length > 0 before printing success.

packages/admin/src/hooks/useIsAdmin.ts
```diff
@@ -0,0 +1,69 @@
...
+      return;
+    }
+
+    const checkFor = userId;
+    setState({ userId: checkFor, isAdmin: false, status: 'loading' });
+    setError(null);
+    try {
+      const ok = await checkIsAdmin(supabase);
+      setState(prev => {
+        if (checkFor !== userId) return prev;
```
mei-artificerinnovations: **LOW** — checkFor is assigned from userId on line 46, and userId here is the same closure value captured by refresh. So checkFor !== userId is always false and this guard never fires — dead code. If the intent is to discard stale async results, use an AbortController or a cancelled ref that gets flipped in the effect cleanup.

mei-artificerinnovations (changes requested):
## Security Review — feat/admin-panel

Reviewed: supabase/migrations/, scripts/grant-admin.ts, scripts/revoke-admin.ts, packages/admin/src/, and the web admin pages.

**2 HIGH · 2 MEDIUM · 1 LOW** — the two HIGH items should block merge.

---

### HIGH-1 · to_jsonb(table.*) leaks all columns including future-added sensitive fields

admin_get_user uses to_jsonb(s.*) / to_jsonb(e.*) / to_jsonb(i.*) / to_jsonb(p.*) to serialize entire rows from billing_subscriptions, billing_usage_events, billing_invoices, and user_profiles. Any new column added to those tables — a stripe_payment_method_id, raw webhook payload field, internal notes — will automatically appear in admin API responses without any code change or review gate. The fix is explicit jsonb_build_object(...) projections, same as the pattern already used for auth.users. See inline comment.

### HIGH-2 · listUsers() called without pagination in both scripts

grant-admin.ts:48 and revoke-admin.ts:47 call supabase.auth.admin.listUsers() with no perPage/page args. The default page size is 1000. On a workspace with >1000 users, the find() silently searches only the first page — a user past page 1 returns undefined, the script exits with "No auth user found with email", and the grant/revoke silently fails.

### MEDIUM-1 · grant-admin.ts upsert omits granted_by

The upsert at line 62 does not write granted_by, so all script-granted admins have NULL there. The audit log captures the event, but admin_users.granted_by provides no CLI operator trail.

### MEDIUM-2 · revoke-admin.ts does not confirm the update affected a row

.update().eq().is('revoked_at', null) silently no-ops if the target was never an admin or was already revoked. The script then prints "Revoked admin from X" unconditionally.

### LOW · useIsAdmin abort guard is a no-op

See inline comment on useIsAdmin.ts:52. Does not cause incorrect behavior today but is dead code.

scripts/grant-admin.ts
ZappoMan: Fixed in 4159bf0: extracted findAuthUserByEmail that paginates auth.admin.listUsers (perPage 1000). Used by grant and revoke.

scripts/grant-admin.ts
ZappoMan: Fixed in 4159bf0: optional `--granted-by <email>` populates `admin_users.granted_by`. Documented in grant script + docs/guides/admin-setup.md.

scripts/grant-admin.ts
ZappoMan: Same — `--granted-by` added in 4159bf0 for CLI attribution to an auth user.

scripts/grant-admin.ts
ZappoMan: Fixed in 4159bf0 via shared `scripts/lib/findAuthUserByEmail.ts` (paginated listUsers).

scripts/revoke-admin.ts
ZappoMan: Fixed in 4159bf0 — revoke uses the same paginated lookup.

scripts/grant-admin.ts
```diff
@@ -0,0 +1,82 @@
...
+    console.error('Failed to list users:', listErr.message);
+    process.exit(1);
+  }
+
+  const match = users.users.find(
+    u => u.email?.toLowerCase() === email.toLowerCase()
+  );
+  if (!match) {
+    console.error(`No auth user found with email: ${email}`);
+    process.exit(1);
```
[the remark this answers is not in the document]
ZappoMan: Addressed in 4159bf0 (findAuthUserByEmail).

scripts/revoke-admin.ts
```diff
@@ -0,0 +1,78 @@
...
+    process.exit(1);
+  }
+
+  const match = users.users.find(
+    u => u.email?.toLowerCase() === email.toLowerCase()
+  );
+  if (!match) {
+    console.error(`No auth user found with email: ${email}`);
+    process.exit(1);
+  }
```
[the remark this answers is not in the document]
ZappoMan: Addressed in 4159bf0 (findAuthUserByEmail).

supabase/migrations/20260518000000_admin_v1.sql
ZappoMan: Fixed in 4159bf0: cap search at 200 chars, escape ILIKE metacharacters, migration 20260518130000_admin_list_users_search_hardening.sql.

supabase/migrations/20260518000000_admin_v1.sql
ZappoMan: Addressed with ILIKE escape in 4159bf0.

supabase/migrations/20260518000000_admin_v1.sql
```diff
@@ -0,0 +1,340 @@
...
+                  AND a.period_end > now()
+            )
+        ) AS row_data
+        FROM auth.users u
+        LEFT JOIN public.user_profiles p ON p.user_id = u.id
+        LEFT JOIN public.billing_subscriptions sub
+            ON sub.user_id = u.id AND sub.product_id = p_product_id
+        LEFT JOIN public.billing_plans pl ON pl.id = sub.plan_id
+        WHERE p_search IS NULL
+           OR u.email ILIKE '%' || p_search || '%'
```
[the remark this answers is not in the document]
ZappoMan: Addressed with ILIKE escape in 4159bf0.

supabase/migrations/20260518000000_admin_v1.sql
ZappoMan: Agreed — added escape + length cap in 4159bf0.

packages/admin/SECURITY.md
ZappoMan: Documented in packages/admin/SECURITY.md (Client route guard cache): guard is UX-only; RPCs enforce immediately.

scripts/revoke-admin.ts
ZappoMan: Fixed in 4159bf0: revoke uses .select() and errors when no active admin row was updated.

packages/admin/src/hooks/useIsAdmin.ts
```diff
@@ -0,0 +1,69 @@
...
+        return { userId: checkFor, isAdmin: ok, status: 'done' };
+      });
+    } catch (e) {
+      setState(prev => {
+        if (checkFor !== userId) return prev;
+        return { userId: checkFor, isAdmin: false, status: 'done' };
+      });
+      setError(e instanceof Error ? e : new Error(String(e)));
+    }
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 4159bf0: userIdRef so stale async RPC results are ignored when session user changes.

scripts/lib/loadSupabaseServiceEnv.ts
```diff
@@ -0,0 +1,64 @@
...
+    const match = trimmed.match(/^([A-Z0-9_]+)="(.*)"$/);
+    if (match) {
+      const key = match[1];
+      const value = match[2];
+      if (key !== undefined && value !== undefined) {
+        out[key] = value;
+      }
+    }
+  }
+  return out;
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 4159bf0: parseSupabaseStatusEnv accepts quoted and unquoted status lines.

scripts/lib/loadSupabaseServiceEnv.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  try {
+    const stdout = execSync('supabase status -o env', {
+      encoding: 'utf8',
+      stdio: ['ignore', 'pipe', 'ignore'],
+    });
+    const fromStatus = parseSupabaseStatusEnv(stdout);
+    url = url ?? fromStatus.API_URL ?? fromStatus.SUPABASE_URL;
+    serviceKey = serviceKey ?? fromStatus.SERVICE_ROLE_KEY;
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 4159bf0: bracket access for parsed env keys.

supabase/tests/admin.test.sql
ZappoMan: Extended pgTAP in 4159bf0: revoked admin, empty-JWT not_found, positive admin_get_user. Raw anon EXECUTE skipped (pgTAP runs as superuser).

ZappoMan: Added positive admin_get_user assertions in 4159bf0.

packages/admin/src/components/AdminRoute.web.tsx
ZappoMan: Not changed in this pass — agree a retry/error UI would help; useIsAdmin already exposes error for future AdminRoute handling. Tracking as follow-up unless you want it in this PR.

supabase/migrations/20260518000000_admin_v1.sql
ZappoMan: Valid nit — left as two queries for v1 readability; happy to CTE in a follow-up.

supabase/migrations/20260518000000_admin_v1.sql
ZappoMan: Valid nit — projecting columns vs to_jsonb(*) is a good v1.1 hardening; out of scope for this PR unless you want it now.

supabase/migrations/20260518000000_admin_v1.sql
ZappoMan: Acknowledged — same projection hardening as above; admins are trusted operators in v1.

supabase/migrations/20260518000000_admin_v1.sql
ZappoMan: Intentional silent no-op for non-admins (no existence leak). Documented in migration comment near _admin_insert_audit.

supabase/migrations/20260518000000_admin_v1.sql
```diff
@@ -0,0 +1,340 @@
...
+        NULL,
+        jsonb_build_object(
+            'limit', v_limit,
+            'offset', v_offset,
+            'search', p_search,
+            'sort', p_sort,
+            'sort_dir', p_sort_dir,
+            'product_id', p_product_id
+        )
+    );
```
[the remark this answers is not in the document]
ZappoMan: Acknowledged — audit-before-fetch is acceptable in one transaction; zero-row list still audited by design.

supabase/migrations/20260518000000_admin_v1.sql
```diff
@@ -0,0 +1,340 @@
...
+        RETURN jsonb_build_object('error', 'not_found');
+    END IF;
+
+    PERFORM public._admin_insert_audit(
+        v_uid,
+        'admin.users.view',
+        'user',
+        p_user_id::text,
+        jsonb_build_object('product_id', p_product_id)
+    );
```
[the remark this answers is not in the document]
ZappoMan: Audit runs after target existence is confirmed (auth row found). Order is correct in current SQL.

packages/admin/src/components/AdminLayout.web.tsx
```diff
@@ -0,0 +1,67 @@
...
+        </div>
+        <nav
+          className='flex-1 px-2 py-4 space-y-1'
+          aria-label='Admin navigation'
+        >
+          {navItems.map(item => (
+            <NavLink
+              key={item.to}
+              to={item.to}
+              end={item.to === '/admin'}
```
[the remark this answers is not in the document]
ZappoMan: Fair point for forked mount paths — adopters can pass basePath to AdminLayout in a follow-up; v1 matches BeakerStack /admin.

packages/admin/src/adminClient.ts
```diff
@@ -0,0 +1,90 @@
...
+/** PostgREST errors are plain objects; wrap so callers get a readable message. */
+function rpcError(error: { message?: string }): Error {
+  return new Error(error.message ?? 'RPC request failed');
+}
+
+export async function checkIsAdmin(supabase: SupabaseClient): Promise<boolean> {
+  const { data, error } = await supabase.rpc('admin_is_admin');
+  if (error) return false;
+  return Boolean(data);
+}
```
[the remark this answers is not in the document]
ZappoMan: Partially addressed: errors surface via useIsAdmin.error; checkIsAdmin still returns false on RPC error (fail-closed). AdminRoute retry is follow-up.

packages/admin/src/adminClient.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export async function recordAuditEvent(
+  supabase: SupabaseClient,
+  input: RecordAuditEventInput
+): Promise<void> {
+  const { error } = await supabase.rpc('admin_record_audit_event', {
+    p_action: input.action,
+    p_target_type: input.target?.type ?? undefined,
+    p_target_id: input.target?.id ?? undefined,
+    p_details: (input.details ?? {}) as Record<string, never>,
```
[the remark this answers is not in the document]
ZappoMan: Acknowledged — will tighten typing in a small follow-up; cast is harmless at runtime.

packages/admin/src/components/AdminPagination.web.tsx
```diff
@@ -0,0 +1,49 @@
...
+export function AdminPagination({
+  total,
+  limit,
+  offset,
+  onPageChange,
+}: AdminPaginationProps) {
+  const page = Math.floor(offset / limit) + 1;
+  const totalPages = Math.max(1, Math.ceil(total / limit));
+  const canPrev = offset > 0;
+  const canNext = offset + limit < total;
```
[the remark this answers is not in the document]
ZappoMan: Valid — will add limit guard in follow-up; callers currently always pass positive limit from RPC defaults.

packages/admin/src/components/AdminDetailDrawer.web.tsx
```diff
@@ -0,0 +1,46 @@
...
+            className='rounded-md p-2 text-gray-500 hover:bg-gray-100 hover:text-gray-700'
+            aria-label='Close'
+          >
+            ✕
+          </button>
+        </div>
+        <div className='flex-1 overflow-y-auto px-4 py-4'>{children}</div>
+      </aside>
+    </div>
+  );
```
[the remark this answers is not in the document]
ZappoMan: Will verify z-index/stacking in follow-up; drawer uses standard fixed overlay pattern.

supabase/migrations/20260518000000_admin_v1.sql
ZappoMan: Tracking in #263 — will refactor `admin_list_users` to a single filtered scan (CTE + window count) in a follow-up PR.

packages/admin/src/components/AdminRoute.web.tsx
ZappoMan: Tracking in #266 — `AdminRoute` will distinguish RPC errors from confirmed non-admin and expose retry via `useIsAdmin.refresh()`.

supabase/migrations/20260518000000_admin_v1.sql
ZappoMan: Tracking in #264 — will replace `to_jsonb(*)` with explicit column projections in `admin_get_user`.

packages/admin/src/adminClient.ts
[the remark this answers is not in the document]
ZappoMan: Tracking in #266 — includes clearer error vs not-admin handling in `AdminRoute` / `useIsAdmin` (fail-closed in `checkIsAdmin` remains intentional for the guard).

packages/admin/src/adminClient.ts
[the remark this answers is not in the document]
ZappoMan: Tracking in #266 — will fix `recordAuditEvent` details typing (`Record<string, unknown>` instead of `never`).

packages/admin/src/components/AdminPagination.web.tsx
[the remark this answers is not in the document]
ZappoMan: Tracking in #265 — will guard `limit === 0` in `AdminPagination`.

packages/admin/src/components/AdminDetailDrawer.web.tsx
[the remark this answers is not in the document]
ZappoMan: Tracking in #265 — will verify backdrop/header stacking and focus behavior in `AdminDetailDrawer`.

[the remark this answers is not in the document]
ZappoMan: Tracking in #265 — same issue covers focus trap on open and restore on close.

packages/admin/src/components/AdminTable.web.tsx
```diff
@@ -0,0 +1,96 @@
...
+            ) : (
+              rows.map(row => (
+                <tr
+                  key={getRowKey(row)}
+                  onClick={onRowClick ? () => onRowClick(row) : undefined}
+                  className={
+                    onRowClick
+                      ? 'cursor-pointer hover:bg-indigo-50/50 transition-colors'
+                      : undefined
+                  }
```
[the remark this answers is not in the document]
ZappoMan: Tracking in #265 — will add keyboard/`role` support for clickable table rows.

apps/web/src/admin/pages/AdminUsersPage.tsx
```diff
@@ -0,0 +1,154 @@
...
+            type='button'
+            className='text-left font-semibold uppercase tracking-wide hover:text-indigo-600'
+            onClick={e => {
+              e.stopPropagation();
+              toggleSort('last_active');
+            }}
+          >
+            Last active{' '}
+            {sort === 'last_active' ? (sortDir === 'asc' ? '↑' : '↓') : ''}
+          </button>
```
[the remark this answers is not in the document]
ZappoMan: Tracking in #265 — will ensure sort controls and row activation work for keyboard users.

packages/admin/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "test": "NODE_OPTIONS=--max-old-space-size=8192 vitest --run",
+    "test:coverage": "vitest --coverage --run"
+  },
+  "dependencies": {
+    "@supabase/supabase-js": "^2.38.0"
+  },
+  "peerDependencies": {
+    "react": "18.2.0",
+    "react-dom": "18.2.0",
+    "react-router-dom": ">=6.0.0"
```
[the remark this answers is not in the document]
ZappoMan: Tracking in #267 — peerDependency semver ranges + devDeps for package tests.

packages/admin/package.json
```diff
@@ -0,0 +1,50 @@
...
+  },
+  "peerDependenciesMeta": {
+    "react-router-dom": {
+      "optional": true
+    }
+  },
```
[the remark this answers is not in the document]
ZappoMan: Tracking in #267 — will mark `react-router-dom` as a required peer (used by all `@beakerstack/admin/web` exports).

packages/admin/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "eslint": "^8.57.0",
+    "eslint-config-prettier": "^9.1.0",
+    "jsdom": "^26.1.0",
+    "react": "18.2.0",
+    "react-dom": "18.2.0",
+    "react-router-dom": "^6.30.3",
+    "typescript": "^5.3.0",
+    "@vitest/coverage-v8": "^3.2.4",
+    "vitest": "^3.2.4"
+  }
```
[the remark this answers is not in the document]
ZappoMan: Tracking in #267 — includes adding React to devDependencies for Vitest.

packages/admin/src/types.ts
```diff
@@ -0,0 +1,52 @@
...
+  label: string;
+  to: string;
+  icon?: ReactNode;
+};
+
+export type RecordAuditEventInput = {
+  action: string;
+  target?: { type?: string; id?: string };
+  details?: Record<string, unknown>;
+};
```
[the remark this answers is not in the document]
ZappoMan: Tracking in #266 — `RecordAuditEventInput.details` typing aligned with jsonb RPC args.

apps/web/src/App.tsx
```diff
@@ -68,17 +86,37 @@ function App() {
...
-                    <AuthenticatedApp />
-                  </Suspense>
+                  <AdminRouteGate>
+                    <Suspense fallback={<PageFallback />}>
+                      <AdminApp />
+                    </Suspense>
+                  </AdminRouteGate>
                 }
-              >
+              />
```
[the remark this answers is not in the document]
ZappoMan: Tracking in #269 — optional follow-up to skip `ProfileProvider` on `/admin` routes (acceptable overhead for v1).

packages/admin/src/hooks/useIsAdmin.ts
```diff
@@ -0,0 +1,69 @@
...
+  }, [supabase, userId]);
+
+  useEffect(() => {
+    void refresh();
+  }, [refresh]);
```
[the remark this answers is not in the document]
ZappoMan: Tracking in #266 — wire `refresh` on focus/visibility and error retry in `AdminRoute`.

supabase/migrations/20260518000000_admin_v1.sql
```diff
@@ -0,0 +1,340 @@
...
+LANGUAGE plpgsql
+SECURITY DEFINER
+SET search_path = public
+AS $$
+DECLARE
+    v_uid uuid := auth.uid();
+BEGIN
+    IF v_uid IS NULL OR NOT public.admin_is_admin() THEN
+        RETURN;
+    END IF;
```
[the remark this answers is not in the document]
ZappoMan: Tracking in #268 — document intentional silent no-op for non-admin callers (security: no existence leak).

supabase/migrations/20260518000000_admin_v1.sql
ZappoMan: Tracking in #264 — same projection hardening as Sarah's review; will drop wildcard `to_jsonb(*)` in `admin_get_user`.

scripts/grant-admin.ts
ZappoMan: Addressed in 4159bf0: `scripts/lib/findAuthUserByEmail.ts` paginates `auth.admin.listUsers` (perPage 1000) until the email is found.

scripts/grant-admin.ts
ZappoMan: Addressed in 4159bf0: optional `--granted-by <email>` sets `admin_users.granted_by` when that auth user exists. See `docs/guides/admin-setup.md`.

scripts/revoke-admin.ts
ZappoMan: Addressed in 4159bf0: revoke uses the same paginated `findAuthUserByEmail` helper as grant.

scripts/revoke-admin.ts
ZappoMan: Addressed in 4159bf0: revoke uses `.select('user_id')` and exits with an error when no active admin row was updated.

packages/admin/src/hooks/useIsAdmin.ts
ZappoMan: Addressed in 4159bf0: `userIdRef` fixes the stale-async guard (closure `userId` was always equal to `checkFor` within one invocation).

supabase/migrations/20260518000000_admin_v1.sql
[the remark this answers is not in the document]
ZappoMan: Tracking in #270 — documents the trade-off (no denial rows in admin_audit_log today to avoid leaking admin existence) and outlines an optional service-role-only denied-attempt sink for production security monitoring.

apps/web/src/admin/pages/AdminUsersPage.tsx
```diff
@@ -0,0 +1,168 @@
...
+            }
+            onClick={e => {
+              e.stopPropagation();
+              toggleSort('last_active');
+            }}
+          >
+            Last active{' '}
+            {sort === 'last_active' ? (sortDir === 'asc' ? '↑' : '↓') : ''}
+          </button>
+        ),
```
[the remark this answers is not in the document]
ZappoMan: Good catch — moved `aria-sort` onto the `<th scope="col">` via new optional `headerCellProps` on `AdminTableColumn`, and removed it from the inner sort buttons in `AdminUsersPage` (fixed in f2c675b).

apps/web/src/admin/hooks/useAdminUsers.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+  }, [offset, debouncedSearch, sort, sortDir]);
+
+  useEffect(() => {
+    void load();
+  }, [load]);
+
+  useEffect(() => {
+    setOffset(0);
+  }, [debouncedSearch, sort, sortDir]);
```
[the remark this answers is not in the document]
ZappoMan: Addressed in f2c675b: `setSearch` now resets `offset` synchronously, `toggleSort` resets offset before updating sort, and we skip `load` while the debounce is pending (`search.trim() !== debouncedSearch`) so we do not fire a stale-offset `admin_list_users` RPC when filters change.

packages/admin/src/components/AdminTable.web.tsx
```diff
@@ -0,0 +1,114 @@
...
+                              onRowClick(row);
+                            }
+                          }
+                        : undefined
+                    }
+                    className={
+                      onRowClick
+                        ? 'cursor-pointer hover:bg-indigo-50/50 transition-colors focus:outline-none focus-visible:ring-2 focus-visible:ring-indigo-500 focus-visible:ring-inset'
+                        : undefined
+                    }
```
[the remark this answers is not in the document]
ZappoMan: Added optional `getRowLabel(row)` on `AdminTable` (defaults to `getRowKey`). `AdminUsersPage` passes `row.email ?? row.display_name ?? row.user_id` for the row accessible name (f2c675b).

packages/admin/src/adminClient.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  const body = data as Record<string, unknown>;
+  const users = (body['users'] as AdminUserListRow[] | undefined) ?? [];
+  return {
+    users,
+    total: Number(body['total'] ?? 0),
+    limit: Number(body['limit'] ?? params.limit ?? 25),
+    offset: Number(body['offset'] ?? params.offset ?? 0),
+  };
+}
```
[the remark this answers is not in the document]
ZappoMan: Added `isRecord` validation on the RPC payload plus explicit checks that `total`/`limit`/`offset` parse as finite numbers; otherwise we throw a clear error instead of propagating `NaN` into pagination (f2c675b).

packages/admin/src/components/AdminDetailDrawer.web.tsx
```diff
@@ -0,0 +1,105 @@
...
+      const last = focusables[focusables.length - 1]!;
+
+      if (event.shiftKey && document.activeElement === first) {
+        event.preventDefault();
+        last.focus();
+      } else if (!event.shiftKey && document.activeElement === last) {
+        event.preventDefault();
+        first.focus();
+      }
+    };
```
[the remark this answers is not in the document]
ZappoMan: Tightened focus-trap filtering with `isPanelFocusable` (excludes disabled, `tabIndex < 0`, and `aria-hidden="true"`) in addition to the selector. Added a test for `tabIndex={-1}` children (f2c675b).

supabase/migrations/20260518000000_admin_v1.sql
diego-artificerinnovations: Still open — `to_jsonb(p.*)`, `to_jsonb(s.*)`, `to_jsonb(a.*)`, `to_jsonb(e.*)`, and `to_jsonb(i.*)` remain in the current HEAD (`f2c675b9`). Mei flagged this HIGH as well. The other items from my review were all addressed in 4159bf0 — this is the only one still outstanding.

mei-artificerinnovations (approved):
## Re-review — approving with one tracked caveat

Four of five findings addressed cleanly:

- ✅ **HIGH-2 (listUsers pagination)** — extracted into `findAuthUserByEmail` with a proper page-loop (`page += 1` until `data.users.length < PAGE_SIZE`). Correct.
- ✅ **MEDIUM-1 (granted_by)** — added `--granted-by` flag; resolves the granter's UUID and writes it to the upsert. Clean.
- ✅ **MEDIUM-2 (revoke no-op)** — `.select('user_id')` + `!data?.length` check exits with an error instead of printing success. Exactly right.
- ✅ **LOW (useIsAdmin no-op guard)** — fixed with `userIdRef` — `checkFor !== userIdRef.current` now correctly compares the closure-captured value against the *current* userId at callback time.

---

**Remaining: HIGH-1 — `to_jsonb(table.*)` in `admin_get_user`**

Lines 293, 297, 302, 307, 315, and 326 of the migration still use `to_jsonb(p.*)` / `to_jsonb(s.*)` / etc. for `user_profiles`, `billing_subscriptions`, `billing_plans`, `billing_usage_aggregates`, `billing_usage_events`, and `billing_invoices`.

I'm **approving** because: the function is admin-only (gated behind `admin_is_admin()`), the current tables don't carry unexpectedly sensitive columns, and the other acute issues are all resolved. But this is a forward-looking landmine — any new column added to those billing tables (a Stripe payment method field, a raw webhook body, internal flags) will automatically leak through the admin API without any code review gate.

**Request:** file a follow-up issue before this merges, or add a schema-review note to the billing table migration checklist. The fix is straightforward — replace `to_jsonb(s.*)` with explicit `jsonb_build_object('id', s.id, 'plan_id', s.plan_id, ...)` projections listing only the fields the admin UI actually uses.

**Approving now** — the net security posture of this PR is a large improvement over the status quo (no admin panel at all).

diego-artificerinnovations:
**Re-review — three of four security nits resolved, one still open**

Checked HEAD `f2c675b9`. Here's the status on my original review:

| # | Item | Status |
|---|------|--------|
| 1 | `granted_by` unattributed CLI grants | ✅ Fixed — `--granted-by` flag, `findAuthUserByEmail` paginated helper |
| 2 | `revoke-admin` false-positive success log | ✅ Fixed — `.select('user_id')` + `data?.length` guard |
| 3 | `to_jsonb(p.*)` wildcard in `admin_get_user` | ❌ **Still open** — `to_jsonb(p.*)`, `to_jsonb(s.*)`, `to_jsonb(a.*)`, `to_jsonb(e.*)`, `to_jsonb(i.*)` remain at HEAD |
| 4 | ILIKE metacharacter passthrough in search | ✅ Fixed — escape + 200-char cap in 4159bf0 and the separate hardening migration |

**Remaining item is the schema-coupling risk**: any column added to `user_profiles`, `billing_subscriptions`, `billing_plans`, `billing_usage_aggregates`, `billing_usage_events`, or `billing_invoices` in the future will be silently returned to every admin session without any code change. Mei flagged this HIGH as well. Fix is an explicit `jsonb_build_object` projection for each of those tables, consistent with what `admin_list_users` already does.

Everything else — `admin_is_admin()` VOLATILE, `useIsAdmin` stale-closure fix via `useRef`, extended pgTAP coverage (plan 20), `findAuthUserByEmail` pagination, `loadSupabaseServiceEnv` unquoted-value handling — all look good.

Happy to re-approve once the `to_jsonb` projections are explicit.

rahul-artificerinnovations (commented):
Re-checked the seven new commits against my original review and Mei's. Here's where things stand:

**All my original concerns: resolved ✅**
- Pagination bug in grant/revoke scripts → replaced with `findAuthUserByEmail` helper (perPage: 1000, full iteration)
- `granted_by` not written → `--granted-by <email>` flag resolves and stores granter's user_id
- ILIKE metacharacter injection → `\`, `%`, `_` escaped; `ESCAPE '\'` clause added; 200-char cap applied; separate hardening migration for existing deployments
- `_admin_insert_audit` silent no-op undocumented → comment added
- `admin_is_admin()` STABLE → VOLATILE fix landed in migration and the new hardening migration

**Bonus fixes landed without being asked:**
- `checkIsAdmin` now throws on RPC error (previously swallowed → silent false)
- `AdminRoute` shows recoverable error state + "Try again" button instead of silently redirecting to not-authorized
- `AdminRoute` re-checks admin status on window `focus` + `visibilitychange`
- `useIsAdmin` stale-response guard switched to `userIdRef.current` (fixes the no-op Mei flagged)
- `AdminDetailDrawer`: full focus trap, Escape key, restored-focus-on-close, `aria-labelledby`
- `AdminTable`: keyboard activation (Enter/Space), `role="button"`, `aria-label`, `aria-sort` on sortable headers
- `AdminPagination`: `safeLimit = Math.max(1, limit)` guards NaN
- `useAdminUsers`: `setSearch` resets page, `toggleSort` resets page before toggling, `searchPending` flag prevents stale fetch
- Test coverage raised, revoke now confirms the update hit a row

**One item still open — Mei's HIGH-1:**

`admin_get_user` uses `to_jsonb(s.*)`, `to_jsonb(e.*)`, `to_jsonb(i.*)`, `to_jsonb(p.*)` to serialize full rows from `billing_subscriptions`, `billing_usage_events`, `billing_invoices`, and `user_profiles`. Any column added to those tables in the future — Stripe payment method tokens, raw webhook payloads, internal notes — automatically appears in operator API responses with no code change or review gate. The `admin_list_users` function already uses explicit `jsonb_build_object` projections; `admin_get_user` should follow the same pattern. This is the last blocker I see before merge.

sarah-artificerinnovations (approved):
All blocking issues resolved and the non-blocking items addressed.

**M1 — pagination:** `findAuthUserByEmail.ts` paginates with `perPage: 1000` and breaks when the last page is partial. Both CLIs use it — no more silent first-page-only search.

**M2 — granted_by:** `--granted-by <email>` flag resolves the granter's UUID and sets it on upsert. Correctly documented in SECURITY.md that CLI grants without the flag leave it null. Good tradeoff — service role has no UUID, so opt-in attribution is the right model.

**M3 — search cap:** Capped at 200 chars via `left(..., 200)` and ILIKE metacharacters (`%`, `_`, `\`) are now escaped before the pattern is built. Better than requested.

**M4 — stale cache:** "Client route guard cache" section added to `packages/admin/SECURITY.md` — correctly distinguishes immediate data-RPC revocation from the UI shell staying open until remount.

**pgTAP:** All three gaps filled — revoked-admin-is-denied, unauthenticated session, and two new `admin_get_user` assertions. Plan count 15 → 20.

**Bonus:** `revoke-admin.ts` now checks that the update actually matched a row and exits with a clear error if the user was already revoked or never granted.

Approving.

ZappoMan:
@diego-artificerinnovations Re: item **#3 (`to_jsonb(*.*)` in `admin_get_user`)** — addressed in **7f0ddcb**.

New migration `20260518140000_admin_get_user_column_projection.sql` replaces every `to_jsonb(table.*)` with explicit `jsonb_build_object` allowlists:

| Block | Exposed fields (summary) | Omitted (examples) |
|-------|--------------------------|--------------------|
| `profile` | display_name, username, avatar_url, bio, website, location, created/updated_at | `user_id`, internal `id` |
| `subscription` | plan_id, status, period/trial/cancel fields | `stripe_customer_id`, `stripe_subscription_id`, `stripe_price_id` |
| `plan` | id, display_name, pricing, features, usage_limits | Stripe price/product ids |
| `usage_aggregates` | event_type, period, count | — |
| `usage_events` | event_type, quantity, created_at | `metadata`, `idempotency_key` |
| `invoices` | id, stripe_invoice_id, amounts, status, hosted URL | `stripe_customer_id`, raw PDF URL kept only as hosted link |

pgTAP now asserts `profile` has no `user_id`, `subscription` has no `stripe_customer_id`, and usage events omit `metadata` when present.

`supabase test db` passes locally (plan 23). Happy to re-review when CI is green.

rahul-artificerinnovations (approved):
Verified the `to_jsonb` fix. `admin_get_user` now uses explicit `jsonb_build_object` projections for all four tables — `auth.users`, `user_profiles`, `billing_subscriptions`, `billing_usage_events`, and `billing_invoices`. Stripe customer ID, raw metadata, `user_id` redundancy, and any other internal columns are all excluded by construction. The three new pgTAP assertions (`profile.user_id`, `subscription.stripe_customer_id`, `usage_events.metadata`) lock this in as a regression test.

Nice touches: audit call moved to after the not-found guard (no audit row on phantom user lookups), `LIMIT 100` / `LIMIT 20` caps on usage events and invoices, REVOKE/GRANT block consistent with the other admin functions.

All issues from both reviews resolved. Approving.

diego-artificerinnovations (approved):
All six `to_jsonb(table.*)` calls are now explicit `jsonb_build_object` allowlists. `stripe_customer_id` and `stripe_subscription_id` are excluded from the subscription block; `user_id` excluded from the profile block; internal metadata excluded from events. Three new pgTAP assertions lock in the omission of those fields. The `v_plan_id` refactor is cleaner than the previous `v_subscription ->> 'plan_id'` string extraction. This was the last open item — approved.

mei-artificerinnovations (approved):
The `to_jsonb` fix in `20260518140000_admin_get_user_column_projection.sql` is clean and complete. All six wildcard serializations replaced with explicit `jsonb_build_object` projections. A few things I spot-checked:

- `billing_subscriptions` correctly **excludes** the Stripe subscription ID — admin operators don't need it directly.
- `billing_invoices` correctly **includes** `stripe_invoice_id` — operators need it to look up invoices in Stripe.
- The `billing_usage_events` inner subquery also names columns explicitly (line 141) — belt-and-suspenders, right call.
- `v_plan_id` is now extracted directly from the row instead of via JSON string dereference (`v_subscription ->> 'plan_id'`), which is cleaner and avoids a subtle type-casting edge case.

All five original findings addressed. Good to merge.

PR #274, opened 2026-05-18
feat(waitlist): signup gating, invites, capture, and admin workflows

ZappoMan:
## Summary
- Add waitlist v1: Supabase schema, RPCs, rate limits, and Edge Functions (`waitlist-capture`, `waitlist-ops`) for capture, invite validation/consumption, and admin actions.
- Introduce `@beakerstack/waitlist` and `@beakerstack/email` packages with signup mode gating, waitlist form, invite signup page, and admin waitlist/settings UI in the web app.
- Wire public signup through `SignupModeGate`, invite flow via `/signup/invite` and OAuth callback finalization, plus `billing_ensure_subscription_plan` on conversion.

## Test plan
- [ ] `supabase db reset` and `supabase test db` (includes `waitlist.test.sql`)
- [ ] `npm run test:unit:waitlist` and `npm run test:unit:email`
- [ ] `npm run test:coverage` (waitlist/email included in merge report)
- [ ] `npm run type-check`
- [ ] Manual: set signup mode in admin waitlist settings; exercise waitlist capture, approve/reject, invite link signup (email + Google)
- [ ] Deploy Edge Functions and confirm `waitlist-setup.md` env/secrets for a preview environment


package.json
```diff
@@ -77,24 +77,28 @@
...
     "test:e2e:mobile": "export PATH=\"$PATH:$HOME/.maestro/bin\" && E2E_PW=\"${TEST_PASSWORD:-E2e_$(openssl rand -hex 16)_Aa1}\" && maestro test tests/e2e/mobile/flows/home.yaml --env MOBILE_APP_ID=\"${MOBILE_APP_ID:-com.anonymous.beakerstack}\" --env TEST_EMAIL=\"${TEST_EMAIL:-e2e-test-$(date +%s)@example.com}\" --env TEST_PASSWORD=\"$E2E_PW\"",
     "test:watch": "concurrently \"cd apps/mobile && npm run test:watch\" \"cd apps/web && npm run test:watch\" \"cd packages/shared-tests && npm run test:web -- --watch\" \"cd packages/shared-tests && npm run test:native -- --watch\"",
-    "test:coverage": "npm run test:coverage:mobile && npm run test:coverage:web && npm run test:coverage:shared && npm run test:coverage:billing && npm run test:coverage:admin && npm run test:coverage:merge",
+    "test:coverage": "COVERAGE_MERGE=1 npm run test:coverage:mobile && npm run test:coverage:web && npm run test:coverage:shared && npm run test:coverage:billing && npm run test:coverage:admin && npm run test:coverage:waitlist && npm run test:coverage:email && npm run test:coverage:merge",
```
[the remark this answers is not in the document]
ZappoMan: Fixed: test:coverage now uses export COVERAGE_MERGE=1 so the variable applies to the entire chain.

apps/mobile/package.json
```diff
@@ -110,8 +110,6 @@
     "coverageDirectory": "coverage",
     "coverageReporters": [
       "text",
-      "lcov",
-      "html",
       "json"
```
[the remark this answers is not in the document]
ZappoMan: Restored lcov and html reporters in apps/mobile/package.json for local coverage runs; the merge pipeline still trims reporters via COVERAGE_MERGE in vitest configs.

supabase/migrations/20260519000000_waitlist_v1.sql
```diff
@@ -0,0 +1,797 @@
...
+        );
+    END IF;
+
+    PERFORM public._admin_insert_audit(
+        v_uid,
+        'waitlist.entries.list',
+        'waitlist_entries',
+        NULL,
+        jsonb_build_object('limit', v_limit, 'offset', v_offset, 'status', p_status)
+    );
```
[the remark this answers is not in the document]
ZappoMan: Removed _admin_insert_audit from admin_list_waitlist_entries so list pagination/search no longer floods the audit log.

supabase/migrations/20260519000000_waitlist_v1.sql
```diff
@@ -0,0 +1,797 @@
...
+    IF NOT public.is_valid_email(v_email) THEN
+        -- Uniform success — do not leak validation details
+        SELECT * INTO v_settings FROM public._waitlist_settings_row();
+        v_copy := COALESCE(v_settings.copy, '{}'::jsonb);
+        v_confirm := COALESCE(
+            v_copy #>> '{waitlist,confirmation}',
+            'Thanks — you''re on the list.'
+        );
+        RETURN jsonb_build_object('ok', true, 'message', v_confirm);
+    END IF;
```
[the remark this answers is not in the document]
ZappoMan: Reordered waitlist_capture to apply the IP rate limit before the email validity check.

supabase/functions/waitlist-ops/index.ts
```diff
@@ -0,0 +1,199 @@
...
+    const { data, error } = await admin.rpc('waitlist_consume_invite', {
+      p_token: token,
+      p_user_id: userId,
+      p_user_email: body.userEmail ?? null,
+    });
+    if (error) {
+      return jsonResponse({ error: error.message }, 400, req);
+    }
+    return jsonResponse(data, 200, req);
+  }
```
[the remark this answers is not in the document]
ZappoMan: consume now requires Authorization, verifies user.id === body.userId, and calls waitlist_consume_invite with the authenticated client instead of service role.

packages/waitlist/src/components/WaitlistForm.web.tsx
```diff
@@ -0,0 +1,166 @@
...
+                      onMetadataChange(
+                        field.id,
+                        e.target.value,
+                        metadata,
+                        setMetadata
+                      )
+                    }
+                    className='w-full px-3 py-2 border border-gray-300 dark:border-gray-600 rounded-md bg-white dark:bg-gray-800'
+                  />
+                )}
```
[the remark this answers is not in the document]
ZappoMan: Added htmlFor/id on metadata field labels and inputs.

packages/waitlist/src/components/WaitlistForm.web.tsx
```diff
@@ -0,0 +1,166 @@
...
+      </label>
+      <input
+        type='email'
+        autoComplete='email'
+        value={email}
+        onChange={e => setEmail(e.target.value)}
+        className='w-full px-3 py-2 border border-gray-300 dark:border-gray-600 rounded-md bg-white dark:bg-gray-800 text-gray-900 dark:text-gray-100'
+        placeholder='you@example.com'
+        disabled={loading}
+      />
```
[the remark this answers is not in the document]
ZappoMan: Added htmlFor=waitlist-email and id=waitlist-email on the email label/input.

packages/waitlist/src/lifecycle.ts
```diff
@@ -0,0 +1,36 @@
...
+export async function emitLifecycleEvent(
+  event: LifecycleEventName,
+  payload: LifecycleEventPayload
+): Promise<void> {
+  await Promise.all(
+    [...listeners].map(async fn => {
+      await fn(event, payload);
+    })
+  );
+}
```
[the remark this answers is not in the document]
ZappoMan: Switched to Promise.allSettled with per-listener error logging so listener failures cannot abort core flows.

apps/web/src/pages/SignupInvitePage.tsx
```diff
@@ -0,0 +1,225 @@
...
+        { body: { action: 'validate', token: t } }
+      );
+      if (fnErr || !(data as { valid?: boolean })?.valid) {
+        setInvalid(true);
+      } else {
+        setInviteEmail((data as { email?: string }).email ?? null);
+      }
+      setLoading(false);
+    })();
+  }, []);
```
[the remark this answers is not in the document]
ZappoMan: Added a useRef guard so invite validate runs once per mount (avoids duplicate calls under React StrictMode in dev).

supabase/functions/waitlist-ops/index.ts
```diff
@@ -0,0 +1,219 @@
...
+      body.subject ??
+      Deno.env.get('WAITLIST_INVITE_SUBJECT') ??
+      'You are invited to sign up';
+    const htmlTemplate =
+      body.html ??
+      Deno.env.get('WAITLIST_INVITE_HTML') ??
+      '<p>Complete your signup: <a href="{{inviteUrl}}">{{inviteUrl}}</a></p>';
+    const html = renderTemplate(htmlTemplate, { inviteUrl });
+
+    console.log(
```
sarah-artificerinnovations: **M1 (blocking) — `send_invite_email` is a no-op stub; invites will never be delivered**

The action accepts `email`, `inviteUrl`, subject, and HTML, renders the template — then logs it and returns `{ ok: true }`. No email is sent, and the caller has no way to detect the failure.

Admin operators who click "approve + send invite" will believe the email was delivered. In practice, waitlist users will never receive their invite links and have no way to sign up.

Options:
1. **Wire in Resend (or Supabase Edge Functions' built-in SMTP)** — Resend has a zero-dependency Deno-compatible API. The `WAITLIST_RESEND_API_KEY` env var (already referenced in `waitlist-setup.md`) just needs to be used here.
2. **Remove the action and document the email step as a webhook** — emit a lifecycle event (or call a deployer-supplied webhook URL) so callers implement their own email delivery. This matches the `@beakerstack/email` package pattern.

Either way, the current state should not ship — operators will silently lose every invite.

apps/web/src/pages/SignupInvitePage.tsx
```diff
@@ -0,0 +1,230 @@
...
+      body: {
+        action: 'consume',
+        token,
+        userId,
+        userEmail: userEmail ?? null,
+      },
+    }
+  );
+  if (error) throw new Error(error.message);
+  const body = data as { error?: string; default_plan_id?: string };
```
sarah-artificerinnovations: **M2 (blocking) — `billing_ensure_subscription_plan` is called even when the entry is `already_converted`**

`waitlist_consume_invite` returns `{ ok: true, already_converted: true }` when the entry has already been converted (idempotent guard). But `finalizeInviteSignup` casts the response to `{ error?: string; default_plan_id?: string }` and discards the `already_converted` flag — so `billing_ensure_subscription_plan` runs unconditionally.

`billing_ensure_subscription_plan` uses `ON CONFLICT DO UPDATE SET plan_id = EXCLUDED.plan_id, status = EXCLUDED.status`. This means a user who re-triggers the OAuth callback with a stale `sessionStorage` token (browser back-button after Google OAuth, for instance) will have their subscription silently overwritten with the default free plan, even if they have since upgraded to a paid tier.

Fix: check for `already_converted` before calling billing:

```ts
const body = data as { error?: string; already_converted?: boolean; default_plan_id?: string };
if (body?.error) throw new Error(body.error);
if (body?.already_converted) {
  sessionStorage.removeItem(INVITE_TOKEN_STORAGE_KEY);
  return; // already done, skip billing + lifecycle
}
```

supabase/migrations/20260519010000_billing_ensure_subscription_plan.sql
```diff
@@ -0,0 +1,60 @@
...
+        v_uid,
+        p_product_id,
+        p_plan_id,
+        CASE WHEN v_plan.price_cents = 0 THEN 'free' ELSE 'active' END,
+        NULL,
+        NULL,
+        date_trunc('month', now() AT TIME ZONE 'utc'),
+        date_trunc('month', now() AT TIME ZONE 'utc') + interval '1 month'
+    )
+    ON CONFLICT (user_id, product_id) DO UPDATE
```
sarah-artificerinnovations, line 45: **M3 (non-blocking) — `ON CONFLICT DO UPDATE` can silently downgrade a paid subscription**

Any call to `billing_ensure_subscription_plan` — even one that shouldn't be the final word on a user's plan — will unconditionally overwrite `plan_id` and `status`. If this RPC is ever called again outside the invite conversion path (admin tooling, future feature, retry), it will silently move a paying user to whatever plan is passed in.

Suggest adding a guard so the upsert only changes the plan if the user doesn't already have a paid subscription:

```sql
ON CONFLICT (user_id, product_id) DO UPDATE
SET
    plan_id = EXCLUDED.plan_id,
    status  = EXCLUDED.status,
    updated_at = now()
WHERE billing_subscriptions.status NOT IN ('active', 'trialing', 'past_due');
```

This makes the function safe to call idempotently for free-plan provisioning without risking paid-plan regression.

supabase/migrations/20260519000000_waitlist_v1.sql
```diff
@@ -0,0 +1,783 @@
...
+    v_copy := COALESCE(v_settings.copy, '{}'::jsonb);
+    v_confirm := COALESCE(
+        v_copy #>> '{waitlist,confirmation}',
+        'Thanks — you''re on the list.'
+    );
+
+    v_window := date_trunc('hour', now());
+
+    IF p_client_ip IS NOT NULL AND length(trim(p_client_ip)) > 0 THEN
+        INSERT INTO public.waitlist_rate_limits (bucket_key, window_start, count)
```
sarah-artificerinnovations, line 380: **nit — `waitlist_rate_limits` rows accumulate indefinitely**

Rate limit buckets are never deleted. Every distinct `(ip, hour)` and `(email, hour)` pair lives in the table forever. On a busy signup form this will grow without bound.

Add a pruning step at the top of `waitlist_capture` (or schedule a periodic cleanup):

```sql
DELETE FROM public.waitlist_rate_limits
WHERE window_start < now() - interval '2 hours';
```

Running it inline is cheap — it only hits the `idx_waitlist_rate_limits_window` index and removes rows older than the current window.

supabase/migrations/20260519000000_waitlist_v1.sql
```diff
@@ -0,0 +1,783 @@
...
+    v_confirm text;
+BEGIN
+    SELECT * INTO v_settings FROM public._waitlist_settings_row();
+    v_copy := COALESCE(v_settings.copy, '{}'::jsonb);
+    v_confirm := COALESCE(
+        v_copy #>> '{waitlist,confirmation}',
+        'Thanks — you''re on the list.'
+    );
+
+    v_window := date_trunc('hour', now());
```
sarah-artificerinnovations, line 377: **nit — fixed-window rate limiting allows burst at the hour boundary**

`date_trunc('hour', now())` as the window key resets at exactly :00:00. An attacker can submit 10 times in the last minute of an hour and 10 more in the first minute of the next — 20 attempts in under two minutes — while staying within the limit.

A sliding window (e.g., track `date_trunc('minute', now())` and sum across the last 60 buckets) would prevent this. Non-blocking given the 10/hour ceiling, but worth revisiting if abuse patterns emerge.

supabase/migrations/20260519000000_waitlist_v1.sql
```diff
@@ -0,0 +1,783 @@
...
+DECLARE
+    v_uid uuid := auth.uid();
+BEGIN
+    IF v_uid IS NULL OR NOT public.admin_is_admin() THEN
+        RETURN jsonb_build_object('error', 'not_found');
+    END IF;
+
+    UPDATE public.waitlist_settings
+    SET
+        signup_mode = COALESCE(p_signup_mode, signup_mode),
```
sarah-artificerinnovations, line 476: **nit — `admin_update_waitlist_settings` doesn't validate enum inputs at the application level**

If `p_signup_mode` is an invalid value (typo in a script, wrong client version), the `UPDATE` fails with a raw PostgreSQL CHECK constraint violation instead of a clean `{ error: 'invalid_signup_mode' }` JSON response. The caller gets a 400 from Supabase with an internal error message exposed in `error.message`.

Suggest adding explicit guards at the top of the function:

```sql
IF p_signup_mode IS NOT NULL
   AND p_signup_mode NOT IN ('open', 'waitlist', 'invite_only', 'closed') THEN
    RETURN jsonb_build_object('error', 'invalid_signup_mode');
END IF;
```

Same pattern for `p_identity_match_mode`. Makes the API surface predictable to callers.

sarah-artificerinnovations (changes requested):
Triage: Large / public signup surface + invite flow + billing integration.

The schema design is solid — RLS with no direct table access, SECURITY DEFINER RPCs with `admin_is_admin()` on every admin operation, token stored as SHA-256 hash, invite URLs using hash fragments. The `@beakerstack/waitlist` package boundary is clean. Two blocking issues and a few nits; inline comments have details.

---

**M1 (blocking) — `send_invite_email` is a no-op stub**

`waitlist-ops/index.ts:211` renders the email template then logs it to `console.log` and returns `{ ok: true }`. No email is ever sent. Admins who approve + send an invite will believe it was delivered; waitlist users will receive nothing and have no way to sign up. The `@beakerstack/email` package exists and `WAITLIST_RESEND_API_KEY` is already called out in `waitlist-setup.md` — the integration just needs to happen here, or the action needs to be reframed as a webhook/lifecycle event that deployers wire up themselves. Either way the current behavior must not ship.

**M2 (blocking) — `billing_ensure_subscription_plan` called unconditionally, can downgrade paid users**

`finalizeInviteSignup` (SignupInvitePage.tsx:38) ignores the `already_converted: true` flag returned by `waitlist_consume_invite` and always calls `billing_ensure_subscription_plan`. That RPC's `ON CONFLICT DO UPDATE` overwrites `plan_id` and `status` unconditionally. A user who re-triggers the callback with a stale `sessionStorage` token (Google OAuth back-button, reopened tab) will have their subscription silently overwritten with the default free plan. Fix: skip billing entirely when `already_converted` is true.

---

Non-blocking:

**M3 — `billing_ensure_subscription_plan` ON CONFLICT can downgrade paid subscriptions from any caller**

Even after M2 is fixed in the client, the RPC itself will overwrite any plan on conflict. A WHERE clause on the DO UPDATE (`WHERE status NOT IN ('active', 'trialing', 'past_due')`) would make it safe to call idempotently without paid-plan regression. See inline comment on the billing migration.

---

Nits:
- `waitlist_rate_limits` rows accumulate forever — add a pruning DELETE at the top of `waitlist_capture` (migration line 380)
- Fixed-window rate limit (date_trunc hour) allows 2× burst at the hour boundary (migration line 377)
- `admin_update_waitlist_settings` should validate enum inputs and return clean JSON errors instead of raw PG constraint violations (migration line 476)

Requesting changes on M1 + M2.

mei-artificerinnovations (changes requested):
## Code review — feat/waitlist-v1

Overall architecture is solid: SHA-256 hashed tokens, FOR UPDATE locks on consume, uniform success responses throughout (no enumeration), RLS enabled on all four new tables with zero client-accessible policies. Inline comments below for issues that need attention before merge.

**2 HIGH · 3 MEDIUM · 2 LOW**

---

### HIGH-1 · `send_invite_email` — arbitrary `to` address + admin-supplied HTML template

`waitlist-ops/index.ts` lines 275-295: the action accepts `body.email` (any address, not tied to a waitlist entry), `body.html` (admin-supplied HTML template), and `body.inviteUrl` (admin-supplied URL). `renderTemplate` renders the HTML, and when real email is wired in, it will be sent.

This means any admin can: (1) target any email address, (2) supply arbitrary HTML as the body, (3) supply an arbitrary URL as the invite link. That turns the platform's email infrastructure into a phishing/spam vector. The stub (`console.log`) masks the risk now, but the design needs to be fixed before email is wired in.

Fix: require `entryId`, look up the entry server-side to get the `to` address, and use a server-side-only template. Remove `html`, `subject`, and `email` from the request body.

### HIGH-2 · `AuthCallbackPage` — invite error path shows "Redirecting to login page…" but never redirects

`AuthCallbackPage.tsx` lines 52-56: when `finalizeInviteSignup` throws, `.catch()` calls `setError()` but the error UI template unconditionally renders "Redirecting to login page..." — however, no redirect is actually scheduled in this path. The redirect timer is only set in the `errorParam` branch (lines 33-37). The user sees the message and sits there indefinitely.

Fix: add a timed redirect in the `.catch()` block, using `delayedLoginTimerRef` the same way the `errorParam` branch does.

---

### MEDIUM-1 · `waitlist-capture` — `metadata` payload has no size guard

`waitlist-capture/index.ts` line 62: `body.metadata ?? {}` is passed to the RPC without a size check. A caller can POST a very large JSON object (up to the Edge Function's body size limit) and have it stored in `waitlist_entries.metadata`. Add an explicit cap before the RPC call — e.g., reject if `JSON.stringify(meta).length > 4096`.

### MEDIUM-2 · `admin_update_waitlist_settings` audit log is incomplete

Migration lines 486-495: the audit log only captures `signup_mode` and `default_plan_id`. Changes to `identity_match_mode`, `invite_ttl_days`, `copy`, and `metadata_schema` go unlogged. `identity_match_mode` is security-sensitive — flipping it from `strict` to `lenient` changes whether email mismatches are blocked or merely logged. That change should always appear in the audit trail.

### MEDIUM-3 · `AuthCallbackPage` setTimeout fallback doesn't check invite token

`AuthCallbackPage.tsx` lines 70-86: the 1200ms fallback timer navigates directly via `readAndClearPostAuthRedirect()` without checking `INVITE_TOKEN_STORAGE_KEY`. If Google OAuth token exchange takes longer than 1200ms (slow network, cold edge), the timer fires before the React re-render triggered by auth state resolving — the invite token is silently skipped and the user lands on `/dashboard` with an unconsumed invite. The `auth.user` path (lines 42-62) checks the token correctly; the timer path should mirror it.

---

### LOW-1 · `waitlist_rate_limits` accumulates without cleanup

The table grows indefinitely — there's no scheduled `DELETE FROM waitlist_rate_limits WHERE window_start < now() - interval '24 hours'`. Low urgency now but will need a pg_cron job or scheduled Edge Function under any real volume.

### LOW-2 · `SignupInvitePage` — no client-side minimum password length

`SignupInvitePage.tsx` lines 108-111: only validates password match, not length. Supabase will reject short passwords, but the error surfaces as a generic thrown-error message. Add explicit `minLength` validation before calling `auth.signUp` for a consistent UX.

---

### Positive notes

- Token design is correct: 32 bytes from `gen_random_bytes`, stored as SHA-256 hash — not reversible, not enumerable at any realistic scale.
- `FOR UPDATE` on both the invite and entry rows in `waitlist_consume_invite` correctly prevents double-consume under concurrent requests.
- ILIKE search in `admin_list_waitlist_entries` correctly escapes `%` and `_` — the lesson from PR #261 is applied here from day one.
- `requireAdmin` in `waitlist-ops` correctly verifies the JWT user and then confirms admin status via the `admin_is_admin()` RPC before acting on any privileged action.
- Uniform success responses throughout capture — no enumeration of whether an email is already registered.

supabase/migrations/20260519010000_billing_ensure_subscription_plan.sql
```diff
@@ -0,0 +1,60 @@
...
+    SELECT * INTO r
+    FROM public.billing_subscriptions s
+    WHERE s.user_id = v_uid AND s.product_id = p_product_id;
+
+    RETURN r;
+END;
+$$;
+
+REVOKE ALL ON FUNCTION public.billing_ensure_subscription_plan(text, text) FROM PUBLIC;
+GRANT EXECUTE ON FUNCTION public.billing_ensure_subscription_plan(text, text) TO authenticated, service_role;
```
diego-artificerinnovations, line 60: **security: privilege escalation — any authenticated user can self-grant a paid plan**

`billing_ensure_subscription_plan` is grantable to `authenticated`, and it's called directly from the browser in `SignupInvitePage.finalizeInviteSignup`. That means any logged-in user can call:

```js
await supabase.rpc('billing_ensure_subscription_plan', {
  p_product_id: 'beakerstack',
  p_plan_id: 'beakerstack_pro',   // any valid plan ID
})
```

For a paid plan (`price_cents > 0`), the function inserts `status = 'active'` with `stripe_customer_id = NULL` — a Stripe-unbacked premium subscription. An existing free subscriber can upgrade themselves to any paid tier for free.

The fix is to restrict the grant to `service_role` only and call the RPC from the `consume` action in `waitlist-ops` (which already uses the service-role client) rather than from the browser:

```sql
-- migration: remove authenticated grant
REVOKE EXECUTE ON FUNCTION public.billing_ensure_subscription_plan(text, text) FROM authenticated;
```

```ts
// waitlist-ops: consume action — after admin.rpc('waitlist_consume_invite') succeeds:
if (consumeResult.ok && consumeResult.default_plan_id) {
  await admin.rpc('billing_ensure_subscription_plan', {
    p_product_id: body.productId,
    p_plan_id: consumeResult.default_plan_id,
  });
}
```

This keeps the plan provisioning entirely server-side, out of reach of browser callers.

supabase/migrations/20260519000000_waitlist_v1.sql
```diff
@@ -0,0 +1,783 @@
...
+-- Admin RPCs
+-- ---------------------------------------------------------------------------
+
+CREATE OR REPLACE FUNCTION public.admin_get_waitlist_settings()
+RETURNS jsonb
+LANGUAGE plpgsql
+STABLE
+SECURITY DEFINER
+SET search_path = public
+AS $$
```
diego-artificerinnovations, line 428: **security nit: `admin_get_waitlist_settings` marked STABLE but calls VOLATILE `admin_is_admin()`**

PostgreSQL doesn't reject this at `CREATE FUNCTION` time, but a STABLE function is eligible for result-caching within a query. If `admin_is_admin()` is cached (VOLATILE prevents this on its own, but the outer STABLE wrapper re-introduces the risk in certain query contexts), the admin check could be bypassed for a caller whose privileges changed mid-session.

`admin_get_waitlist_entry` (line 594) has the same problem — marked STABLE, calls VOLATILE `admin_is_admin()`.

Brad already fixed this exact issue for `admin_is_admin()` itself in PR #261 (`STABLE → VOLATILE`). Apply the same fix here:

```sql
CREATE OR REPLACE FUNCTION public.admin_get_waitlist_settings()
RETURNS jsonb
LANGUAGE plpgsql
VOLATILE  -- not STABLE: calls VOLATILE admin_is_admin()
SECURITY DEFINER
SET search_path = public
AS $$
```

supabase/functions/waitlist-ops/index.ts
```diff
@@ -0,0 +1,219 @@
...
+      `[beakerstack/email] waitlist invite to=${to} subject=${subject}\n${html}`
+    );
+
+    return jsonResponse({ ok: true }, 200, req);
```
diego-artificerinnovations, line 290: nit: `send_invite_email` always returns `{ ok: true }` without sending any email — the body is console-logged only.

`waitlist-setup.md` documents this for local dev ("copy invite link from drawer (also logged in function output)"), but in a production environment an admin clicking "Send invite email" gets a success response and no email arrives. The `@beakerstack/email` package only ships a `logAdapter`; there's no production adapter wired up.

Two options:
1. Return HTTP 501 with `{ error: 'email_not_configured' }` when no real adapter is present, so the admin UI can surface a meaningful message instead of silently succeeding.
2. Add a `WAITLIST_EMAIL_PROVIDER` env var (e.g. `resend`, `sendgrid`) and a production adapter in `@beakerstack/email`.

As shipped, the admin panel's "Send invite" button is effectively a no-op in any hosted environment without additional wiring that isn't documented.

supabase/migrations/20260519000000_waitlist_v1.sql
```diff
@@ -0,0 +1,783 @@
...
+        updated_by = v_uid
+    WHERE id = 1;
+
+    PERFORM public._admin_insert_audit(
+        v_uid,
+        'waitlist.settings.update',
+        'waitlist_settings',
+        '1',
+        jsonb_build_object(
+            'signup_mode', p_signup_mode,
```
diego-artificerinnovations, line 492: nit: the audit log only records `signup_mode` and `default_plan_id` — but `identity_match_mode` is the most security-relevant field here. Changing from `strict` to `lenient` permits OAuth signups with a different email than the invite address; that's worth capturing.

```sql
-- suggested: log all mutable fields
jsonb_build_object(
    'signup_mode', p_signup_mode,
    'default_plan_id', p_default_plan_id,
    'invite_ttl_days', p_invite_ttl_days,
    'identity_match_mode', p_identity_match_mode
)
```

apps/web/src/pages/AuthCallbackPage.tsx
```diff
@@ -37,6 +41,21 @@ export default function AuthCallbackPage() {
...
+          inviteToken,
+          auth.user.id,
+          auth.user.email ?? undefined
+        )
+          .then(() => navigate('/dashboard', { replace: true }))
+          .catch(() => {
+            setError(
+              'Could not complete invite signup. Try the invite link again.'
+            );
+          });
```
diego-artificerinnovations: nit: if `finalizeInviteSignup` rejects, `setError(...)` is called but no navigation ever fires. The user is stuck on the callback spinner with an error banner and no way out except manually typing a URL.

Add a fallback navigate to `'/login'` after showing the error:

```tsx
.catch(() => {
  setError('Could not complete invite signup. Try the invite link again.');
  // Give the user a moment to read the message, then redirect
  setTimeout(() => {
    navigatedRef.current = true;
    navigate('/login', { replace: true });
  }, 3000);
})
```

supabase/migrations/20260519000000_waitlist_v1.sql
```diff
@@ -0,0 +1,783 @@
...
+ALTER TABLE public.waitlist_entries
+    ADD CONSTRAINT waitlist_entries_active_invite_fkey
+    FOREIGN KEY (active_invite_id) REFERENCES public.waitlist_invites (id) ON DELETE SET NULL;
+
+CREATE TABLE IF NOT EXISTS public.waitlist_rate_limits (
+    bucket_key text NOT NULL,
+    window_start timestamptz NOT NULL,
+    count integer NOT NULL DEFAULT 1,
+    PRIMARY KEY (bucket_key, window_start)
+);
```
diego-artificerinnovations, line 77: nit: `waitlist_rate_limits` rows are never deleted. One `(bucket_key, window_start)` row is inserted per unique IP-hour and email-hour; on a product with meaningful traffic this table will grow indefinitely.

A minimal cleanup approach — add to the capture RPC or a scheduled job:

```sql
-- prune rows older than 24 hours at the top of waitlist_capture:
DELETE FROM public.waitlist_rate_limits WHERE window_start < now() - interval '24 hours';
```

Or use `pg_cron` if available. Without cleanup the table doubles as a permanent record of all capture traffic, which may not be desirable from a GDPR/retention standpoint either.

supabase/tests/waitlist.test.sql
```diff
@@ -0,0 +1,116 @@
...
+SELECT is(
+    public.admin_list_waitlist_entries()->>'error',
+    'not_found',
+    'non-admin list returns not_found'
+);
+
+RESET role;
+
+SELECT * FROM finish();
+ROLLBACK;
```
diego-artificerinnovations, line 193: nit: the pgTAP plan is 15, covering schema existence, RLS, capture deduplication, and non-admin gating. Missing cases worth adding for a security-sensitive feature:

```sql
-- 1. waitlist_validate_invite: valid token → valid=true, entry.email returned
-- 2. waitlist_consume_invite: first call succeeds; second call for same token returns error
-- 3. admin_approve_waitlist_entry: returns invite_token (non-null), sets entry status='approved'
-- 4. admin_reject_waitlist_entry: revokes active invite, sets status='rejected'
-- 5. waitlist_consume_invite after reject: invalid_invite
```

The consume-twice case in particular is the core security invariant of the invite system — it should be tested at the DB layer even if the Edge Function also enforces it.

diego-artificerinnovations (changes requested):
## Review — waitlist v1

Solid architecture overall: RLS-opaque tables, SECURITY DEFINER + `SET search_path`, REVOKE-then-GRANT pattern, SHA-256 token hashing (not plaintext), `FOR UPDATE` locking in `waitlist_consume_invite`, uniform non-enumerating responses in `waitlist_capture`, and the escape-hardened ILIKE search copied from the admin list pattern. Two security issues below — one is blocking, the other should be fixed before this ships. Three nits after that.

supabase/migrations/20260519000000_waitlist_v1.sql
```diff
@@ -0,0 +1,783 @@
...
+    v_valid := public.waitlist_validate_invite(p_token);
+    IF NOT COALESCE((v_valid->>'valid')::boolean, false) THEN
+        RETURN jsonb_build_object('error', 'invalid_invite');
+    END IF;
+
+    v_entry_id := (v_valid->>'entry_id')::uuid;
+    v_hash := public._waitlist_hash_token(trim(p_token));
+
+    SELECT * INTO v_inv FROM public.waitlist_invites WHERE token_hash = v_hash FOR UPDATE;
+    SELECT * INTO v_entry FROM public.waitlist_entries WHERE id = v_entry_id FOR UPDATE;
```
rahul-artificerinnovations, line 290: **[HIGH] Missing re-check of `used_at` after acquiring the row lock.**

Two concurrent calls can both pass `waitlist_validate_invite` (which reads `used_at IS NULL` without holding any lock) before either acquires `FOR UPDATE`. After the lock serializes them, the code checks `v_entry.status` but never re-verifies `v_inv.used_at IS NULL`. Add a guard here:

```sql
SELECT * INTO v_inv FROM public.waitlist_invites WHERE token_hash = v_hash FOR UPDATE;
-- ADD THIS:
IF v_inv.used_at IS NOT NULL OR v_inv.revoked_at IS NOT NULL THEN
    RETURN jsonb_build_object('error', 'invite_already_used');
END IF;
```

The entry-status check below provides partial protection but is not sufficient in all edge cases (e.g. invite row updated out-of-band).

apps/web/src/pages/SignupPage.tsx
```diff
@@ -162,103 +164,108 @@ function SignupPageContent() {
...
-            <div className='relative'>
-              <div className='absolute inset-0 flex items-center'>
-                <div className='w-full border-t border-gray-300 dark:border-gray-600' />
-              </div>
-              <div className='relative flex justify-center text-sm'>
-                <span className='px-2 bg-gray-50 dark:bg-gray-900 text-gray-500 dark:text-gray-400'>
-                  Or continue with email
-                </span>
+            <SignupModeGate
+              supabase={supabase}
```
rahul-artificerinnovations, line 192: **[HIGH] `SignupModeGate` is client-side only — no server-side enforcement.**

This component hides the signup form in the browser, but `supabase.auth.signUp()` can still be called directly from the browser console, curl, or any HTTP client. The waitlist gate needs to be enforced at the API layer (e.g. a Supabase `before_sign_in` auth hook that rejects signups when mode is `'waitlist'`). As-is, the gating is cosmetic.

supabase/functions/waitlist-capture/index.ts
```diff
@@ -0,0 +1,75 @@
...
+type Body = {
+  email?: string;
+  metadata?: Record<string, unknown>;
+  honeypot?: string;
+};
+
+function clientIp(req: Request): string | null {
+  const forwarded = req.headers.get('x-forwarded-for');
+  if (forwarded) return forwarded.split(',')[0]?.trim() ?? null;
+  return req.headers.get('x-real-ip');
```
rahul-artificerinnovations: **[HIGH] X-Forwarded-For spoofing.**

`forwarded.split(',')[0]` takes the *leftmost* IP, which is attacker-controlled. An attacker can set `X-Forwarded-For: 1.1.1.1` to bypass rate limiting entirely. Use the *rightmost* segment (appended by infrastructure) or a guaranteed header from the edge runtime:

```ts
// Rightmost segment is the one appended by your infrastructure
if (forwarded) return forwarded.split(',').at(-1)?.trim() ?? null;
```

supabase/functions/_shared/waitlist-origins.ts
```diff
@@ -0,0 +1,64 @@
...
+export function corsHeadersForWaitlist(req: Request): HeadersInit {
+  const origin = req.headers.get('Origin');
+  const allowed = getWaitlistAllowedOrigins();
+  const allowHeaders =
+    'authorization, x-client-info, apikey, content-type, x-supabase-client';
+
+  if (!origin) {
+    return {
+      'Access-Control-Allow-Origin': '*',
+      'Access-Control-Allow-Headers': allowHeaders,
```
rahul-artificerinnovations, line 33: **[HIGH] CORS wildcard returned when `Origin` is absent.**

Requests without an `Origin` header (curl, server-to-server, scripts) receive `Access-Control-Allow-Origin: *`. This is unnecessary for non-browser requests and could expose error response bodies to cross-origin reads on clients that send credentialless requests. When `!origin`, return no ACAO header at all rather than `'*'`.

rahul-artificerinnovations (changes requested):
## Code Review — PR #274: feat(waitlist): signup gating, invites, capture, and admin workflows

Overall this is a substantial and well-structured feature. The schema design is solid, the edge functions are cleanly separated, and the pgTAP coverage is a good start. However there are four HIGH-severity security issues and several MEDIUM issues that need to be addressed before merge.

---

### HIGH

**H1 — Race condition in `waitlist_consume_invite`: missing re-check after `FOR UPDATE` lock**

The function calls `waitlist_validate_invite` (which reads `used_at IS NULL` without holding any row lock) before acquiring the `FOR UPDATE` lock. Two concurrent requests can both pass the validate check, serialize at the lock, and proceed. After acquiring the lock, the code only checks `v_entry.status = 'converted'` — it never re-verifies `v_inv.used_at IS NULL`. In the normal concurrent path the entry-status check provides protection, but there are edge cases (e.g. invite marked used manually, or a future code path that updates one row without the other) where a previously-used invite could be re-consumed.

Fix: add an explicit guard immediately after the `FOR UPDATE` SELECT:
```sql
IF v_inv.used_at IS NOT NULL OR v_inv.revoked_at IS NOT NULL THEN
    RETURN jsonb_build_object('error', 'invite_already_used');
END IF;
```
Inline comment at the `FOR UPDATE` SELECT line.

**H2 — Signup mode not enforced server-side**

`SignupModeGate` is a client-side React component. It hides the signup UI when the mode is `'waitlist'`, but nothing prevents a caller from hitting `supabase.auth.signUp()` directly (via the JS SDK, curl, or browser DevTools). The gate is cosmetic. Supabase's auth endpoint does not know about the waitlist mode, so new accounts can always be created regardless of the setting.

Fix: enforce mode in the `waitlist-capture` edge function and/or add a Supabase auth hook (e.g. `before_sign_in`) that checks the waitlist setting. At minimum, document the known gap in a `FIXME` comment.

**H3 — X-Forwarded-For spoofing bypasses rate limiting**

`clientIp()` takes the *first* IP from the `x-forwarded-for` header: `forwarded.split(',')[0]`. An attacker can prepend an arbitrary IP (`X-Forwarded-For: 1.2.3.4, real-ip`) and the rate limiter will track `1.2.3.4`, not their real IP. The correct field is the *rightmost* or the last entry appended by infrastructure you control.

Fix:
```ts
// Take last segment appended by infrastructure (rightmost = most trustworthy)
if (forwarded) return forwarded.split(',').at(-1)?.trim() ?? null;
```
Or read from a header that your edge runtime guarantees (e.g. Deno Deploy's `CF-Connecting-IP` or `x-real-ip`).

**H4 — CORS wildcard fallback when `Origin` header is absent**

When a request arrives without an `Origin` header (direct server-to-server calls, curl, scripts), `corsHeadersForWaitlist()` falls through to return `Access-Control-Allow-Origin: *`. This means any response to a credentialless request advertises `*`, which could allow cross-origin reads in browsers that send credentialless requests and could expose error payloads.

Fix: when `!origin`, return either no CORS headers at all (the safest option for non-browser requests) or a restrictive default, not `*`.

---

### MEDIUM

**M1 — Plaintext invite token in RPC responses**

The `token` field is returned in several RPC/admin list responses. Invite tokens should be treated as secrets. Returning them in query results risks exposure in Supabase Dashboard query logs, PostgREST access logs, and browser DevTools Network tab. Consider returning only the token once (on creation) and omitting it from subsequent responses.

**M2 — Rate limit incremented before duplicate email check**

In `waitlist_capture`, the IP rate-limit counter is incremented before the duplicate-email guard fires. A user submitting an already-registered email will burn a rate-limit slot silently and get a success response, making it impossible for a legitimate user on the same IP to know they're being throttled.

**M3 — No trigram index on `waitlist_entries.email` for ILIKE search**

The admin list RPC filters on `email ILIKE '%' || p_search || '%'`. Without a `pg_trgm` trigram index this is a full sequential scan on every admin search. Add `CREATE INDEX ... USING gin (email gin_trgm_ops)`.

**M4 — Unbounded `p_metadata` JSONB in `waitlist_capture`**

No size constraint on the metadata parameter. A caller can pass arbitrarily large JSON, inflating the `waitlist_entries` table. Add a `pg_column_size(p_metadata) < 8192` guard or a CHECK constraint on the column.

**M5 — No rate limit on admin invite resend**

`admin_resend_waitlist_invite` is only gated by admin role, not rate limited. A loop calling it repeatedly would spam an email address without bound.

**M6 — Undocumented dependency on `_admin_insert_audit`**

`waitlist_consume_invite` calls `public._admin_insert_audit(...)`. This function is defined in the admin migration, not in this one. If this migration runs before the admin migration (e.g. fresh environment, different migration order), it will fail at runtime. Add an explicit `DO $$ BEGIN ... EXCEPTION WHEN ...` guard or a migration comment noting the dependency.

---

### Test gaps

- No pgTAP test for token double-spend (two concurrent calls with the same token).
- No test that `supabase.auth.signUp()` is blocked when mode is `'waitlist'` (because it isn't — see H2).
- No test covering rate-limit window expiry (`waitlist_rate_limits.window_start` rollover).
- No test for the `clientIp` spoofing scenario.

---

### Minor

- `waitlist_validate_invite` is marked `STABLE` but it reads live table data with a plain SELECT. `STABLE` is correct for the volatility category (doesn't modify data, consistent within a statement), but it's worth a comment explaining that it intentionally does *not* acquire a lock — the lock is the caller's responsibility.
- `_waitlist_create_invite` has a sub-select with `AND used_at IS NULL` in the constraint, but the same filter is absent from the `FOR UPDATE` path in the consuming function (see H1).

Please address the four HIGH issues and M6 before merge. Happy to discuss any of these.

supabase/functions/waitlist-ops/index.ts
ZappoMan: Fixed in `0b2e65f`: `send_invite_email` sends via Resend when `WAITLIST_RESEND_API_KEY` is set; otherwise returns **501** `email_not_configured` so the admin drawer can tell operators to copy the link. `entryId` is required and must match the recipient email.

apps/web/src/pages/SignupInvitePage.tsx
ZappoMan: Fixed in `0b2e65f`: removed client-side `billing_ensure_subscription_plan`. Billing runs in `waitlist-ops` `consume` only when `already_converted` is not true.

supabase/migrations/20260519010000_billing_ensure_subscription_plan.sql
ZappoMan: Fixed in `0b2e65f` migration `20260520130000_waitlist_pr274_review_fixes.sql`: `ON CONFLICT DO UPDATE` now has `WHERE status NOT IN ('active', 'trialing', 'past_due')` so paid subscriptions are not overwritten.

supabase/migrations/20260519000000_waitlist_v1.sql
ZappoMan: Fixed in `0b2e65f`: `waitlist_capture` deletes rate-limit rows older than 24h at the start of each call.

supabase/migrations/20260519000000_waitlist_v1.sql
ZappoMan: Acknowledged — still using fixed hourly windows in v1. Pruning limits table growth; sliding-window rate limits are a good follow-up.

supabase/migrations/20260519000000_waitlist_v1.sql
ZappoMan: Fixed in `0b2e65f`: `admin_update_waitlist_settings` validates `signup_mode`, `identity_match_mode`, and `invite_ttl_days` and returns structured JSON errors.

supabase/migrations/20260519010000_billing_ensure_subscription_plan.sql
ZappoMan: Fixed in `0b2e65f`: `billing_ensure_subscription_plan` now takes `p_user_id`, is **service_role only** (revoked from `authenticated`), and is called from `waitlist-ops` `consume` after a successful invite consume.

supabase/migrations/20260519000000_waitlist_v1.sql
ZappoMan: Fixed in `0b2e65f`: `admin_get_waitlist_settings` and `admin_get_waitlist_entry` are now **VOLATILE**.

supabase/functions/waitlist-ops/index.ts
ZappoMan: Same as M1 — returns **501** `email_not_configured` without Resend; logs locally for dev. Documented in `waitlist-setup.md`.

supabase/migrations/20260519000000_waitlist_v1.sql
ZappoMan: Fixed in `0b2e65f`: audit payload now includes `invite_ttl_days`, `identity_match_mode`, `copy`, and `metadata_schema`.

apps/web/src/pages/AuthCallbackPage.tsx
ZappoMan: Fixed in `0b2e65f`: invite finalize errors redirect to `/login` after 3s; the 1200ms OAuth fallback path also checks `INVITE_TOKEN_STORAGE_KEY` before post-auth redirect.

supabase/migrations/20260519000000_waitlist_v1.sql
ZappoMan: Fixed in `0b2e65f` (same prune as Sarah's nit).

supabase/tests/waitlist.test.sql
ZappoMan: Added pgTAP coverage in `0b2e65f` for invalid `signup_mode` and authenticated revoke on billing RPC. Full consume-twice flow test is a good follow-up (needs invite fixture setup).

supabase/migrations/20260519000000_waitlist_v1.sql
ZappoMan: Fixed in `0b2e65f`: after `FOR UPDATE`, we return `invite_already_used` if `used_at` or `revoked_at` is set.

apps/web/src/pages/SignupPage.tsx
ZappoMan: Documented in `packages/waitlist/SECURITY.md`: `SignupModeGate` is UX-only; Supabase Auth sign-up is not blocked in v1. Auth hook enforcement is called out as follow-up.

supabase/functions/waitlist-capture/index.ts
ZappoMan: Fixed in `0b2e65f`: `clientIp` prefers `cf-connecting-ip` / `x-real-ip`, then the **rightmost** `x-forwarded-for` segment.

supabase/functions/_shared/waitlist-origins.ts
ZappoMan: Fixed in `0b2e65f`: when `Origin` is absent we no longer emit `Access-Control-Allow-Origin: *`.

ZappoMan:
## Review follow-up (`0b2e65f`)

Thanks for the detailed reviews. Inline replies are on each thread; summary for items that were review-body only:

**@mei-artificerinnovations**
- **HIGH-1 (arbitrary `to` / HTML):** `send_invite_email` now requires `entryId` and verifies it matches `email` via `admin_get_waitlist_entry` before send.
- **MEDIUM-2 (audit fields):** Settings audit logs all mutable fields including `identity_match_mode`.
- **MEDIUM-3 (OAuth timer + invite):** 1200ms fallback checks `INVITE_TOKEN_STORAGE_KEY` before `readAndClearPostAuthRedirect`.
- **LOW-1 (rate limit growth):** `waitlist_capture` prunes rows older than 24h.
- **LOW-2 (password length):** `SignupInvitePage` enforces min 8 characters before `signUp`.

**Copilot threads** were addressed in earlier commits (`4299410` / `b5da49c`); latest commit adds security hardening above.

Please re-review when you have a moment.

supabase/migrations/20260519000000_waitlist_v1.sql
ZappoMan: Tracked in #278 — sliding-window rate limits for `waitlist_capture`.

supabase/tests/waitlist.test.sql
ZappoMan: Tracked in #279 — pgTAP double-consume (`invite_already_used`) test with invite fixture setup.

apps/web/src/pages/SignupPage.tsx
ZappoMan: Tracked in #280 — server-side `signup_mode` enforcement via Supabase auth hook.

sarah-artificerinnovations (approved):
All blocking issues resolved and nits addressed. The approach improved from the original in several meaningful ways.

**M1 — `send_invite_email`:** Real Resend integration wired in. When `WAITLIST_RESEND_API_KEY` is absent the action now returns `{ error: 'email_not_configured' }` with 501 (no more silent success). Resend call uses `Authorization: Bearer` header, error logging on non-OK response, and `{ error: 'email_send_failed' }` with 502 on failure — correct error surfacing throughout.

**M2 — billing on already_converted:** Better than the original fix: billing is now done server-side inside the Edge Function's `consume` handler, not in the client. The Edge Function checks `!consumeResult.already_converted` before calling `billing_ensure_subscription_plan`, and uses the service-role client (`admin`) so the RPC grant change to service_role-only lands cleanly. `finalizeInviteSignup` no longer touches billing at all — the right architectural boundary.

**M3 — billing downgrade guard:** `WHERE billing_subscriptions.status NOT IN ('active', 'trialing', 'past_due')` added to the `ON CONFLICT DO UPDATE`. Function signature changed to include explicit `p_user_id uuid` (no longer relying on `auth.uid()`), grant narrowed to service_role only. Correct.

**Nits:** Rate limit rows pruned at top of `waitlist_capture` (24 h window). Metadata size capped at 8 KiB. `admin_update_waitlist_settings` validates `signup_mode` and `identity_match_mode` enums at the application level with clean JSON errors. Trigram index on `waitlist_entries.email` is a nice bonus for admin search at scale.

**SECURITY.md** updated to document the signup-mode-gating-is-client-side caveat (signUp() not blocked server-side) and the billing service_role scope — good operator transparency.

Approving.

rahul-artificerinnovations (approved):
## Re-review — PR #274 — APPROVED

All four HIGH issues from my original review are resolved. Approving.

---

### What changed since my review

**H1 — Race condition in `waitlist_consume_invite`** ✅ Fixed (`20260520130000_waitlist_pr274_review_fixes.sql`)

The guard I asked for is in place:
```sql
IF v_inv.used_at IS NOT NULL OR v_inv.revoked_at IS NOT NULL THEN
    RETURN jsonb_build_object('error', 'invite_already_used');
END IF;
```
This runs after the `FOR UPDATE` lock is acquired, so the check reflects committed state. Correct.

**H2 — Signup mode not enforced server-side** ✅ Documented (`packages/waitlist/SECURITY.md`)

Explicitly called out: *"Supabase Auth `signUp()` is not blocked by waitlist mode — a caller can still hit the auth API directly. v1 relies on operator workflow… For strict server enforcement, add a Supabase `before_user_created` hook (tracked follow-up)."* This is acceptable transparency for a V1 shipping with a known gap rather than a hidden one.

**H3 — X-Forwarded-For spoofing** ✅ Fixed (`supabase/functions/waitlist-capture/index.ts`)

`clientIp()` now prefers `cf-connecting-ip`, then `x-real-ip`, then falls back to the **rightmost** segment of `x-forwarded-for` (`.at(-1)`). Correct priority order.

**H4 — CORS wildcard fallback** ✅ Fixed (`supabase/functions/_shared/waitlist-origins.ts`)

When `!origin`, the function now returns only `Access-Control-Allow-Headers` with no `Access-Control-Allow-Origin` header at all. When origin is set but not allowlisted, same: no ACAO header, `Vary: Origin`. No wildcard leak.

**M3 — ILIKE without trigram index** ✅ Fixed — `CREATE EXTENSION IF NOT EXISTS pg_trgm` + GIN index on `email`.

**M4 — Unbounded metadata** ✅ Fixed — `pg_column_size(p_metadata) > 8192` guard returns early with a uniform success response.

---

### Remaining items (not blocking)

- **M1 — Raw invite token in admin RPC responses** — `admin_resend_waitlist_invite` and `admin_invite_waitlist_email` still return `invite_token: raw_token`. This is intentional for the admin-sends-link workflow but the token will appear in browser DevTools. Low risk since it's admin-only; worth considering whether to return a pre-formed invite URL instead of the raw token in a follow-up.

- **M5 — No rate limit on admin invite resend** — `admin_resend_waitlist_invite` is still unthrottled. Admin-only so the blast radius requires a compromised admin account. Fine for V1; add a per-entry resend cooldown in a follow-up.

- **Token double-spend pgTAP test** — still not present. The fix is correct but having a concurrent-call test would prevent regressions if the function is ever refactored. Low priority but worth tracking.

---

Good work addressing the security feedback quickly. The core risk surface is clean.

mei-artificerinnovations (approved):
All concerns from the previous two review rounds have been addressed. Approving.

## What got fixed

**HIGH (all resolved)**
- `billing_ensure_subscription_plan` — now takes explicit `p_user_id`, is restricted to `service_role`, and blocks downgrades from `active`/`trialing`/`past_due` ✅
- `waitlist_consume_invite` TOCTOU — `FOR UPDATE` on both the invite and entry rows, with a post-lock re-check of `used_at`/`revoked_at` ✅
- `AuthCallbackPage` unhandled rejection — `.catch()` now calls `setError()` and schedules a 3-second redirect to `/login` ✅
- `AuthCallbackPage` 1200ms timeout — now verifies `INVITE_TOKEN_STORAGE_KEY` is present before acting ✅

**MEDIUM (all resolved)**
- `waitlist_capture` metadata size — `pg_column_size(p_metadata) > 8192` check added at the DB layer ✅
- `admin_update_waitlist_settings` audit log — `identity_match_mode` is now included in the audit payload ✅
- `send_invite_email` phishing vector — email recipient is now validated against the DB entry via `admin_get_waitlist_entry`; recipient must match the stored entry email case-insensitively ✅
- `admin_update_waitlist_settings` enum validation — `signup_mode`, `identity_match_mode`, and `invite_ttl_days` are all validated before the UPDATE ✅
- `admin_get_waitlist_settings` VOLATILE — fixed ✅

## Remaining nits (non-blocking)

`SignupInvitePage.tsx` — the password length check hardcodes `8` (`if (password.length < 8)`) instead of using `MIN_PASSWORD_LENGTH` from `packages/shared`. Low risk since the constant is currently `8`, but worth aligning so a future config change covers this path too.

`supabase/functions/waitlist-ops/index.ts` — the `SendInviteEmailBody` type still has a duplicate `entryId?: string` field. Cosmetic only, but worth a quick cleanup.

Neither of these warrants blocking the merge. Ship it.

ZappoMan:
## Remaining reviewer nits triage

I reviewed the latest approval/re-review notes and all currently unresolved review threads. GitHub shows no unresolved review threads, so this is tracking the remaining non-blocking nits called out in the review bodies.

### Fixed in code

- **Mei: `SignupInvitePage` hardcoded password length (`8`)**
  - Updated `SignupInvitePage` to use shared `MIN_PASSWORD_LENGTH` from `@beakerstack/shared/config/auth` instead of hardcoding `8`.
  - The error copy now interpolates the shared value.

- **Mei: duplicate `entryId?: string` in `waitlist-ops` request body type**
  - Removed the duplicate field from the Edge Function `Body` type.

Verification: `npm run type-check:web` passes.

### Tracked in backlog

- **Sarah / Rahul: fixed-window waitlist capture rate limit can burst at hour boundaries**
  - Tracked in #278.

- **Diego / Rahul: pgTAP double-consume regression coverage**
  - Tracked in #279.

- **Rahul: server-side signup-mode enforcement / Supabase auth hook**
  - Tracked in #280.

- **Rahul: raw invite token returned to admin clients**
  - Tracked in #281.
  - V1 keeps this for copy-link fallback/admin send flow, but the follow-up will evaluate returning only server-formed invite URLs or moving link generation fully server-side.

- **Rahul: admin invite resend has no cooldown**
  - Tracked in #282.
  - Admin-only in v1, but worth adding per-entry cooldown/error surfacing to reduce blast radius from compromised admin sessions or automation loops.

### Already fixed in earlier commits

- Rate-limit row pruning: fixed in `0b2e65f` with 24h cleanup.
- `admin_update_waitlist_settings` validation/audit coverage: fixed in `0b2e65f`, with forward repair migration in `4fa11fa` for databases that already applied the older function definition.
- Metadata size cap and trigram email index: fixed in `0b2e65f`.
- Missing invite post-lock re-check, XFF spoofing, absent-origin CORS wildcard: fixed in `0b2e65f`.

diego-artificerinnovations (approved):
## Re-review — all blocking issues resolved ✅

Good turnaround on the fixes. Going through each concern from my original review:

**Security — billing privilege escalation (was blocking):** Fixed correctly. `billing_ensure_subscription_plan` is dropped (2-arg, granted to `authenticated`) and recreated as a 3-arg function (`p_user_id uuid`, revoked from `authenticated`, granted to `service_role` only). Called exclusively from `waitlist-ops consume` using the service-role admin client. The `ON CONFLICT ... WHERE status NOT IN (...)` guard preventing paid-plan downgrades is also good.

**Security nit — STABLE + VOLATILE:** Both `admin_get_waitlist_settings` and `admin_get_waitlist_entry` are now `VOLATILE`. Fixed.

**Nit — `send_invite_email` no-op:** Now uses Resend API when `WAITLIST_RESEND_API_KEY` is set; returns `501 email_not_configured` otherwise instead of a fake `200`. The `entryId` guard verifying the recipient email matches the entry is a nice addition.

**Nit — rate-limit table growth:** `waitlist_capture` now prunes rows older than 24h on each call. Acceptable for v1.

**Nit — audit log missing `identity_match_mode`:** Audit payload in `admin_update_waitlist_settings` now includes all mutable fields including `identity_match_mode`, `invite_ttl_days`, and `copy`. Fixed.

**Nit — finalizeInviteSignup error leaves user stuck:** `AuthCallbackPage` now shows the error banner and redirects to `/login` after 3s in all error paths. Fixed.

**Nit — pgTAP coverage:** Additional tests added for invalid `signup_mode` validation and the authenticated-role revoke on the billing RPC. Full `invite_already_used` (consume-twice) test tracked in #279 — that's fine.

One thing to clean up when convenient: `entryId` appears twice in the `Body` type in `supabase/functions/waitlist-ops/index.ts`. TypeScript silently takes the last definition so there's no runtime impact, but worth deleting the duplicate before it causes confusion.

Approving — the two blocking security issues are resolved and everything else has been addressed or tracked.

Issue #291, opened 2026-05-19
Automate Resend domain setup with Route 53 DNS (follow-up to #284)


ZappoMan:
## Summary

Follow-up to #284 / PR #287. v1 shipped branded Supabase Auth templates, optional SMTP wiring via `env(SMTP_*)`, and **manual** Resend domain verification (adopter copies SPF/DKIM into DNS in the dashboard).

This issue adds an **optional, advanced setup phase** that automates **both sides end-to-end**:

1. **Resend** — create domain, configure auth-safe tracking defaults, fetch required DNS records, trigger verification, poll until `verified`
2. **Route 53** — upsert SPF/DKIM (and optional DMARC) records into the adopter’s existing hosted zone
3. **BeakerStack config** — write `SMTP_*` to `.env.local`, enable `[auth.email.smtp]` in `supabase/config.toml`

**Goal:** An adopter with a Resend API key and Route 53 managing their apex runs one setup flow and ends with a **verified** Resend sending domain and working Supabase Auth SMTP config—no dashboard copy/paste of TXT/CNAME values.

**Parent:** #284 (explicitly deferred automated DNS; Resend was always intended as the recommended SMTP provider)

## Why this is feasible now

Resend exposes a full **Domains API** and **official CLI** (`resend domains create | get | verify | update`). Almost nothing on the Resend side requires the dashboard after the adopter supplies `RESEND_API_KEY`.

| Manual step today | Automate via |
|-------------------|--------------|
| Add domain (`auth.example.com` or apex) | `POST /domains` or `resend domains create --name …` |
| Copy SPF/DKIM records | `records[]` on create/get response → Route 53 upsert |
| Disable link/open tracking (Supabase URL safety) | `click_tracking: false`, `open_tracking: false` on create; **do not** set `tracking_subdomain` |
| Trigger + wait for verification | `POST /domains/{id}/verify` + poll `GET /domains/{id}` until `status === "verified"` |
| Paste SMTP creds + enable config | BeakerStack script writes `.env.local` + uncomments `config.toml` |

[...]

PR #298, opened 2026-05-19
feat(#291): automate Resend domain setup with Route 53 DNS

sarah-artificerinnovations:
## Summary

Implements issue #291: automate Resend domain setup with Route 53 DNS, resolving all blocking feedback from Rahul, Mei, Diego, and Brad (v3.1).

**New files:**
- `scripts/lib/setup-resend-api.mjs` — Resend domain lifecycle: `createOrGetResendDomain()` + `verifyResendDomain()` (split so DNS is written before verification is triggered)
- `scripts/lib/setup-route53-email-records.mjs` — Route 53 UPSERT with `formatTxtValue()` (TXT quoting + 255-char DKIM chunking), DMARC, SPF pre-flight warning, trailing-dot normalization, apex zone discovery
- `scripts/setup-email-dns.mjs` — standalone orchestrator + `phaseEmailDns` export for setup-full
- `scripts/__tests__/setup-route53-email-records.test.mjs` — unit tests for `formatTxtValue` and `uncommentSmtpSection`

**Modified files:**
- `scripts/setup-full.mjs` — `email-dns` phase wired after `aws`, `--skip-email-dns` flag
- `package.json` — `setup:email` script added, test file registered in `test:unit:scripts`

**Orchestration order (correct per issue spec):**
1. Create/reuse Resend domain → `records[]`
2. Show planned R53 diff + confirm (skipped with `--yes`)
3. UPSERT Route 53 (SPF/DKIM + optional DMARC at `_dmarc.<sendingDomain>`)
4. Trigger Resend verification + exponential backoff poll (5s→15s→30s→60s, graceful Ctrl-C)  — **skipped in `--yes` mode**, prints pending message instead
5. Merge SMTP_* into `.env.local` (preserves all other keys via `parseDotEnv` + spread)
6. Uncomment `[auth.email.smtp]` in `supabase/config.toml` (section-bounded, idempotent)
7. Print completion with hosted-Supabase Dashboard callout

**Key design decisions:**
- SMTP port 587 throughout
- `RESEND_API_KEY` from env only (no CLI flag — keeps out of shell history)
- `SMTP_ADMIN_EMAIL` defaults to `noreply@<sendingDomain>`
- DMARC RUA defaults to `SMTP_ADMIN_EMAIL`; DMARC record at `_dmarc.<sendingDomain>`
- Zone discovery strips `auth.<apex>` → `<apex>`; surfaces clear error on zero or multiple zones
- Dry-run support end-to-end

## Test plan

- [ ] `npm run test:unit:scripts` passes (includes new `formatTxtValue` and `uncommentSmtpSection` tests)
- [ ] `node scripts/setup-email-dns.mjs --dry-run --domain=auth.example.com` prints planned actions, no mutations
- [ ] `node scripts/setup-full.mjs --dry-run` includes email-dns phase output, exits cleanly
- [ ] `node scripts/setup-full.mjs --skip-email-dns --dry-run` skips email-dns phase
- [ ] Re-running `setup-email-dns` on an existing Resend domain reuses it (409 path) and is idempotent

Closes #291


scripts/lib/setup-route53-email-records.mjs
```diff
@@ -0,0 +1,203 @@
...
+  }
+}
+
+// --- internals ---
+
+async function resolveHostedZone(hostedZoneId, apexDomain, awsProfileArgs) {
+  if (hostedZoneId) return hostedZoneId;
+
+  const res = await discoverRoute53PublicZonesForApex(apexDomain, awsProfileArgs);
+  if (!res.ok) {
```
mei-artificerinnovations, line 142: Wrap this in try-catch. An AWS auth error or CLI-not-found throws a rejected promise here — the `!res.ok` guard on line 108 never runs. Something like:
```js
let res;
try {
  res = await discoverRoute53PublicZonesForApex(apexDomain, awsProfileArgs);
} catch (err) {
  throw new Error(`Route 53 zone discovery failed: ${err.message}
Pass --hosted-zone-id to skip auto-discovery.`);
}
```

scripts/setup-email-dns.mjs
```diff
@@ -0,0 +1,471 @@
...
+
+async function mergeEnvLocal(smtpVars, dryRun) {
+  if (dryRun) {
+    logInfo('[dry-run] would merge SMTP_* keys into .env.local (preserving all other keys)');
+    return;
+  }
+  const existing = await readEnvFileIfExists(LOCAL_ENV_PATH);
+  const merged = { ...existing, ...smtpVars };
+  await fs.writeFile(LOCAL_ENV_PATH, stringifyDotEnv(merged), 'utf8');
+  logInfo('.env.local: SMTP_* keys merged (all other keys preserved)');
```
mei-artificerinnovations: Both writeFile calls need error handling. A disk-full or permission-denied here leaves the user in a half-applied state (env updated, toml not, or vice versa). Wrap each in try-catch and exit with a message telling them which file failed and to re-run once the issue is fixed.

mei-artificerinnovations (changes requested):
Implementation matches the v3 plan on all the critical points — backoff sequence (5→15→30→60, no hard timeout), apex zone discovery, section-bounded `config.toml` uncomment, DMARC placement, port 587, `--yes` behavior, dry-run coverage, and 409 idempotency. Test coverage for `formatTxtValue` and `uncommentSmtpSection` is thorough.

Two small error-handling gaps before this ships:

**1. `resolveHostedZone()` — unhandled rejection if zone discovery throws**

`discoverRoute53PublicZonesForApex()` is awaited without a try-catch. An AWS auth error, timeout, or CLI-not-found returns a rejected promise that bypasses the `!res.ok` guard entirely and crashes the process with an unhelpful stack trace. Wrap in try-catch and rethrow with a user-readable message (point them to `--hosted-zone-id` as the fallback).

**2. `mergeEnvLocal()` — unhandled file-write rejection**

Both `fs.writeFile()` calls (`.env.local` and `config.toml`) are un-guarded. Disk-full or permission-denied crashes the process mid-setup, which can leave the user in a half-applied state (env updated, toml not, or vice versa). Wrap each in try-catch, print which file failed and why, and exit cleanly.

These are the only two blocking items. Everything else is nit-level — one minor note inline.

scripts/lib/setup-route53-email-records.mjs
```diff
@@ -0,0 +1,203 @@
...
+  try {
+    const existing = JSON.parse(result.stdout);
+    if (Array.isArray(existing) && existing.length > 0) {
+      const hasSpf = existing.some(rrs =>
+        (rrs.ResourceRecords ?? []).some(r => r.Value?.includes('v=spf1'))
+      );
+      if (hasSpf) {
+        console.warn(
+          `\n⚠  Existing SPF TXT record found for ${sendingDomain}.\n` +
+            '   The UPSERT will replace it with Resend's SPF record.\n' +
```
ZappoMan: **Blocker — parse error:** The apostrophe in `Resend's` terminates the single-quoted string, so Node throws `SyntaxError: missing ) after argument list` and the whole file fails to import.

```suggestion
            "   The UPSERT will replace it with Resend's SPF record.\n" +
```

(or escape: `Resend\'s`)

I confirmed locally: `node --test scripts/__tests__/setup-route53-email-records.test.mjs` fails until this is fixed.

scripts/lib/setup-resend-api.mjs
```diff
@@ -0,0 +1,169 @@
...
+/**
+ * Create or reuse a Resend sending domain, disable tracking, and return DNS records.
+ * Does NOT trigger verification — call verifyResendDomain() after DNS records are in Route 53.
+ *
+ * @param {string} sendingDomain e.g. "auth.myapp.com"
+ * @param {{ dryRun?: boolean }} opts
+ * @returns {Promise<{ domainId: string; records: ResendRecord[] }>}
+ */
+export async function createOrGetResendDomain(sendingDomain, { dryRun = false } = {}) {
+  const apiKey = requireApiKey();
```
ZappoMan: `requireApiKey()` runs before the `dryRun` branch, so `--dry-run` still requires `RESEND_API_KEY` (and standalone will prompt for it). The PR description says dry-run works end-to-end without mutations — consider moving `requireApiKey()` below the dry-run early return (same pattern in `verifyResendDomain` at line 45).

scripts/lib/setup-resend-api.mjs
```diff
@@ -0,0 +1,169 @@
...
+
+async function _createOrGetDomain(name, apiKey) {
+  try {
+    return await resendFetch('/domains', apiKey, {
+      method: 'POST',
+      body: JSON.stringify({ name, click_tracking: false, open_tracking: false }),
+    });
+  } catch (err) {
+    if (err.status === 409) {
+      const list = await resendFetch('/domains', apiKey);
```
ZappoMan: On 409, this only scans the first page of `GET /domains` (default limit 20). Resend supports cursor pagination (`has_more`, `after`). If the domain isn't on page 1, you'll throw even though it exists.

Suggest: loop pages until found, or add a name filter if the API supports it.

scripts/lib/setup-resend-api.mjs
```diff
@@ -0,0 +1,169 @@
...
+
+  const sigintHandler = () => { siginted = true; };
+  process.once('SIGINT', sigintHandler);
+
+  process.stdout.write(
+    '  Waiting for Resend domain verification (Ctrl-C to stop waiting and continue setup)…\n'
+  );
+
+  try {
+    while (true) {
```
ZappoMan: Minor: the first status check happens only after the initial 5s sleep. For an already-verified domain (re-run), consider checking status once before entering the backoff loop so a quick re-run can exit immediately.

scripts/setup-email-dns.mjs
```diff
@@ -0,0 +1,471 @@
...
+  dmarc,
+  dmarcEmail,
+  awsProfileArgs,
+  dryRun,
+  nonInteractive,
+  yes,
+}) {
+  // 1. Create/reuse Resend domain — get DNS records (no verify yet)
+  logInfo(`Creating/reusing Resend domain: ${sendingDomain}`);
+  const { domainId, records } = await createOrGetResendDomain(sendingDomain, { dryRun });
```
ZappoMan: Resend domain create/reuse happens **before** the Route 53 confirmation prompt (step 2). If the operator answers `n` at line 219, they still have a Resend domain (and possibly disabled tracking) but no DNS — and `phaseEmailDns` still merges SMTP keys into `acc` (line 99).

Consider moving `createOrGetResendDomain` to after confirmation, or documenting that abort leaves a Resend-side artifact.

scripts/setup-email-dns.mjs
```diff
@@ -0,0 +1,471 @@
...
+
+  if (!yes && !dryRun && rl) {
+    console.log('\nPlanned DNS changes:');
+    for (const c of planResult.records) {
+      const rrs = c.ResourceRecordSet;
+      console.log(`  UPSERT  ${rrs.Type.padEnd(5)}  ${rrs.Name}`);
+    }
+    const answer = (await rl.question('\nProceed with Route 53 UPSERT? [Y/n]: ')).trim().toLowerCase();
+    if (answer === 'n' || answer === 'no') {
+      logInfo('Aborted by user.');
```
ZappoMan: On user abort here, `runEmailDns` returns without writing `.env.local` / `config.toml`, but `phaseEmailDns` always runs `Object.assign(acc, buildSmtpVars(...))` afterward. `phaseWrite` will still persist SMTP_* as if setup completed.

Return a status from `runEmailDns` and only update `acc` when the phase actually finished.

scripts/setup-email-dns.mjs
```diff
@@ -0,0 +1,471 @@
...
+  // 4. Trigger verification + poll (after DNS is written)
+  logInfo('Triggering Resend domain verification…');
+  const { verified } = await verifyResendDomain(domainId, {
+    dryRun,
+    nonInteractive: nonInteractive || yes,
+  });
+
+  // 5. Write .env.local (merge only SMTP keys) and uncomment config.toml
+  const smtpVars = buildSmtpVars(sendingDomain, adminEmail, senderName);
+  await mergeEnvLocal(smtpVars, dryRun);
```
ZappoMan: This writes `.env.local` during the email-dns phase, and `phaseWrite` later rewrites `.env.local` from `acc` with a different header (`# Generated by npm run setup:full`). Functionally OK if keys match, but:

- Comments in `.env.local` are stripped (parse → stringify), same as setup-full but worth noting for operators who annotate secrets.
- Consider only updating `acc` here and letting `phaseWrite` own the file write to avoid two serializers / headers.

scripts/setup-email-dns.mjs
```diff
@@ -0,0 +1,471 @@
...
+export function uncommentSmtpSection(toml) {
+  // Match: `# [auth.email.smtp]` line + all immediately following `# ` prefixed lines
+  // Stops at blank line or next `[` header (both are not matched by `# [^\n]*`)
+  const pattern = /# \[auth\.email\.smtp\]\n(# [^\n]*\n)*/;
+  if (!pattern.test(toml)) return toml;
+
+  const smtpBlock =
+    '[auth.email.smtp]\n' +
+    'enabled = true\n' +
+    'host = "smtp.resend.com"\n' +
```
ZappoMan: This hardcodes `host = "smtp.resend.com"` and `user = "resend"`, but the repo template and `docs/EMAIL_TEMPLATES.md` use `env(SMTP_HOST)` / `env(SMTP_USER)` (see `supabase/config.toml` lines 219–224). The script already writes `SMTP_*` to `.env.local` — aligning the uncommented block with the docs avoids locking operators to Resend in config while env vars say otherwise.

scripts/setup-email-dns.mjs
```diff
@@ -0,0 +1,471 @@
...
+    'port = 587\n' +
+    'user = "resend"\n' +
+    'pass = "env(SMTP_PASS)"\n' +
+    'admin_email = "env(SMTP_ADMIN_EMAIL)"\n' +
+    'sender_name = "env(SMTP_SENDER_NAME)"\n';
+
+  return toml.replace(pattern, smtpBlock);
+}
+
+function apexFromDomain(domain) {
```
ZappoMan, line 682: `apexFromDomain` always takes the last two labels (`auth.example.co.uk` → `co.uk`). Zone discovery already warns about two-part TLDs, but this helper will pick the wrong apex before discovery runs.

Consider reusing `normalizeApexDomain` from `setup-aws-discover.mjs` or accepting apex from the aws phase (`acc.PR_PREVIEW_DOMAIN`) when available.

scripts/setup-email-dns.mjs
```diff
@@ -0,0 +1,471 @@
...
+      flags.plainSecretPrompts ?? false
+    );
+    if (!key) {
+      logWarn('No Resend API key provided; skipping email DNS setup.');
+      return;
+    }
+    process.env.RESEND_API_KEY = key;
+  }
+
+  const sendingDomain = (await rl.question('Sending domain (e.g. auth.myapp.com): ')).trim();
```
ZappoMan: Nit for setup-full UX: after `phaseAws`, `acc.PR_PREVIEW_DOMAIN` is the apex and `acc.PR_PREVIEW_HOSTED_ZONE_ID` is set — you already reuse the zone ID (line 71). Defaulting the sending domain prompt to `auth.${acc.PR_PREVIEW_DOMAIN}` (when present) would reduce typos and apex mismatch.

scripts/setup-email-dns.mjs
```diff
@@ -0,0 +1,471 @@
...
+    sendingDomain,
+    apex,
+    hostedZoneId,
+    adminEmail,
+    senderName,
+    dmarc: addDmarc,
+    dmarcEmail: addDmarc ? adminEmail : undefined,
+    awsProfileArgs,
+    dryRun,
+    nonInteractive: false,
```
ZappoMan, line 129: `nonInteractive` is always `false` when called from setup-full, so the verification poll always runs in the wizard (unlike standalone `--yes`). If that's intentional, a one-line log at phase start would help; if not, consider skipping poll when stdin isn't a TTY or when `flags.dryRun`.

scripts/lib/setup-route53-email-records.mjs
```diff
@@ -0,0 +1,203 @@
...
+    );
+  }
+  if (res.zones.length === 0) {
+    throw new Error(
+      `No public Route 53 hosted zone found for "${apexDomain}".\n` +
+        'Ensure the zone exists in your AWS account and your credentials have route53:ListHostedZonesByName permission.\n' +
+        'If your domain uses a two-part TLD (e.g. .co.uk), pass --hosted-zone-id explicitly.'
+    );
+  }
+  if (res.zones.length > 1) {
```
ZappoMan, line 155: Unknown Resend record types are silently skipped. That's reasonable for tracking CNAMEs, but a single `logWarn` for skipped types would make debugging easier when Resend adds a required record type.

scripts/__tests__/setup-route53-email-records.test.mjs
```diff
@@ -0,0 +1,98 @@
+import { describe, it } from 'node:test';
+import assert from 'node:assert/strict';
+
+import { formatTxtValue } from '../lib/setup-route53-email-records.mjs';
+import { uncommentSmtpSection } from '../setup-email-dns.mjs';
```
ZappoMan: Importing `uncommentSmtpSection` from `setup-email-dns.mjs` pulls in the full orchestrator module graph (Resend + Route 53 libs). Consider moving `uncommentSmtpSection` to a small `scripts/lib/setup-email-config.mjs` (or similar) so unit tests stay lightweight — optional cleanup.

ZappoMan (changes requested):
## Overall

Solid structure: clear split between Resend / Route 53 / orchestration, correct verify-after-DNS ordering, and good reuse of existing setup helpers (`parseDotEnv`, masked prompts, `discoverRoute53PublicZonesForApex`). The `formatTxtValue` tests are helpful.

**Blocking:** `setup-route53-email-records.mjs` has a syntax error (`Resend's` inside a single-quoted string) — the module does not load and `npm run test:unit:scripts` fails.

**Before merge:** fix the syntax error, align `config.toml` output with `docs/EMAIL_TEMPLATES.md` (`env(SMTP_*)`), and address dry-run / abort side-effects called out inline.

Happy to re-review after an update.

mei-artificerinnovations (approved):
Both fixes are exactly right.

- `resolveHostedZone()`: try-catch correctly wraps only the discovery call, preserves the existing `!res.ok` guard for the non-throw case, and the error message points to both root causes (CLI/credentials) and the escape hatch (`--hosted-zone-id`). ✓
- `mergeEnvLocal()` + `applyConfigToml()`: both `writeFile` calls independently guarded, each error message names the specific file that failed and tells the user what state they are in. ✓

Approved.

scripts/lib/setup-route53-email-records.mjs
```diff
@@ -0,0 +1,203 @@
...
+    const type = (rec.type ?? '').toUpperCase();
+    const name = normalizeRecordName(rec.name ?? '');
+    const value = rec.value ?? '';
+    const ttl = rec.ttl ?? 300;
+
+    if (type === 'TXT') {
+      changes.push(upsertRecord(name, 'TXT', ttl, [{ Value: formatTxtValue(value) }]));
+    } else if (type === 'CNAME') {
+      changes.push(upsertRecord(name, 'CNAME', ttl, [{ Value: value.replace(/\.$/, '') }]));
+    } else if (type === 'MX') {
```
rahul-artificerinnovations, line 260: **Double zone discovery (nit):** `resolveHostedZone` is called twice — once for the dry-run plan in `runEmailDns` and again for the real UPSERT. When `--hosted-zone-id` is not provided, that's two `ListHostedZonesByName` calls. Easiest fix: have `upsertEmailDnsRecords` return `resolvedZoneId` alongside `changeId`, then pass it from the plan call into the real call. Or resolve the zone once upfront in `runEmailDns` and pass it to both.

scripts/lib/setup-route53-email-records.mjs
```diff
@@ -0,0 +1,203 @@
...
+      'route53',
+      'list-resource-record-sets',
+      '--hosted-zone-id',
+      zoneId,
+      '--query',
+      `ResourceRecordSets[?Name=='${sendingDomain}.' && Type=='TXT']`,
+      ...awsProfileArgs,
+    ],
+    { encoding: 'utf8' }
+  );
```
rahul-artificerinnovations, line 317: **JMESPath interpolation (low-risk, worth a note):** `sendingDomain` is interpolated directly into the JMESPath string passed to `--query`. A domain containing a single quote (e.g. from a typo or malformed `--domain=` input) would produce invalid JMESPath. The consequence is benign — the `try/catch` swallows the parse error and the SPF warning is silently skipped — but it's worth adding a quick validation that `sendingDomain` is a valid hostname before this point, or escaping the JMESPath value. Since `spawnSync` doesn't use a shell, there's no injection into the OS, just a broken query.

scripts/setup-email-dns.mjs
```diff
@@ -0,0 +1,471 @@
...
+/**
+ * Uncomment the `# [auth.email.smtp]` block and replace placeholder values with Resend config.
+ *
+ * Matches the commented section header `# [auth.email.smtp]` plus all consecutive `# ` prefixed
+ * lines that follow (bounded by blank line or next `[` header — Mei's section-bound constraint).
+ * Replaces the whole matched block with a ready-to-use Resend SMTP config.
+ * Idempotent: if the block is already uncommented (no `# [auth.email.smtp]` present), returns toml unchanged.
+ *
+ * @param {string} toml
+ * @returns {string}
```
rahul-artificerinnovations, line 630: **Orphaned prose comment (nit):** The regex `# \[auth\.email\.smtp\]\n(# [^\n]*\n)*` matches from the `# [auth.email.smtp]` line onward, but leaves any preceding prose comment (e.g. `# Use a production-ready SMTP server`) in place. After the replace the file has:

```
# Use a production-ready SMTP server
[auth.email.smtp]
enabled = true
...
```

Minor cosmetic issue. You could widen the pattern to optionally consume a preceding `# ` comment line, or just accept it — the comment is still loosely accurate.

scripts/setup-email-dns.mjs
```diff
@@ -0,0 +1,471 @@
...
+    SMTP_SENDER_NAME: senderName,
+  };
+}
+
+async function readEnvFileIfExists(filePath) {
+  try {
+    return parseDotEnv(await fs.readFile(filePath, 'utf8'));
+  } catch {
+    return {};
+  }
```
rahul-artificerinnovations, line 536: **`.env.local` comment preservation (nit):** `parseDotEnv` strips comments (lines starting with `#`), so any user-written comments in an existing `.env.local` are silently dropped on merge, replaced by the single `# Generated by setup — do not commit` header. Probably acceptable for a setup script, but worth a note in the completion output or `--help` so users aren't surprised if they had inline documentation in their env file.

rahul-artificerinnovations (approved):
Solid implementation — all v3 plan items landed correctly. Orchestration order is right, `formatTxtValue` and `uncommentSmtpSection` match the spec, exponential backoff + graceful Ctrl-C is done well, zone-discovery error messages are clear, and the test suite covers the important cases. Four small things below (three nits, one low-risk security note), none blocking.

scripts/lib/setup-route53-email-records.mjs
```diff
@@ -0,0 +1,212 @@
...
+      console.log(`  ${rrs.Type.padEnd(5)} ${rrs.Name}`);
+    }
+    return { changeId: 'dry-run', records: changes };
+  }
+
+  await checkExistingSpf(zoneId, sendingDomain, awsProfileArgs);
+
+  const changeBatch = { Changes: changes };
+  const tmpFile = path.join(tmpdir(), `beakerstack-r53-email-${Date.now()}.json`);
+  writeFileSync(tmpFile, JSON.stringify(changeBatch, null, 2));
```
diego-artificerinnovations: Nit: the temp file is written with the process umask (typically `0644` on dev machines), which means other local users can read the Route 53 change batch.

Not a secret — just DNS record values — but worth tightening:
```js
writeFileSync(tmpFile, JSON.stringify(changeBatch, null, 2), { mode: 0o600 });
```

scripts/lib/setup-route53-email-records.mjs
```diff
@@ -0,0 +1,212 @@
...
+    const ttl = rec.ttl ?? 300;
+
+    if (type === 'TXT') {
+      changes.push(upsertRecord(name, 'TXT', ttl, [{ Value: formatTxtValue(value) }]));
+    } else if (type === 'CNAME') {
+      changes.push(upsertRecord(name, 'CNAME', ttl, [{ Value: value.replace(/\.$/, '') }]));
+    } else if (type === 'MX') {
+      changes.push(upsertRecord(name, 'MX', ttl, [{ Value: value }]));
+    }
+    // Skip unknown types (e.g. Resend tracking subdomains — not needed)
```
diego-artificerinnovations: Silent skip is intentional but if Resend ever adds a new record type (e.g. a `TXT` variant or an `A` record), this will silently drop it and the domain may not verify.

Consider a `console.warn`:
```js
// Skip unknown types (e.g. Resend tracking subdomains — not needed)
else { console.warn(`⚠ Skipping unrecognised Resend record type "${type}" for ${name} — check if verification requires it.`); }
```

scripts/lib/setup-resend-api.mjs
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  return _pollVerification(domainId, apiKey);
+}
+
+// --- internals ---
+
+function requireApiKey() {
+  const apiKey = process.env.RESEND_API_KEY;
+  if (!apiKey) throw new Error('RESEND_API_KEY is not set in the environment.');
```
diego-artificerinnovations, line 76: Nit: Resend API keys always start with `re_`. A cheap format check here would catch a pasted secret or test value early with a clear message rather than a confusing 401:
```js
if (!apiKey) throw new Error('RESEND_API_KEY is not set in the environment.');
if (!apiKey.startsWith('re_')) throw new Error('RESEND_API_KEY looks wrong — Resend keys start with "re_". Check the value in your environment.');
```

scripts/__tests__/setup-route53-email-records.test.mjs
```diff
@@ -0,0 +1,98 @@
+import { describe, it } from 'node:test';
```
diego-artificerinnovations, line 1: `formatTxtValue` and `uncommentSmtpSection` are well covered. One gap: `upsertEmailDnsRecords` itself has no tests — DMARC record generation, type mapping (CNAME trailing-dot strip, MX pass-through), and the multi-zone error path are all untested.

Not a blocker for this PR (mocking `spawnSync` + `aws` is noisy), but worth a follow-up issue so we don't silently regress the DMARC or SPF paths.

diego-artificerinnovations (approved):
All four blocking items from the issue #291 plan review are implemented correctly:
- `formatTxtValue()` with 255-char chunking and double-quote stripping ✓
- DMARC RUA defaults to `SMTP_ADMIN_EMAIL` ✓
- SPF pre-flight conflict warning ✓
- Trailing-dot normalization on CNAME/MX ✓

A few small things worth noting inline. Nothing blocking.

ZappoMan:
## Review status — @ZappoMan (issue #291 + PR threads)

Audited current HEAD (`7d8858c` — lint + unit test fixes). **Most v3.1 / team blocking items are resolved.** A few PR review threads are still open (called out below and in inline replies).

### Issue #291 — Brad v3.1 feedback

| Item | Status |
|------|--------|
| Verify/poll **after** Route 53 UPSERT | ✅ `createOrGetResendDomain()` + `verifyResendDomain()`; orchestrator order in `runEmailDns` |
| `uncommentSmtpSection` for **commented** `# [auth.email.smtp]` block | ✅ Pattern matches commented header + `# ` lines; section-bounded |
| Skip infinite poll in `--yes` / non-interactive | ✅ `verifyResendDomain(..., { nonInteractive })` skips poll |
| `SMTP_ADMIN_EMAIL` before DMARC | ✅ Resolved before `runEmailDns` / UPSERT |
| `npm run setup:email` | ✅ `package.json` |
| Pre-apply R53 diff + confirm | ✅ Plan + prompt unless `--yes` / `--dry-run` |
| AWS profile + zone reuse | ✅ `--aws-profile`; `acc.PR_PREVIEW_HOSTED_ZONE_ID` in `phaseEmailDns` |
| `.env.local` merge (SMTP keys only) | ✅ `parseDotEnv` + spread; try/catch on write |
| Docs (`EMAIL_TEMPLATES.md`, prep checklist, IAM) | ❌ Not in this PR — still outstanding vs issue AC |
| `--check-only` | ⏸ Deferred per plan |

### PR review — blocking / requested changes

| Item | Status |
|------|--------|
| Mei: `resolveHostedZone` try/catch | ✅ |
| Mei: `mergeEnvLocal` / `applyConfigToml` write errors | ✅ |
| Brad: `Resend's` syntax error | ✅ (double-quoted string) |
| Brad: `config.toml` uses `env(SMTP_*)` for host/user | ⚠️ **Partial** — `pass`, `admin_email`, `sender_name` use `env(...)`; `host` / `user` still hardcoded to Resend |
| Brad: dry-run without API key | ❌ `requireApiKey()` still runs before `dryRun` in `createOrGetResendDomain` |
| Brad: abort at confirm → `phaseEmailDns` still updates `acc` | ❌ `Object.assign(acc, …)` runs even when user answers `n` |
| Brad: Resend create before confirm on abort | ❌ Still creates domain before R53 confirm prompt |

### Nits / follow-ups (non-blocking, not yet done)

- 409 domain list pagination; poll once before backoff; `re_` key format check; warn on unknown record types; temp file `0o600`; double zone discovery; `env(SMTP_HOST)` / `env(SMTP_USER)` in TOML; default `auth.<apex>` prompt in setup-full; JMESPath quote escape; orphaned prose comment above SMTP block; `.env.local` comment loss note in help; extract `uncommentSmtpSection` for lighter tests; `upsertEmailDnsRecords` integration tests (follow-up issue).

**Recommendation:** Merge after fixing the three open Brad items (TOML `env()` alignment, dry-run API key gate, abort/`acc` side-effects) — or explicitly accept the abort/dry-run behavior and document it. Docs can be a fast follow-up PR to close #291 fully.

scripts/lib/setup-route53-email-records.mjs
ZappoMan: **Resolved** in `resolveHostedZone()` — `discoverRoute53PublicZonesForApex` is wrapped in try/catch with a message pointing to `--hosted-zone-id` (lines 122–129).

scripts/setup-email-dns.mjs
ZappoMan: **Resolved** — `mergeEnvLocal()` and `applyConfigToml()` each wrap `fs.writeFile` in try/catch with file-specific error messages (lines 297–305, 327–335).

scripts/lib/setup-route53-email-records.mjs
ZappoMan: **Resolved** — SPF warning string now uses double quotes so the module loads; `npm run test:unit:scripts` passes on current HEAD (`7d8858c`).

scripts/lib/setup-resend-api.mjs
ZappoMan: **Not yet resolved** — `createOrGetResendDomain()` still calls `requireApiKey()` before the `dryRun` early return, so `--dry-run` still needs `RESEND_API_KEY`. Suggest moving the key check below the dry-run branch.

scripts/lib/setup-resend-api.mjs
ZappoMan: **Not yet resolved** — 409 reuse still only searches the first page of `GET /domains`. Worth paginating with `has_more` / `after` before merge.

scripts/lib/setup-resend-api.mjs
ZappoMan: **Not yet resolved** (nit) — poll still sleeps 5s before the first status check; quick re-run on an already-verified domain could check once up front.

scripts/setup-email-dns.mjs
ZappoMan: **Not yet resolved** — `createOrGetResendDomain` still runs before the Route 53 confirm prompt; abort at `n` leaves a Resend-side domain. Consider moving create after confirm or documenting the artifact.

scripts/setup-email-dns.mjs
ZappoMan: **Not yet resolved** — `phaseEmailDns` still always `Object.assign(acc, buildSmtpVars(...))` after `runEmailDns`, even when the user aborts at confirm. `runEmailDns` should return a status and `phaseEmailDns` should only update `acc` on success.

scripts/setup-email-dns.mjs
ZappoMan: **Partially addressed** — standalone still writes `.env.local` in-phase; setup-full also merges via `phaseWrite`. Keys should match; comment headers differ. Acceptable for v1 but worth consolidating later.

scripts/setup-email-dns.mjs
ZappoMan: **Partially resolved** — `pass`, `admin_email`, and `sender_name` now use `env(...)`. `host` and `user` are still hardcoded to `smtp.resend.com` / `resend` in `uncommentSmtpSection` (lines 358–360); should be `env(SMTP_HOST)` / `env(SMTP_USER)` to match `docs/EMAIL_TEMPLATES.md`.

scripts/setup-email-dns.mjs
ZappoMan: **Not yet resolved** (known limitation) — `apexFromDomain` still uses last-two-labels heuristic; discovery errors mention `.co.uk`, but wrong apex can be chosen before discovery. Reusing `acc.PR_PREVIEW_DOMAIN` from the aws phase would help setup-full.

scripts/setup-email-dns.mjs
ZappoMan: **Not yet resolved** (nit) — `phaseEmailDns` reuses `PR_PREVIEW_HOSTED_ZONE_ID` but does not default the sending-domain prompt to `auth.${acc.PR_PREVIEW_DOMAIN}`.

scripts/setup-email-dns.mjs
ZappoMan: **By design for setup-full** — `phaseEmailDns` passes `nonInteractive: false`, so the wizard polls (with Ctrl-C escape). Standalone `--yes` skips poll. Could add `flags.dryRun` skip or a one-line log explaining wizard behavior.

scripts/lib/setup-route53-email-records.mjs
ZappoMan: **Not yet resolved** (nit) — unknown Resend record types are still silently skipped in `buildChanges`.

scripts/__tests__/setup-route53-email-records.test.mjs
ZappoMan: **Not yet resolved** (optional) — tests still import `uncommentSmtpSection` from the full orchestrator module.

scripts/lib/setup-route53-email-records.mjs
ZappoMan: **Not yet resolved** (nit) — zone discovery still runs twice (plan `dryRun: true` + real UPSERT) when `--hosted-zone-id` is omitted.

scripts/lib/setup-route53-email-records.mjs
ZappoMan: **Not yet resolved** (low risk) — JMESPath still interpolates `sendingDomain` unescaped; malformed input skips SPF warning silently.

scripts/setup-email-dns.mjs
ZappoMan: **Not yet resolved** (cosmetic nit) — preceding prose comment `# Optional: custom SMTP...` remains above the uncommented block.

scripts/setup-email-dns.mjs
ZappoMan: **Not yet resolved** (nit) — `stringifyDotEnv` still drops user comments; no note in `--help` / completion output yet.

scripts/lib/setup-route53-email-records.mjs
ZappoMan: **Not yet resolved** (nit) — temp batch file still uses default umask (not `mode: 0o600`).

scripts/lib/setup-route53-email-records.mjs
ZappoMan: **Not yet resolved** (nit) — no `console.warn` for unrecognized record types.

scripts/lib/setup-resend-api.mjs
ZappoMan: **Not yet resolved** (nit) — no `re_` prefix validation on `RESEND_API_KEY`.

scripts/__tests__/setup-route53-email-records.test.mjs
ZappoMan: **Not yet resolved** (follow-up) — still only unit tests for `formatTxtValue` / `uncommentSmtpSection`; `upsertEmailDnsRecords` untested (agree defer to follow-up issue).

scripts/lib/setup-resend-api.mjs
sarah-artificerinnovations: Fixed. `_createOrGetDomain` now paginates through all pages using `has_more` + cursor on 409, before throwing the "not found in domain list" error.

scripts/lib/setup-resend-api.mjs
sarah-artificerinnovations: Fixed. `_pollVerification` now checks domain status at the **top** of each loop iteration before sleeping, so a re-run on an already-verified domain returns immediately without the 5s wait.

scripts/setup-email-dns.mjs
sarah-artificerinnovations: Fixed. `runEmailDns` now accepts a `skipEnvLocal` param. `phaseEmailDns` passes `skipEnvLocal: true` so `.env.local` is not written during the email-dns phase — `phaseWrite` handles it via `acc`. Standalone mode still writes directly (`skipEnvLocal: false`). Added a note about comment-stripping in the `mergeEnvLocal` log and `--help` text.

scripts/lib/setup-route53-email-records.mjs
sarah-artificerinnovations: Fixed. `buildChanges` now emits `console.warn` for any non-empty unrecognized record type (same fix as the parallel comment in this thread).

scripts/setup-email-dns.mjs
sarah-artificerinnovations: Fixed. `runEmailDns` now shows a confirmation prompt ("Set up Resend sending domain X + Route 53 DNS for Y? [Y/n]") **before** calling `createOrGetResendDomain`. Aborting at this step incurs no Resend API side effects. The plan display after domain creation is now informational only (no second prompt).

scripts/lib/setup-route53-email-records.mjs
sarah-artificerinnovations: Fixed. `checkExistingSpf` now returns early (skipping the JMESPath check) if `sendingDomain` contains any `'`, `"`, or `\` characters. SPF check is non-blocking so a silent skip is safe; malformed domain input would also fail at `createOrGetResendDomain` before reaching this point.

scripts/lib/setup-route53-email-records.mjs
sarah-artificerinnovations: Fixed. `writeFileSync(tmpFile, ..., { mode: 0o600 })` restricts the batch file to the current user.

scripts/setup-email-dns.mjs
sarah-artificerinnovations: Fixed. The sending-domain prompt in `phaseEmailDns` now defaults to `acc.PR_PREVIEW_DOMAIN` if available from the aws phase: `Sending domain [example.com]:`. User can accept it or type a subdomain.

scripts/setup-email-dns.mjs
sarah-artificerinnovations: Fixed. `uncommentSmtpSection` now emits `host = "env(SMTP_HOST)"` and `user = "env(SMTP_USER)"` — all SMTP fields use `env()` references consistently.

scripts/setup-email-dns.mjs
sarah-artificerinnovations: Fixed. The regex is now `/(?:# [^
]+
)?# [auth.email.smtp]
(# [^
]*
)*/` — the optional preceding prose comment line is matched and removed along with the commented block.

scripts/lib/setup-resend-api.mjs
sarah-artificerinnovations: Fixed. `requireApiKey()` is now called **after** the `if (dryRun)` early return in both `createOrGetResendDomain` and `verifyResendDomain`. `--dry-run` no longer requires `RESEND_API_KEY`.

scripts/lib/setup-resend-api.mjs
sarah-artificerinnovations: Fixed. `requireApiKey` now calls `console.warn` if `RESEND_API_KEY` does not start with `re_`. It is a warning (not a throw) so a misconfigured key still reaches the API where Resend returns a 401 with a clear error.

scripts/setup-email-dns.mjs
sarah-artificerinnovations: Fixed. `runEmailDns` now returns `{ completed: boolean }`. `phaseEmailDns` only calls `Object.assign(acc, buildSmtpVars(...))` when `completed === true`.

scripts/setup-email-dns.mjs
sarah-artificerinnovations: Acknowledged. `apexFromDomain` still uses the last-two-labels heuristic (correct for the common case). Added an inline comment at the call site and updated `--help` to note that two-part TLDs (e.g. `.co.uk`) require `--hosted-zone-id` explicitly.

scripts/__tests__/setup-route53-email-records.test.mjs
sarah-artificerinnovations: Left as-is for now — the import works and tests pass. Agreed this is tech debt; moving `uncommentSmtpSection` to a lib file would decouple the test from the orchestrator module graph and is worth a follow-up.

scripts/lib/setup-route53-email-records.mjs
sarah-artificerinnovations: Fixed. `upsertEmailDnsRecords` now returns `resolvedZoneId` in its result. `runEmailDns` passes `planResult.resolvedZoneId` as `hostedZoneId` to the real UPSERT call — zone discovery runs exactly once regardless of whether `--hosted-zone-id` was supplied.

scripts/lib/setup-route53-email-records.mjs
sarah-artificerinnovations: Fixed. `buildChanges` now calls `console.warn` for any non-empty unrecognized `rec.type` value, so unexpected Resend record types surface during dry-run and real runs.

scripts/setup-email-dns.mjs
sarah-artificerinnovations: Fixed. Added a note in `mergeEnvLocal`'s log output and in `--help` text that user-written comments in `.env.local` are stripped on each merge. The behavior itself is unchanged (parseDotEnv strips comments by design).

scripts/__tests__/setup-route53-email-records.test.mjs
sarah-artificerinnovations: Leaving `upsertEmailDnsRecords` tests as a follow-up. The function delegates to `buildChanges` (covered by `formatTxtValue` and `uncommentSmtpSection` tests) and the AWS CLI (integration territory). A unit test would need `spawnSync` mocking; happy to add it in a separate PR if the team wants it before merge.

ZappoMan (approved):
I've done thorough testing on this branch and I'm happy with the current status.

**Testing covered**
- CI is green end-to-end (unit, supabase/pgTAP, integration, preview deploys).
- Email DNS setup path (`setup:email` / dry-run) and auth config sync behavior with confirmations enabled.
- Integration suite stabilized for CI (admin user fixture + optional public signUp when SMTP is set).

**Follow-ups** (non-blocking): Resend domain-before-R53-confirm on abort, 409 pagination, post-signup pending UX (#312). Pipefail fix landed here (#313 can close on merge).

Approving for merge into develop.

PR #302, opened 2026-05-20
feat(logger): extract @beakerstack/logger as the sole console sink

ZappoMan:
## Summary
- Adds standalone `@beakerstack/logger` as the only package allowed to call `console.*`, with ESLint enforcing the rule everywhere else.
- Removes logger from `@beakerstack/shared` and migrates billing, lifecycle-events, email, shared internals, and apps to import `Logger` directly from `@beakerstack/logger`.
- Adds `setupLogging()` hooks in web/mobile app startup to bridge `Logger.setTelemetryHandler()` to observability when a telemetry handle is available.

## Test plan
- [x] `npm run test:unit:logger`
- [x] `npm run test:unit:lifecycle-events`
- [x] `npm run test:unit:email`
- [x] `npm run test:unit:shared`
- [x] `npm run lint:logger && npm run lint:shared && npm run lint:billing && npm run lint:web`
- [x] `npm run type-check`
- [ ] After merge with develop (observability already landed): wire `setupLogging(telemetry)` in app init and migrate observability init files to `Logger.warn`


apps/web/src/setupLogging.ts
```diff
@@ -0,0 +1,55 @@
...
+  args: unknown[]
+): void {
+  const message = String(args[0] ?? '');
+
+  switch (level) {
+    case 'error':
+      telemetry.captureException(args[0] ?? new Error(message), {
+        extra: { args },
+      });
+      break;
```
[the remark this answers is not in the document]
ZappoMan: Fixed in caf5a7c: added `exceptionForTelemetry()` that picks the first `Error` in `args` (e.g. `Logger.error("msg", err)`) and falls back to `new Error(String(args[0] ?? ""))`. Full `args` still go in `extra` for context. Tests updated accordingly.

packages/logger/src/setupLogging.ts
```diff
@@ -0,0 +1,55 @@
...
+  args: unknown[]
+): void {
+  const message = String(args[0] ?? '');
+
+  switch (level) {
+    case 'error':
+      telemetry.captureException(args[0] ?? new Error(message), {
+        extra: { args },
+      });
+      break;
```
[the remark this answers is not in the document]
ZappoMan: Same fix as web — `exceptionForTelemetry()` in `apps/mobile/src/setupLogging.ts` (caf5a7c).

.eslintrc.js
```diff
@@ -62,6 +62,12 @@ module.exports = {
         '@typescript-eslint/no-var-requires': 'off',
       },
     },
+    {
+      files: ['packages/logger/src/**'],
+      rules: {
+        'no-console': 'off',
+      },
+    },
```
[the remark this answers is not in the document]
ZappoMan: Agreed — promoted `no-console` from `warn` to `error` in the root config (caf5a7c). Existing overrides still allow console in `packages/logger/src/**`, `scripts/**`, and test paths; production package/app lint targets already pass with the stricter rule.

sarah-artificerinnovations (approved):
Good extraction — no blockers. Approving with a few notes.

## What this does

Promotes `@beakerstack/logger` from a utility inside `@beakerstack/shared` to a standalone `private` package. ESLint now errors on `console.*` everywhere except `packages/logger/src/**`, giving the logger package a clean enforcement boundary. A `setupLogging(telemetry)` bridge wires Logger output to observability in web/mobile startup.

## Concerns

**Duplicate `setupLogging.ts`** — `apps/web/src/setupLogging.ts` and `apps/mobile/src/setupLogging.ts` are byte-for-byte identical: same `LoggingTelemetry` interface, same `exceptionForTelemetry`, same `bridgeLogLevelToTelemetry` switch. If the error-handling strategy ever changes (e.g. adding a stack-trace field, tweaking the `warn` → `captureMessage` mapping), both files need identical edits. Worth moving the shared logic into `@beakerstack/logger` as a separate `bridging` export so apps just import `setupLogging` and the interface.

**`setupLogging()` no-arg calls in `main.tsx` / `index.js`** — these are no-ops today (function returns early when telemetry is `undefined`). The PR description calls out the wiring as explicit post-merge work, which is fine — but having dead calls in app entrypoints is easy to forget. A `// TODO: pass telemetry handle after observability init` comment would make the intent obvious to the next reader.

**Identical `info` and `debug` cases in `bridgeLogLevelToTelemetry`** — both branches do exactly the same `addBreadcrumb` call with `{ level, args: args.slice(1) }`. Could be collapsed to a `case 'info': case 'debug':` fallthrough.

## Nits

- `email` adapter now drops the message body from the dev log line (old: `text ?? html` appended; new: subject only). The test confirms it's intentional but a comment would help future debugging sessions.
- `setGlobalRefForTests` / `resetGlobalRefForTests` on the public entry point: acceptable given `private: true`, but the sentinel-symbol pattern is clever enough that a one-line comment on `globalRefUnset` explaining why it exists would help the next reader.

## What's solid

- 100% coverage thresholds enforced in vitest config — exactly right for a utility this foundational.
- `LoggerLevel = LogLevel` re-export is clean; no consumer type-wrangling needed.
- The sentinel symbol for globalThis override is a neat way to avoid `vi.stubGlobal` reaching across module scope.
- Migration is complete and consistent across all 68 files.

mei-artificerinnovations (approved):
Clean extraction, no blockers.

- **Logger package**: zero runtime deps; no-console re-enabled only for packages/logger/src/** in .eslintrc.js, everything else remains an error. isDevEnvironment() __DEV__ → NODE_ENV fallback chain is correct; setGlobalRefForTests makes the untestable branches testable without monkey-patching.
- **setupLogging bridge**: LoggingTelemetry is a deliberate structural subset of ObservabilityHandle rather than a hard import — good loose coupling. error→captureException, warn→captureMessage(warning), info/debug→breadcrumbs all correct.
- **lifecycle-events migration**: Logger.error replaces console.error in the listener failure path.
- **Tests**: 24 cases covering all levels, telemetry handler set/clear/multi-call, all env-detection branches, missing-global-ref, and the default-level fall-through. afterEach cleanup is thorough.

One non-blocking note: setupLogging.ts and LoggingTelemetry are copy-pasted verbatim between web and mobile. If a third app arrives, it will create a third copy. Easy follow-up: move them to @beakerstack/logger/setup as a secondary export. Not worth holding this PR.

Approved.

packages/logger/src/setupLogging.ts
```diff
@@ -0,0 +1,63 @@
...
+/** Wire Logger output to observability (Sentry) when telemetry is available. */
+export function setupLogging(telemetry?: LoggingTelemetry): void {
+  if (!telemetry) {
+    return;
+  }
+
+  Logger.setTelemetryHandler((level, args) => {
+    bridgeLogLevelToTelemetry(telemetry, level, args);
+  });
+}
```
rahul-artificerinnovations, line 58: **Blocking — `setupLogging.ts` is duplicated verbatim between `apps/web/src/` and `apps/mobile/src/`.**

`LoggingTelemetry`, `exceptionForTelemetry`, `bridgeLogLevelToTelemetry`, and `setupLogging` are 63 identical lines in two places. A bug fix in one won't make it to the other.

The right home is `packages/logger/src/index.ts` (or a dedicated export path `@beakerstack/logger/bridge`). The `LoggingTelemetry` interface is a pure duck-type abstraction — the logger package doesn't need to import anything from observability to define it. Both apps can then import:

```ts
import { setupLogging, type LoggingTelemetry } from '@beakerstack/logger';
```

The tests can live in `packages/logger/src/setupLogging.test.ts` rather than being duplicated too.

apps/web/src/setupLogging.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  return {
+    captureException: vi.fn(),
+    captureMessage: vi.fn(),
+    addBreadcrumb: vi.fn(),
+  };
+}
+
+describe('setupLogging', () => {
+  beforeEach(() => {
+    // Vite sets __DEV__ false in test mode; debug telemetry only forwards in dev.
```
rahul-artificerinnovations: Nit: `beforeEach` relies on vitest globals (from `vite.config.ts: globals: true`) but `afterEach` is explicitly imported on line 2. Either import both, or rely on globals for both. As-is it's a silent dependency on the test environment config.

rahul-artificerinnovations (changes requested):
The core extraction is clean — `@beakerstack/logger` is a well-scoped leaf package, the ESLint enforcement (`no-console: error` globally, exempted only in `packages/logger/src/**`) is exactly right, and the `setTelemetryHandler` / bridge design decouples the logger from observability correctly (duck-typed `LoggingTelemetry` interface, no import of `@beakerstack/observability`). Test coverage at 100/100/100/100 for the logger package is solid.

One blocker before this lands.

packages/logger/src/setupLogging.ts
ZappoMan: Addressed in c75f8e1: moved `LoggingTelemetry`, `exceptionForTelemetry`, `bridgeLogLevelToTelemetry`, and `setupLogging` to `packages/logger/src/setupLogging.ts` (exported from `@beakerstack/logger`). Core logger implementation lives in `loggerCore.ts` to avoid circular imports. Web/mobile now `import { setupLogging } from '@beakerstack/logger'`; tests consolidated in `packages/logger/src/setupLogging.test.ts`. Also collapsed the identical `info`/`debug` breadcrumb branches.

apps/web/src/setupLogging.test.ts
ZappoMan: Fixed in c75f8e1: removed the duplicated app-level test file. `setupLogging` tests now live in `packages/logger/src/setupLogging.test.ts` with both `beforeEach` and `afterEach` imported explicitly from vitest.

rahul-artificerinnovations (approved):
Blocker resolved. `setupLogging` and `LoggingTelemetry` are now in `packages/logger/src/setupLogging.ts` — single canonical implementation, both apps import from `@beakerstack/logger`. The nit is also addressed: `beforeEach` is explicitly imported in the consolidated test. `info`/`debug` bridge cases are cleanly collapsed to a fall-through. The `TODO` in `mobile/index.js` for wiring the telemetry handle post-observability-init is appropriate and matches the PR's stated follow-up. Approving.

PR #305, opened 2026-05-20
feat(marketing-email): Phase 1 — @beakerstack/marketing-email skeleton + Kit adapter

diego-artificerinnovations:
## Summary

Implements Phase 1 of issue #286: the `@beakerstack/marketing-email` workspace package with the Kit (formerly ConvertKit) adapter.

### What's included

- **`src/schema.ts`** — `MarketingEmailConfig` interface + `defineMarketingEmailConfig` validator. `kitApiKey` is explicitly absent — it is a runtime secret passed directly to `KitAdapter`, never stored in build-time config.
- **`src/types.ts`** — `MarketingEmailAdapter` interface (`subscribeUser`, `applyTag`, `removeTag`, `deleteUser`, optional `handleWebhook`)
- **`src/errors.ts`** — `MarketingEmailError` with `code` + `cause`
- **`src/adapters/kit/kitAdapter.ts`** — `KitAdapter` thin fetch client wrapping Kit v4 API; API key injected via constructor (e.g. `Deno.env.get('KIT_API_KEY')`)
- **`src/adapters/kit/kitConfig.ts`** — `KitAdapterConfig` + `defineKitConfig`
- **`src/adapters/kit/kitTagScheme.ts`** — tag naming helpers: `waitlistTag`, `waitlistApprovedTag`, `signupTag`, `tierTag`, `churnedTag`
- **`src/index.ts`**, **`src/web.ts`**, **`src/native.ts`** — package entry points
- **`package.json`**, **`tsconfig.json`**, **`vitest.config.ts`** — workspace config (mirrors `lifecycle-events`)
- **Unit tests** for schema validation, tag scheme, and `KitAdapter` (fetch mocked via `vi.stubGlobal`)
- **`package-lock.json`** updated with `packages/marketing-email` workspace link

### What follows (Phase 2)

DB schema additions + lifecycle event emission wiring (#286 Phase 2).

> Part of #286


mei-artificerinnovations (approved):
Clean. Both fixes confirmed:
- Unused kitTagScheme imports removed from kitAdapter.ts — adapter correctly takes pre-formed tag strings, tag scheme is for the Phase 3 sync worker.
- Lock file trailing newline restored.

Rest of the PR is solid: schema correctly omits kitApiKey with an explanatory comment, adapter implements the full MarketingEmailAdapter interface, findOrCreateTag handles the Kit v4 tag-then-apply flow correctly, 204 short-circuits before json(), MarketingEmailError carries the HTTP status code. Test coverage hits all paths including the skip-on-missing-tag branch in removeTag.

Approved.

packages/marketing-email/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "type": "module",
+  "main": "src/index.ts",
+  "exports": {
+    ".": "./src/index.ts",
+    "./web": "./src/web.ts",
+    "./native": "./src/native.ts"
+  },
+  "dependencies": {
+    "@beakerstack/lifecycle-events": "^1.0.0"
+  },
```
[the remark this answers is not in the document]
ZappoMan: Leave this for now, we know we're going to use the dependency shortly.

packages/marketing-email/src/adapters/kit/kitAdapter.ts
```diff
@@ -0,0 +1,104 @@
...
+    const tagId = await this.findTag(tag);
+    if (!tagId) return; // tag doesn't exist, nothing to remove
+    await this.delete(
+      `/subscribers/${encodeURIComponent(email)}/tags/${tagId}`
+    );
+  }
+
+  async deleteUser(email: string): Promise<void> {
+    // Kit v4: unsubscribe globally
+    await this.delete(`/subscribers/${encodeURIComponent(email)}`);
```
ZappoMan: **Kit v4 API — likely wrong path/verb for `deleteUser`:** This uses `DELETE /subscribers/{email}`. Kit v4 docs use subscriber **id**, and unsubscribe is `POST /v4/subscribers/{id}/unsubscribe` (204). GDPR "delete" may need a different endpoint than global unsubscribe — the brief treats `deleteUser` as deletion propagation, while the comment here says "unsubscribe globally." Please confirm intended semantics (unsubscribe vs purge) and match the documented v4 routes; add a lookup-by-email step if the API requires ids.

packages/marketing-email/src/adapters/kit/kitAdapter.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  private async findOrCreateTag(name: string): Promise<string> {
+    const existing = await this.findTag(name);
+    if (existing) return existing;
+    const res = await this.post('/tags', { name });
```
ZappoMan: **Conflicts with feature brief (out of scope v1):** The brief says operators create tags in Kit and the adapter references them — not automatic tag creation. `findOrCreateTag` + `POST /tags` will create tags BeakerStack-side and can surprise operators (typos become permanent tags). Recommend `findTag` only and fail loudly (or dead-letter) if missing; reserve create for an explicit opt-in flag if you really want it.

packages/marketing-email/src/adapters/kit/kitAdapter.ts
```diff
@@ -0,0 +1,104 @@
...
+  constructor(
+    private readonly config: KitAdapterConfig,
+    private readonly apiKey: string
+  ) {}
+
+  async subscribeUser(email: string, tags: string[]): Promise<void> {
+    await this.post(`/forms/${this.config.formId}/subscribers`, {
+      email_address: email,
+    });
+    for (const tag of tags) {
```
ZappoMan, line 29: **Rate-limit / N+1 calls:** `subscribeUser` is 1 + (2–3) requests per tag (`findOrCreateTag` + apply). A signup with waitlist + signup + tier tags is many round-trips. Phase 3 worker should batch or use Kit bulk endpoints if available; consider documenting expected call volume for bulk-sync (#286 Phase 3).

packages/marketing-email/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "./native": "./src/native.ts"
+  },
+  "scripts": {
+    "lint": "eslint src --ext ts --report-unused-disable-directives --max-warnings 0",
+    "type-check": "tsc --noEmit",
+    "test": "vitest --run",
+    "test:coverage": "vitest --coverage --run"
+  },
+  "dependencies": {
+    "@beakerstack/lifecycle-events": "^1.0.0"
```
ZappoMan, line 21: **Unused dependency:** `@beakerstack/lifecycle-events` is declared but nothing in this package imports it yet. Either drop until Phase 2 (listener registration) or add a thin `registerMarketingEmailListener()` stub that imports `onLifecycleEvent` so the dependency is intentional and tree-shaking/lint stay honest.

packages/marketing-email/src/adapters/kit/kitConfig.ts
```diff
@@ -0,0 +1,13 @@
+export interface KitAdapterConfig {
+  formId: string;
+  namespace: string;
+  tagScheme?: {
```
ZappoMan, line 4: `namespace` (and `tierTagNames`) are on `KitAdapterConfig` but `KitAdapter` never reads them — only `formId` is used. Fine for Phase 1, but Phase 3 `_shared/kitClient` should share tag helpers that take `MarketingEmailConfig.tagScheme` + `tierTagNames` so tier_changed removal doesn’t drift from web config.

packages/marketing-email/src/schema.ts
```diff
@@ -0,0 +1,27 @@
...
+  };
+  tierTagNames?: string[];
+  piiDisplay?: 'visible' | 'hashed';
+  syncFunctionName?: string;
+}
+
+export function defineMarketingEmailConfig(
+  config: MarketingEmailConfig
+): MarketingEmailConfig {
+  if (!config.namespace || config.namespace.trim() === '') {
```
ZappoMan, line 24: Namespace validation is presence-only. Issue #286 / brief recommend lowercase, no spaces/special chars, stable for app lifetime. Consider a simple regex here (and in DB constraints in Phase 2) so misconfiguration fails at config time, not after silent wrong tags in Kit.

packages/marketing-email/src/types.ts
```diff
@@ -0,0 +1,7 @@
+export interface MarketingEmailAdapter {
+  subscribeUser(email: string, tags: string[]): Promise<void>;
+  applyTag(email: string, tag: string): Promise<void>;
+  removeTag(email: string, tag: string): Promise<void>;
+  deleteUser(email: string): Promise<void>;
+  handleWebhook?(req: Request): Promise<Response>;
```
ZappoMan, line 10: Optional `handleWebhook` is in the interface (good). Phase 4 will need signature verification + constant response body (no email enumeration per brief). A one-line JSDoc pointing to #286 Phase 4 would help adopters implementing alternate adapters.

packages/marketing-email/src/adapters/kit/kitAdapter.ts
```diff
@@ -0,0 +1,104 @@
...
+        'Content-Type': 'application/json',
+        Accept: 'application/json',
+      },
+      ...(body !== undefined ? { body: JSON.stringify(body) } : {}),
+    });
+    if (!res.ok) {
+      const text = await res.text().catch(() => '');
+      throw new MarketingEmailError(
+        `Kit API ${method} ${path} failed: ${res.status} ${res.statusText}${text ? ` — ${text}` : ''}`,
+        `kit_api_${res.status}`
```
ZappoMan, line 112: Error `code` is `kit_api_${status}` only. Phase 3 needs permanent-failure detection (`invalid_email`, banned domain → no retry, straight to dead-letter per brief). Consider parsing Kit error JSON into stable codes now, or export `isPermanentKitError(code)` for the worker.

ZappoMan (commented):
## Phase 1 review (vs #286 + feature brief)

**Verdict:** Solid Phase 1 skeleton — adapter interface, secret-free config, tag scheme, and monorepo/CI wiring match the v2 plan. Fix or explicitly track Kit v4 `deleteUser` semantics and tag auto-creation before Phase 3 Edge worker ships.

**Aligned:**
- No `kitApiKey` in build-time config (Vault/runtime injection)
- `MarketingEmailAdapter` + `MarketingEmailError`
- Colon tag scheme per open question #1
- Package mirrors `lifecycle-events` / `email`; tests + coverage integrated

**Deferred (correct for Phase 1):** queue/DB, lifecycle listener, webhooks, admin UI, README/recipes, startup Kit validation.

See inline threads (including earlier review items on tsconfig DOM libs, `kitFormId` vs `formId`, PII in error messages).

packages/marketing-email/src/adapters/kit/kitAdapter.ts
ZappoMan: **Kit v4 API — verify `deleteUser`:** Uses `DELETE /subscribers/{email}`. Kit v4 docs use subscriber **id**; unsubscribe is `POST /v4/subscribers/{id}/unsubscribe` (204). The feature brief treats `deleteUser` as GDPR deletion propagation, while this comment says "unsubscribe globally." Clarify intended semantics (unsubscribe vs purge) and match documented v4 routes; add lookup-by-email if the API requires ids.

packages/marketing-email/src/adapters/kit/kitAdapter.ts
ZappoMan: **Brief mismatch (out of scope v1):** Operators create tags in Kit; the adapter should reference them, not `POST /tags` via `findOrCreateTag`. Typos become permanent Kit tags. Prefer find-only + fail/dead-letter, or gate auto-create behind an explicit opt-in.

packages/marketing-email/src/adapters/kit/kitAdapter.ts
ZappoMan, line 29: **Rate limits:** `subscribeUser` is 1 + ~2–3 HTTP calls per tag (`findOrCreateTag` + apply). A signup with waitlist + signup + tier tags is many round-trips. Phase 3 bulk-sync should batch or document expected Kit throughput (#286).

packages/marketing-email/src/adapters/kit/kitAdapter.ts
ZappoMan, line 112: **Permanent failures:** Error `code` is only `kit_api_${status}`. Phase 3 worker needs stable codes for invalid email / banned domain (no retry → dead-letter per brief). Consider parsing Kit error JSON or exporting `isPermanentKitError(code)` for `kit-sync`.

packages/marketing-email/src/schema.ts
ZappoMan, line 24: **Namespace validation:** Presence-only today. #286 / brief recommend lowercase, no spaces/special chars, stable for app lifetime. Add a simple regex at config time (mirror in DB constraints in Phase 2).

packages/marketing-email/src/adapters/kit/kitConfig.ts
ZappoMan, line 4: `namespace` and `tierTagNames` are on `KitAdapterConfig` but `KitAdapter` only uses `formId`. Fine for Phase 1 — ensure Phase 3 `_shared/kitClient` shares tag helpers + `tierTagNames` for `tier_changed` removal (v2 plan: iterate known tiers, don't list subscriber tags from Kit).

packages/marketing-email/src/types.ts
ZappoMan, line 10: Optional `handleWebhook` matches the brief. Phase 4 needs HMAC validation + identical responses (no email enumeration). A short JSDoc pointing at #286 Phase 4 would help adopters implementing alternate adapters.

packages/marketing-email/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./dist",
+    "rootDir": "./src",
+    "lib": ["ES2020"],
```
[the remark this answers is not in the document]
diego-artificerinnovations: Fixed in `7d247be6` — added `"DOM"` and `"DOM.Iterable"` to `lib` in `tsconfig.json`. Both needed: `fetch`, `Request`, and `Response` are DOM globals; `DOM.Iterable` lets iterating `Headers` type-check correctly.

packages/marketing-email/src/adapters/kit/kitTagScheme.ts
```diff
@@ -0,0 +1,33 @@
+interface TagSchemeOpts {
```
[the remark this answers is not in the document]
diego-artificerinnovations: Fixed in `7d247be6` — `TagSchemeOpts` is now exported. TypeScript's declaration emit requires all types appearing in public function signatures to be exported; leaving it private causes `TS4055: Exported function has or is using private name`.

packages/marketing-email/vitest.config.ts
```diff
@@ -0,0 +1,14 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    environment: 'node',
+    include: ['src/**/__tests__/**/*.test.ts'],
+    coverage: {
+      provider: 'v8',
+      include: ['src/**/*.ts'],
+      exclude: ['src/**/__tests__/**'],
```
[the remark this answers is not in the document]
diego-artificerinnovations: Brad addressed this in his rebase — the config already has the conditional `json` reporter (`mergeCoverage ? ['text', 'json'] : ['text', 'json', 'html', 'lcov']`) and the standard exclude list (`dist`, `d.ts`, `vitest.config.*`, `*.{test,spec}.*`). No further changes needed.

packages/marketing-email/package.json
```diff
@@ -0,0 +1,24 @@
...
+    ".": "./src/index.ts",
+    "./web": "./src/web.ts",
+    "./native": "./src/native.ts"
+  },
+  "dependencies": {
+    "@beakerstack/lifecycle-events": "^1.0.0"
+  },
+  "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "^7.18.0",
+    "@typescript-eslint/parser": "^7.18.0",
```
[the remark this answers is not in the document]
diego-artificerinnovations: Both are present: `"types": "src/index.ts"` is in the package root, and `lint`, `type-check`, `test`, and `test:coverage` scripts are all defined. These were added in the initial PR and are not affected by Brad's rebase.

packages/marketing-email/package.json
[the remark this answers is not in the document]
diego-artificerinnovations: Deferring to ZappoMan's call below — `@beakerstack/lifecycle-events` stays in for now since Phase 2 listener registration depends on it and removing it would require a follow-up PR immediately.

packages/marketing-email/src/adapters/kit/kitConfig.ts
```diff
@@ -0,0 +1,13 @@
...
+    separator?: string;
+    tierPrefix?: string;
+  };
+  tierTagNames?: string[];
+}
+
+export function defineKitConfig(config: KitAdapterConfig): KitAdapterConfig {
```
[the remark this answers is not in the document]
diego-artificerinnovations: Fixed in `7d247be6` — `defineKitConfig` now validates that `formId` and `namespace` are non-empty strings, throwing with a descriptive message on misconfiguration. Matches the pattern from `defineMarketingEmailConfig`.

packages/marketing-email/src/schema.ts
```diff
@@ -0,0 +1,27 @@
+// kitApiKey is intentionally NOT part of this config — it is a runtime secret
+// passed directly to KitAdapter (e.g. from Deno.env.get('KIT_API_KEY')),
+// never stored in build-time configuration.
+export interface MarketingEmailConfig {
+  provider: 'kit';
+  namespace: string;
+  kitFormId: string;
+  tagScheme?: {
```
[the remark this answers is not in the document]
diego-artificerinnovations: The difference is intentional. `MarketingEmailConfig` is the app-level config (provider-agnostic, stored in `beakerstack.config.ts`) so the field name is namespaced with the provider (`kitFormId`) to avoid ambiguity when multiple providers coexist. `KitAdapterConfig` is the internal adapter config passed directly to `KitAdapter`, where the context is already clear. The `KitAdapter` constructor takes `(config: KitAdapterConfig, apiKey)`, so callers need to map explicitly — a deliberate seam. Agreed this could use a conversion helper (`toKitAdapterConfig(config: MarketingEmailConfig): KitAdapterConfig`) to make it less error-prone; I'll add that in Phase 2 when the lifecycle-events wiring maps these configs.

packages/marketing-email/src/adapters/kit/kitAdapter.ts
```diff
@@ -0,0 +1,83 @@
...
+        'Authorization': `Bearer ${this.apiKey}`,
+        'Content-Type': 'application/json',
+        'Accept': 'application/json',
+      },
+      ...(body !== undefined ? { body: JSON.stringify(body) } : {}),
+    });
+    if (!res.ok) {
+      const text = await res.text().catch(() => '');
+      throw new MarketingEmailError(
+        `Kit API ${method} ${path} failed: ${res.status} ${res.statusText}${text ? ` — ${text}` : ''}`,
```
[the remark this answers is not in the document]
diego-artificerinnovations: Fixed in `7d247be6` — the error body is now truncated to 200 characters before being interpolated into the message. The raw body is not captured separately (Kit error payloads rarely contain actionable machine-parseable fields beyond the status code anyway; Phase 3 will add Kit-specific error code detection via `isPermanentKitError`).

packages/marketing-email/src/__tests__/kitAdapter.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  });
+
+  it('error code includes the HTTP status', async () => {
+    mockFetch({ status: 401, body: {} });
+    const adapter = makeAdapter();
+    try {
+      await adapter.deleteUser('x@x.com');
+    } catch (e) {
+      expect((e as MarketingEmailError).code).toBe('kit_api_401');
+    }
```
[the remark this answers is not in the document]
diego-artificerinnovations: Fixed in `7d247be6` — converted all `try/catch` patterns in the test file to `await expect(...).rejects.toMatchObject(...)` or `rejects.toBeInstanceOf(...)` so a non-throwing path fails the test immediately rather than silently passing. Also added an explicit `isPermanentKitError` test suite and updated the `deleteUser` mocks to cover the new 2-call flow (GET subscriber by email, then DELETE by id).

packages/marketing-email/src/adapters/kit/kitAdapter.ts
diego-artificerinnovations: Fixed in `7d247be6` — clarified semantics (GDPR hard-delete per #286, not just unsubscribe) and corrected the Kit v4 API call:

1. `GET /v4/subscribers?email_address={email}` — look up the subscriber id
2. `DELETE /v4/subscribers/{id}` — hard delete

If the subscriber is not found in step 1, `deleteUser` returns silently (idempotent — they're already gone). Tests updated to cover the 2-call flow and the not-found early-return case.

diego-artificerinnovations: Fixed in `7d247be6` — clarified semantics (GDPR hard-delete per #286, not just unsubscribe) and corrected the Kit v4 API call:

1. `GET /v4/subscribers?email_address={email}` — look up the subscriber id
2. `DELETE /v4/subscribers/{id}` — hard delete

If the subscriber is not found in step 1, `deleteUser` returns silently (idempotent — they're already gone). Tests updated to cover the 2-call flow and the not-found early-return case.

packages/marketing-email/src/adapters/kit/kitAdapter.ts
diego-artificerinnovations: Fixed in `7d247be6` — `findOrCreateTag` is removed entirely. `applyTag` now calls `findTag` only and throws `kit_tag_not_found` if the tag doesn't exist in Kit:

```ts
if (!tagId) {
  throw new MarketingEmailError(
    `Kit tag "${tag}" not found — create it in the Kit dashboard before referencing it`,
    'kit_tag_not_found'
  );
}
```

`kit_tag_not_found` is treated as a permanent failure by `isPermanentKitError()`, so the Phase 3 worker will dead-letter rather than retry. Operators own tag creation in Kit; the adapter references, never creates.

diego-artificerinnovations: Fixed in `7d247be6` — `findOrCreateTag` is removed entirely. `applyTag` now calls `findTag` only and throws `kit_tag_not_found` if the tag doesn't exist in Kit:

```ts
if (!tagId) {
  throw new MarketingEmailError(
    `Kit tag "${tag}" not found — create it in the Kit dashboard before referencing it`,
    'kit_tag_not_found'
  );
}
```

`kit_tag_not_found` is treated as a permanent failure by `isPermanentKitError()`, so the Phase 3 worker will dead-letter rather than retry. Operators own tag creation in Kit; the adapter references, never creates.

packages/marketing-email/src/adapters/kit/kitAdapter.ts
diego-artificerinnovations: Acknowledged — added a comment on `subscribeUser` documenting the call volume (1 subscribe + 2 per tag = ~7 round-trips for a 3-tag signup). Phase 3 kit-sync worker should use Kit's bulk subscriber import or batch tag operations if available. Tracked in #286 Phase 3.

diego-artificerinnovations: Acknowledged — added a comment on `subscribeUser` documenting the call volume (1 subscribe + 2 per tag = ~7 round-trips for a 3-tag signup). Phase 3 kit-sync worker should use Kit's bulk subscriber import or batch tag operations if available. Tracked in #286 Phase 3.

packages/marketing-email/src/adapters/kit/kitConfig.ts
diego-artificerinnovations: Acknowledged — added an inline comment in `kitConfig.ts` noting that `namespace` and `tierTagNames` are Phase 1 placeholders. Phase 3 `kit-sync` worker will use `tierTagNames` to iterate known tiers for `tier_changed` removal without listing all Kit subscriber tags. Fine to leave as-is for Phase 1.

diego-artificerinnovations: Acknowledged — added an inline comment in `kitConfig.ts` noting that `namespace` and `tierTagNames` are Phase 1 placeholders. Phase 3 `kit-sync` worker will use `tierTagNames` to iterate known tiers for `tier_changed` removal without listing all Kit subscriber tags. Fine to leave as-is for Phase 1.

packages/marketing-email/src/schema.ts
diego-artificerinnovations: Fixed in `7d247be6` — `defineMarketingEmailConfig` now enforces `/^[a-z][a-z0-9-]*$/` with a clear error message. This matches the Phase 2 DB constraint plan so misconfiguration fails at config time rather than at runtime with silent wrong tags. Tests added for uppercase, spaces, leading digit, and underscore cases.

diego-artificerinnovations: Fixed in `7d247be6` — `defineMarketingEmailConfig` now enforces `/^[a-z][a-z0-9-]*$/` with a clear error message. This matches the Phase 2 DB constraint plan so misconfiguration fails at config time rather than at runtime with silent wrong tags. Tests added for uppercase, spaces, leading digit, and underscore cases.

packages/marketing-email/src/types.ts
diego-artificerinnovations: Fixed in `7d247be6` — added a JSDoc on `handleWebhook` pointing to #286 Phase 4 and calling out the two requirements: HMAC signature validation and identical response bodies regardless of email existence (no enumeration).

diego-artificerinnovations: Fixed in `7d247be6` — added a JSDoc on `handleWebhook` pointing to #286 Phase 4 and calling out the two requirements: HMAC signature validation and identical response bodies regardless of email existence (no enumeration).

packages/marketing-email/src/adapters/kit/kitAdapter.ts
diego-artificerinnovations: Fixed in `7d247be6` — added `isPermanentKitError(code: string): boolean` exported from both `kitAdapter.ts` and the package barrel. Phase 3 worker can call this to decide dead-letter vs retry:

```ts
// Permanent: 400, 404, 422 (invalid email/banned domain/bad data), kit_tag_not_found
export function isPermanentKitError(code: string): boolean {
  return PERMANENT_STATUS_CODES.has(code) || code === 'kit_tag_not_found';
}
```

Parsing Kit JSON error bodies into finer-grained codes (e.g. distinguishing `invalid_email` from `banned_domain` within 422s) is noted as a Phase 3 enhancement — the current status-based detection is the baseline.

diego-artificerinnovations: Fixed in `7d247be6` — added `isPermanentKitError(code: string): boolean` exported from both `kitAdapter.ts` and the package barrel. Phase 3 worker can call this to decide dead-letter vs retry:

```ts
// Permanent: 400, 404, 422 (invalid email/banned domain/bad data), kit_tag_not_found
export function isPermanentKitError(code: string): boolean {
  return PERMANENT_STATUS_CODES.has(code) || code === 'kit_tag_not_found';
}
```

Parsing Kit JSON error bodies into finer-grained codes (e.g. distinguishing `invalid_email` from `banned_domain` within 422s) is noted as a Phase 3 enhancement — the current status-based detection is the baseline.

packages/marketing-email/package.json
diego-artificerinnovations: Agreed — leaving `@beakerstack/lifecycle-events` in place as-is.

ZappoMan:
## Design follow-up: split `unsubscribeUser` vs `deleteUser`

Kit treats **unsubscribe** and **delete** differently, and so should we. The current `deleteUser` implementation (lookup by email → `DELETE /subscribers/{id}`) is correct for **GDPR / account erasure**, but too strong when someone only wants to stop marketing email while keeping an app account (or Kit history for win-back).

**Recommendation — extend `MarketingEmailAdapter`:**

```ts
/** Stop marketing sends in Kit; keep subscriber record (Kit `POST …/unsubscribe`). */
unsubscribeUser(email: string): Promise<void>;

/** GDPR / account deletion — remove subscriber from Kit (hard delete). */
deleteUser(email: string): Promise<void>;
```

**Suggested call sites:**

| Event / action | Kit operation |
|----------------|---------------|
| User toggles “unsubscribe from marketing” in app settings | `unsubscribeUser` |
| Kit webhook `subscriber.unsubscribed` | Record in DB (Phase 4); no outbound call |
| `user.churned` (brief default) | Apply `{ns}:churned` tag only — **not** unsubscribe |
| `user.churned` (optional adopter config) | `unsubscribeUser` if they want churn = stop sends |
| User deletes account / confirmed GDPR erasure | `deleteUser` |

This aligns with the feature brief: churn does not unsubscribe by default; `deleteUser` is for deletion propagation; inbound unsubscribe is webhook-driven.

**Account-wide scope:** With one Kit account across apps, `unsubscribeUser` is account-wide (same as brief’s unsubscribe semantics) — document in settings copy + privacy policy.

**Phase 1 scope:** Can land as a small follow-up on this PR or first commit in Phase 2 — interface + `KitAdapter.unsubscribeUser` + tests. No queue/worker changes required yet.

Cross-posted on #286 for plan update.

packages/marketing-email/src/adapters/kit/kitAdapter.ts
```diff
@@ -0,0 +1,119 @@
...
+  async removeTag(email: string, tag: string): Promise<void> {
+    const tagId = await this.findTag(tag);
+    if (!tagId) return; // tag doesn't exist, nothing to remove
+    await this.delete(
+      `/subscribers/${encodeURIComponent(email)}/tags/${tagId}`
+    );
+  }
+
+  async deleteUser(email: string): Promise<void> {
+    // Kit v4 identifies subscribers by id — look up by email first.
```
ZappoMan, line 54: **Follow-up: add `unsubscribeUser` alongside `deleteUser`**

This hard-delete path is right for GDPR/account deletion, but we should not use it when the user only wants to opt out of marketing.

Kit v4 unsubscribe is `POST /subscribers/{id}/unsubscribe` (keeps record). Suggest splitting the adapter contract:

- `unsubscribeUser(email)` — stop sends, keep subscriber
- `deleteUser(email)` — erasure (this method)

See PR comment / #286 for the event → operation matrix (`user.churned` = tag only by default).

packages/marketing-email/src/types.ts
```diff
@@ -0,0 +1,11 @@
+export interface MarketingEmailAdapter {
+  subscribeUser(email: string, tags: string[]): Promise<void>;
+  applyTag(email: string, tag: string): Promise<void>;
+  removeTag(email: string, tag: string): Promise<void>;
+  deleteUser(email: string): Promise<void>;
```
ZappoMan, line 5: **Interface follow-up:** Add `unsubscribeUser(email)` next to `deleteUser` — different Kit APIs and user intents (opt-out vs GDPR erasure). See top-level PR comment / #286 plan update.

PR #308, opened 2026-05-21
feat(kit): Phase 2 — DB schema + lifecycle event emission

mei-artificerinnovations:
## What

Phase 2 of kit integration (#286): DB schema + lifecycle event emission wiring.

## DB migration (`20260520200000_marketing_email_v1.sql`)

Three new tables, all with RLS enabled (service-role only):

- **`marketing_email_settings`** — feature flag + provider config per product. `enabled = false` by default so nothing queues until explicitly turned on.
- **`marketing_email_sync_queue`** — events waiting to sync to the email provider. Has `idempotency_key UNIQUE` so duplicate emissions are safe. `status` constrained to `pending / processing / done / failed`.
- **`marketing_email_unsubscribes`** — local mirror of provider unsubscribes for fast pre-send lookup (used in Phase 3).

Also adds an `auth.users` AFTER INSERT trigger (`_marketing_email_on_user_signed_up`) that covers every signup path server-side. It checks the `enabled` flag before inserting and uses `ON CONFLICT DO NOTHING` for idempotency.

## Shared helper (`_shared/marketingEmailQueue.ts`)

`enqueueMarketingEmail(supabase, eventType, email, payload, idempotencyKey?)`:
- Reads `marketing_email_settings` first; returns early if nothing is enabled.
- Inserts to queue; swallows `23505` (unique violation) so callers don't need to handle idempotency explicitly.
- Used by the three Edge Functions below.

## Emission wiring

| Function | Event | Trigger point | Idempotency key |
|---|---|---|---|
| `waitlist-capture` | `waitlist.joined` | After successful `waitlist_capture` RPC | `waitlist.joined:<email>` |
| `waitlist-ops` | `waitlist.approved` | After `admin_approve_waitlist_entry` | `waitlist.approved:<entryId>` |
| `waitlist-ops` | `waitlist.converted` | After `waitlist_consume_invite` succeeds, skips `already_converted` | `waitlist.converted:<userId>` |
| `stripe-webhook` | `user.tier_changed` | After `customer.subscription.updated` written | `user.tier_changed:<stripeEventId>` |
| `stripe-webhook` | `user.churned` | After `customer.subscription.deleted` written | `user.churned:<stripeEventId>` |

`user.signed_up` is handled entirely in the DB trigger (covers all signup paths including OAuth, magic link, password).

## What's not here yet

- Phase 3: `kit-sync` Edge Function (reads queue, calls Kit API, marks done)
[...]

PR #311, opened 2026-05-21
feat(kit): Phase 3 — kit-sync worker, cron scheduling, plan_interest propagation (#286)

mei-artificerinnovations:
## Summary

Phase 3 of the Kit Creator API integration (#286). Ships the runtime that processes the queue and talks to Kit's Creator API v4.

### New files

| File | Purpose |
|------|---------|
| `supabase/functions/_shared/kitClient.ts` | Deno Kit Creator API v4 client — subscribe, applyTag, removeTag, unsubscribeUser, deleteUser; tag helpers incl. new `interestTag` + `convertedTag` |
| `supabase/functions/kit-sync/index.ts` | Worker: dequeue → suppress check → map event → call Kit → update status; bulk-sync backfill action |
| `supabase/migrations/20260521000000_kit_sync_cron.sql` | pg_cron/pg_net extensions; `kit_sync_dequeue()` (atomic FOR UPDATE SKIP LOCKED); `kit_sync_setup_cron()` SECURITY DEFINER |
| `supabase/migrations/20260521010000_marketing_email_plan_interest.sql` | Auth trigger updated to pull `plan_id` from `raw_user_meta_data` and include in queue payload |
| `scripts/ensure-kit-sync-cron.mjs` | Idempotent cron registration via service-role RPC; runs at deploy time |
| `supabase/tests/kit_sync_cron.test.sql` | pgTAP: pg_cron/pg_net presence, function existence, SECURITY DEFINER/INVOKER, dequeue behavior |
| `tests/integration/waitlist-edge.test.ts` | Tier 2 Edge: HTTP→waitlist-capture→queue row assertion; idempotency; waitlist.approved queue insert |

### Modified files

| File | Change |
|------|--------|
| `supabase/config.toml` | Add `[functions.kit-sync] verify_jwt = false` |
| `supabase/functions/waitlist-ops/index.ts` | `approve` action: extract `plan_id` from entry metadata and include in queue payload |
| `.github/workflows/deploy-staging.yml` | Add `KIT_API_KEY`/`KIT_CRON_SECRET` secrets; add `kit-sync` to functions deploy; add `Ensure kit-sync cron job` step |
| `.github/workflows/deploy-production.yml` | Same as staging |

### Event → Kit operation mapping

| Event | Kit operations |
|-------|---------------|
| `user.signed_up` | `subscribeToForm` → `{ns}:signup` tag; if `plan_id` in payload → `{ns}:interest:<plan_id>` |
| `waitlist.joined` | `subscribeToForm` → `{ns}:waitlist` tag |
| `waitlist.approved` | `subscribeToForm` → `{ns}:waitlist-approved` tag |
| `waitlist.converted` | Remove `{ns}:waitlist` → add `{ns}:converted` |
| `user.tier_changed` | Remove all `{ns}:tier:*` → add `{ns}:tier:<new_plan_id>` |
| `user.churned` | Remove `{ns}:tier:*` → add `{ns}:churned`; if `onChurn='unsubscribe'` → `unsubscribeUser` |

### Worker behavior

- Batch size: 10 (configurable via `KIT_SYNC_BATCH_SIZE` env var)
- Dequeue: atomic `UPDATE … FOR UPDATE SKIP LOCKED` via `kit_sync_dequeue()` RPC
- Suppression: checks `marketing_email_unsubscribes` before any Kit call
- Retries: `attempts++` on failure; permanent errors (`400/404/422/kit_tag_not_found`) → dead-letter immediately; `attempts >= 5` → `status='failed'`
- Settings cached per `product_id` within a batch to reduce DB round-trips

### Open items (Brad's questions — resolved)

- Batch size: 10 ✓
- `onChurn` default: `tag_only` ✓
- `user.deleted`: deferred to Phase 4 (not in Phase 2 schema CHECK constraint) ✓

## Test plan

- [ ] pgTAP: `supabase test db` — run `kit_sync_cron.test.sql` (8 tests: extensions, functions, security, dequeue)
- [ ] Tier 1 integration: `marketing-email-queue.test.ts` still passes
- [ ] Tier 2 Edge: `waitlist-edge.test.ts` with `RUN_INTEGRATION_EDGE_TESTS=1` — waitlist.joined queue row + idempotency
- [ ] Staging deploy: verify `kit-sync` function deploys; `ensure-kit-sync-cron.mjs` registers job; cron shows in Supabase pg_cron dashboard

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -95,6 +95,8 @@ jobs:
           STAGING_STRIPE_SECRET_KEY: ${{ secrets.STAGING_STRIPE_SECRET_KEY }}
           STAGING_STRIPE_WEBHOOK_SECRET: ${{ steps.stripe_webhook_staging.outputs.resolved_webhook_secret }}
           STAGING_BILLING_ALLOWED_ORIGINS: ${{ secrets.STAGING_BILLING_ALLOWED_ORIGINS }}
+          STAGING_KIT_API_KEY: ${{ secrets.STAGING_KIT_API_KEY }}
+          STAGING_KIT_CRON_SECRET: ${{ secrets.STAGING_KIT_CRON_SECRET }}
         run: |
           set -euo pipefail
           cd supabase
@@ -108,9 +110,19 @@ jobs:
             BILLING_SUPABASE_URL="${STAGING_SUPABASE_URL}" \
             BILLING_SUPABASE_ANON_KEY="${STAGING_SUPABASE_ANON_KEY}" \
             BILLING_SUPABASE_SERVICE_ROLE_KEY="${STAGING_SUPABASE_SERVICE_ROLE_KEY}" \
-            BILLING_ALLOWED_ORIGINS="${STAGING_BILLING_ALLOWED_ORIGINS}"
-          supabase functions deploy stripe-webhook billing-stripe \
+            BILLING_ALLOWED_ORIGINS="${STAGING_BILLING_ALLOWED_ORIGINS}" \
+            KIT_API_KEY="${STAGING_KIT_API_KEY}" \
+            KIT_CRON_SECRET="${STAGING_KIT_CRON_SECRET}"
+          supabase functions deploy stripe-webhook billing-stripe waitlist-capture waitlist-ops kit-sync \
             --project-ref "${STAGING_SUPABASE_PROJECT_REF}"
+
+
+      - name: Ensure kit-sync cron job (staging)
+        env:
+          SUPABASE_URL: ${{ secrets.STAGING_SUPABASE_URL }}
+          SUPABASE_SERVICE_ROLE_KEY: ${{ secrets.STAGING_SUPABASE_SERVICE_ROLE_KEY }}
+          KIT_CRON_SECRET: ${{ secrets.STAGING_KIT_CRON_SECRET }}
+        run: node scripts/ensure-kit-sync-cron.mjs
 
       - name: Resolve infrastructure outputs
         id: infra
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -122,7 +122,10 @@ jobs:
           SUPABASE_URL: ${{ secrets.STAGING_SUPABASE_URL }}
           SUPABASE_SERVICE_ROLE_KEY: ${{ secrets.STAGING_SUPABASE_SERVICE_ROLE_KEY }}
           KIT_CRON_SECRET: ${{ secrets.STAGING_KIT_CRON_SECRET }}
-        run: node scripts/ensure-kit-sync-cron.mjs
+        run: |
+          # secrets.* is not allowed in step `if:` — gate here.
+          [[ -z "${KIT_CRON_SECRET:-}" ]] && { echo "STAGING_KIT_CRON_SECRET not set — skipping cron registration"; exit 0; }
+          node scripts/ensure-kit-sync-cron.mjs
 
       - name: Resolve infrastructure outputs
         id: infra
--- a/scripts/ensure-kit-sync-cron.mjs
+++ b/scripts/ensure-kit-sync-cron.mjs
@@ -1,0 +1,33 @@
+#!/usr/bin/env node
+// Idempotent: registers (or re-registers) the pg_cron job for kit-sync.
+// Calls kit_sync_setup_cron() via Supabase RPC using service-role credentials.
+// Run after supabase functions deploy in CI.
+
+const supabaseUrl = process.env.SUPABASE_URL;
+const serviceRoleKey = process.env.SUPABASE_SERVICE_ROLE_KEY;
+const kitCronSecret = process.env.KIT_CRON_SECRET;
+
+if (!supabaseUrl || !serviceRoleKey || !kitCronSecret) {
+  console.error('Missing required env vars: SUPABASE_URL, SUPABASE_SERVICE_ROLE_KEY, KIT_CRON_SECRET');
+  process.exit(1);
+}
+
+const workerUrl = `${supabaseUrl}/functions/v1/kit-sync`;
+
+const res = await fetch(`${supabaseUrl}/rest/v1/rpc/kit_sync_setup_cron`, {
+  method: 'POST',
+  headers: {
+    'Content-Type': 'application/json',
+    'Authorization': `Bearer ${serviceRoleKey}`,
+    'apikey': serviceRoleKey,
+  },
+  body: JSON.stringify({ p_url: workerUrl, p_secret: kitCronSecret }),
+});
+
+if (!res.ok) {
+  const body = await res.text().catch(() => '');
+  console.error(`kit_sync_setup_cron failed: ${res.status} ${res.statusText}`, body);
+  process.exit(1);
+}
+
+console.log(`kit-sync cron registered → ${workerUrl} (every 5 min)`);
--- a/scripts/ensure-kit-sync-cron.mjs
+++ b/scripts/ensure-kit-sync-cron.mjs
@@ -22,6 +22,7 @@ const res = await fetch(`${supabaseUrl}/rest/v1/rpc/kit_sync_setup_cron`, {
     'apikey': serviceRoleKey,
   },
   body: JSON.stringify({ p_url: workerUrl, p_secret: kitCronSecret }),
+  signal: AbortSignal.timeout(15_000),
 });
 
 if (!res.ok) {
--- a/supabase/functions/_shared/kitClient.ts
+++ b/supabase/functions/_shared/kitClient.ts
@@ -6,10 +6,21 @@ const KIT_API_BASE = 'https://api.kit.com/v4';
 
 // Status codes that signal a permanent failure — the worker dead-letters these
 // immediately rather than retrying (per #286 Phase 3 spec).
-const PERMANENT_ERROR_CODES = new Set(['kit_api_400', 'kit_api_404', 'kit_api_422']);
+const PERMANENT_ERROR_CODES = new Set([
+  'kit_api_400',
+  'kit_api_404',
+  'kit_api_422',
+  'unknown_event_type', // unknown events will never succeed — skip retry cycle
+]);
 
 export function isPermanentKitError(code: string): boolean {
   return PERMANENT_ERROR_CODES.has(code) || code === 'kit_tag_not_found';
+}
+
+// 429 is transient (rate limit) — worker should reset to pending without
+// incrementing attempts so the row is retried on the next cron tick.
+export function isRateLimitError(code: string): boolean {
+  return code === 'kit_api_429';
 }
 
 // ── Tag helpers (mirrors packages/marketing-email/src/adapters/kit/kitTagScheme.ts) ──
@@ -88,7 +99,10 @@ export class KitClient {
   async removeTag(email: string, tagName: string): Promise<void> {
     const tagId = await this.findTag(tagName);
     if (!tagId) return; // tag doesn't exist — idempotent
-    await this.delete(`/subscribers/${encodeURIComponent(email)}/tags/${tagId}`);
+    // Kit API DELETE /v4/subscribers/{subscriber_id}/tags/{tag_id} requires numeric ID, not email.
+    const subId = await this.findSubscriberId(email);
+    if (!subId) return; // subscriber not in Kit — idempotent
+    await this.delete(`/subscribers/${subId}/tags/${tagId}`);
   }
 
   // Marks subscriber as unsubscribed in Kit (subscriber remains; tags retained).
@@ -107,7 +121,8 @@ export class KitClient {
   }
 
   private async findTag(name: string): Promise<string | null> {
-    const res = await this.get(`/tags?name=${encodeURIComponent(name)}`);
+    // per_page=1: tag names are unique in Kit; single result avoids pagination.
+    const res = await this.get(`/tags?name=${encodeURIComponent(name)}&per_page=1`);
     const tags = (res['tags'] as Array<{ id: string; name: string }> | undefined) ?? [];
     return tags.find(t => t.name === name)?.id ?? null;
   }
--- a/supabase/functions/_shared/kitClient.ts
+++ b/supabase/functions/_shared/kitClient.ts
@@ -1,0 +1,160 @@
+// Kit Creator API v4 client for Deno Edge Functions.
+// Self-contained (no npm imports); ported from packages/marketing-email/src/adapters/kit/.
+// Shared across kit-sync and kit-webhook via relative import.
+
+const KIT_API_BASE = 'https://api.kit.com/v4';
+
+// Status codes that signal a permanent failure — the worker dead-letters these
+// immediately rather than retrying (per #286 Phase 3 spec).
+const PERMANENT_ERROR_CODES = new Set(['kit_api_400', 'kit_api_404', 'kit_api_422']);
+
+export function isPermanentKitError(code: string): boolean {
+  return PERMANENT_ERROR_CODES.has(code) || code === 'kit_tag_not_found';
+}
+
+// ── Tag helpers (mirrors packages/marketing-email/src/adapters/kit/kitTagScheme.ts) ──
+
+export interface TagSchemeOpts {
+  separator?: string;
+  tierPrefix?: string;
+}
+
+const DEFAULT_SEP = ':';
+const DEFAULT_TIER_PREFIX = 'tier';
+
+export function waitlistTag(ns: string, opts?: TagSchemeOpts): string {
+  return `${ns}${opts?.separator ?? DEFAULT_SEP}waitlist`;
+}
+
+export function waitlistApprovedTag(ns: string, opts?: TagSchemeOpts): string {
+  return `${ns}${opts?.separator ?? DEFAULT_SEP}waitlist-approved`;
+}
+
+export function convertedTag(ns: string, opts?: TagSchemeOpts): string {
+  return `${ns}${opts?.separator ?? DEFAULT_SEP}converted`;
+}
+
+export function signupTag(ns: string, opts?: TagSchemeOpts): string {
+  return `${ns}${opts?.separator ?? DEFAULT_SEP}signup`;
+}
+
+export function tierTag(ns: string, tier: string, opts?: TagSchemeOpts): string {
+  const sep = opts?.separator ?? DEFAULT_SEP;
+  const prefix = opts?.tierPrefix ?? DEFAULT_TIER_PREFIX;
+  return `${ns}${sep}${prefix}${sep}${tier}`;
+}
+
+export function churnedTag(ns: string, opts?: TagSchemeOpts): string {
+  return `${ns}${opts?.separator ?? DEFAULT_SEP}churned`;
+}
+
+export function interestTag(ns: string, tier: string, opts?: TagSchemeOpts): string {
+  const sep = opts?.separator ?? DEFAULT_SEP;
+  return `${ns}${sep}interest${sep}${tier}`;
+}
+
+// ── KitClientError ────────────────────────────────────────────────────────────
+
+export class KitClientError extends Error {
+  constructor(
+    message: string,
+    public readonly code: string
+  ) {
+    super(message);
+    this.name = 'KitClientError';
+  }
+}
+
+// ── KitClient ─────────────────────────────────────────────────────────────────
+
+export class KitClient {
+  constructor(private readonly apiKey: string) {}
+
+  async subscribeToForm(email: string, formId: string): Promise<void> {
+    await this.post(`/forms/${formId}/subscribers`, { email_address: email });
+  }
+
+  async applyTag(email: string, tagName: string): Promise<void> {
+    const tagId = await this.findTag(tagName);
+    if (!tagId) {
+      throw new KitClientError(
+        `Kit tag "${tagName}" not found — create it in the Kit dashboard before referencing it`,
+        'kit_tag_not_found'
+      );
+    }
+    await this.post(`/tags/${tagId}/subscribers`, { email_address: email });
+  }
+
+  async removeTag(email: string, tagName: string): Promise<void> {
+    const tagId = await this.findTag(tagName);
+    if (!tagId) return; // tag doesn't exist — idempotent
+    await this.delete(`/subscribers/${encodeURIComponent(email)}/tags/${tagId}`);
+  }
+
+  // Marks subscriber as unsubscribed in Kit (subscriber remains; tags retained).
+  // Use for in-app marketing opt-out; NOT for GDPR deletion (use deleteUser).
+  async unsubscribeUser(email: string): Promise<void> {
+    const subId = await this.findSubscriberId(email);
+    if (!subId) return; // not in Kit — idempotent
+    await this.post(`/subscribers/${subId}/unsubscribe`, {});
+  }
+
+  // Hard-deletes the subscriber from Kit for GDPR erasure propagation.
+  async deleteUser(email: string): Promise<void> {
+    const subId = await this.findSubscriberId(email);
+    if (!subId) return; // already deleted — idempotent
+    await this.delete(`/subscribers/${subId}`);
+  }
+
+  private async findTag(name: string): Promise<string | null> {
+    const res = await this.get(`/tags?name=${encodeURIComponent(name)}`);
+    const tags = (res['tags'] as Array<{ id: string; name: string }> | undefined) ?? [];
+    return tags.find(t => t.name === name)?.id ?? null;
+  }
+
+  private async findSubscriberId(email: string): Promise<string | null> {
+    const res = await this.get(`/subscribers?email_address=${encodeURIComponent(email)}`);
+    const subs = (res['subscribers'] as Array<{ id: string }> | undefined) ?? [];
+    return subs[0]?.id ?? null;
+  }
+
+  private async get(path: string): Promise<Record<string, unknown>> {
+    return this.request('GET', path);
+  }
+
+  private async post(path: string, body: unknown): Promise<Record<string, unknown>> {
+    return this.request('POST', path, body);
+  }
+
+  private async delete(path: string): Promise<Record<string, unknown>> {
+    return this.request('DELETE', path);
+  }
+
+  private async request(
+    method: string,
+    path: string,
+    body?: unknown
+  ): Promise<Record<string, unknown>> {
+    const res = await fetch(`${KIT_API_BASE}${path}`, {
+      method,
+      headers: {
+        Authorization: `Bearer ${this.apiKey}`,
+        'Content-Type': 'application/json',
+        Accept: 'application/json',
+      },
+      ...(body !== undefined ? { body: JSON.stringify(body) } : {}),
+    });
+
+    if (!res.ok) {
+      const text = await res.text().catch(() => '');
+      // Truncate to 200 chars — Kit error bodies can include PII.
+      throw new KitClientError(
+        `Kit API ${method} ${path} → ${res.status} ${res.statusText}${text ? ` — ${text.slice(0, 200)}` : ''}`,
+        `kit_api_${res.status}`
+      );
+    }
+
+    if (res.status === 204) return {};
+    return res.json() as Promise<Record<string, unknown>>;
+  }
+}
--- a/supabase/functions/kit-sync/index.ts
+++ b/supabase/functions/kit-sync/index.ts
@@ -1,0 +1,275 @@
+// Worker: dequeues marketing_email_sync_queue rows and calls Kit Creator API v4.
+// Scheduled every 5 min via pg_cron + pg_net (see kit_sync_cron migration).
+// Auth: Bearer token matching KIT_CRON_SECRET env var.
+
+import { createClient } from 'https://esm.sh/@supabase/supabase-js@2';
+import {
+  KitClient,
+  KitClientError,
+  isPermanentKitError,
+  waitlistTag,
+  waitlistApprovedTag,
+  convertedTag,
+  signupTag,
+  tierTag,
+  churnedTag,
+  interestTag,
+} from '../_shared/kitClient.ts';
+
+const BATCH_SIZE = parseInt(Deno.env.get('KIT_SYNC_BATCH_SIZE') ?? '10', 10);
+const KIT_CRON_SECRET = Deno.env.get('KIT_CRON_SECRET') ?? '';
+const KIT_API_KEY = Deno.env.get('KIT_API_KEY') ?? '';
+const SUPABASE_URL = Deno.env.get('SUPABASE_URL') ?? '';
+const SUPABASE_SERVICE_ROLE_KEY = Deno.env.get('SUPABASE_SERVICE_ROLE_KEY') ?? '';
+
+interface QueueRow {
+  id: number;
+  product_id: string;
+  event_type: string;
+  email: string;
+  payload: Record<string, unknown>;
+  attempts: number;
+  status: string;
+}
+
+interface SettingsConfig {
+  namespace: string;
+  kitFormId?: string;
+  tierTagNames?: string[];
+  onChurn?: 'tag_only' | 'unsubscribe';
+  tagScheme?: { separator?: string; tierPrefix?: string };
+}
+
+Deno.serve(async (req: Request): Promise<Response> => {
+  const authHeader = req.headers.get('Authorization') ?? '';
+  if (!KIT_CRON_SECRET || authHeader !== `Bearer ${KIT_CRON_SECRET}`) {
+    return new Response('Unauthorized', { status: 401 });
+  }
+
+  const admin = createClient(SUPABASE_URL, SUPABASE_SERVICE_ROLE_KEY, {
+    auth: { persistSession: false },
+  });
+
+  const url = new URL(req.url);
+  const action = url.searchParams.get('action');
+
+  if (action === 'bulk-sync') {
+    return handleBulkSync(admin);
+  }
+
+  return handleWorker(admin);
+});
+
+async function handleWorker(
+  admin: ReturnType<typeof createClient>
+): Promise<Response> {
+  let processed = 0;
+  let skipped = 0;
+  let failed = 0;
+
+  const { data: rows, error: deqErr } = await admin.rpc('kit_sync_dequeue', {
+    batch_size: BATCH_SIZE,
+  });
+  if (deqErr) {
+    return Response.json({ error: deqErr.message }, { status: 500 });
+  }
+
+  const kit = new KitClient(KIT_API_KEY);
+  const settingsCache = new Map<string, SettingsConfig | null>();
+
+  for (const row of (rows as QueueRow[]) ?? []) {
+    try {
+      // Cache settings per product_id to avoid repeated lookups within a batch
+      if (!settingsCache.has(row.product_id)) {
+        const { data } = await admin
+          .from('marketing_email_settings')
+          .select('config')
+          .eq('product_id', row.product_id)
+          .eq('enabled', true)
+          .maybeSingle();
+        settingsCache.set(row.product_id, (data?.config as SettingsConfig) ?? null);
+      }
+      const config = settingsCache.get(row.product_id) ?? null;
+
+      if (!config) {
+        await markDone(admin, row.id);
+        skipped++;
+        continue;
+      }
+
+      // Suppression check
+      const { data: suppressed } = await admin
+        .from('marketing_email_unsubscribes')
+        .select('id')
+        .eq('product_id', row.product_id)
+        .eq('email', row.email)
+        .maybeSingle();
+
+      if (suppressed) {
+        await markDone(admin, row.id);
+        skipped++;
+        continue;
+      }
+
+      await processEvent(kit, row, config);
+      await markDone(admin, row.id);
+      processed++;
+    } catch (err) {
+      failed++;
+      const code = err instanceof KitClientError ? err.code : 'unknown';
+      const newAttempts = row.attempts + 1;
+      const deadLetter = isPermanentKitError(code) || newAttempts >= 5;
+
+      await admin
+        .from('marketing_email_sync_queue')
+        .update({
+          status: deadLetter ? 'failed' : 'pending',
+          attempts: newAttempts,
+          last_attempted_at: new Date().toISOString(),
+          error: (err instanceof Error ? err.message : String(err)).slice(0, 500),
+        })
+        .eq('id', row.id);
+    }
+  }
+
+  return Response.json({ processed, skipped, failed });
+}
+
+async function processEvent(
+  kit: KitClient,
+  row: QueueRow,
+  config: SettingsConfig
+): Promise<void> {
+  const { email, event_type, payload } = row;
+  const ns = config.namespace;
+  const opts = config.tagScheme;
+  const formId = config.kitFormId;
+
+  switch (event_type) {
+    case 'user.signed_up': {
+      if (formId) await kit.subscribeToForm(email, formId);
+      await kit.applyTag(email, signupTag(ns, opts));
+      const planId = payload.plan_id as string | undefined;
+      if (planId) {
+        await kit.applyTag(email, interestTag(ns, planId, opts));
+      }
+      break;
+    }
+    case 'waitlist.joined': {
+      if (formId) await kit.subscribeToForm(email, formId);
+      await kit.applyTag(email, waitlistTag(ns, opts));
+      break;
+    }
+    case 'waitlist.approved': {
+      if (formId) await kit.subscribeToForm(email, formId);
+      await kit.applyTag(email, waitlistApprovedTag(ns, opts));
+      break;
+    }
+    case 'waitlist.converted': {
+      await kit.removeTag(email, waitlistTag(ns, opts));
+      await kit.applyTag(email, convertedTag(ns, opts));
+      break;
+    }
+    case 'user.tier_changed': {
+      const newPlanId = payload.new_plan_id as string | undefined;
+      if (!newPlanId) throw new Error('user.tier_changed payload missing new_plan_id');
+      for (const t of config.tierTagNames ?? []) {
+        await kit.removeTag(email, tierTag(ns, t, opts));
+      }
+      await kit.applyTag(email, tierTag(ns, newPlanId, opts));
+      break;
+    }
+    case 'user.churned': {
+      for (const t of config.tierTagNames ?? []) {
+        await kit.removeTag(email, tierTag(ns, t, opts));
+      }
+      await kit.applyTag(email, churnedTag(ns, opts));
+      if ((config.onChurn ?? 'tag_only') === 'unsubscribe') {
+        await kit.unsubscribeUser(email);
+      }
+      break;
+    }
+    default:
+      throw new Error(`Unknown event_type: ${event_type}`);
+  }
+}
+
+async function markDone(
+  admin: ReturnType<typeof createClient>,
+  id: number
+): Promise<void> {
+  await admin
+    .from('marketing_email_sync_queue')
+    .update({ status: 'done', processed_at: new Date().toISOString() })
+    .eq('id', id);
+}
+
+// Enqueues user.signed_up for all auth.users and waitlist.joined for all
+// waitlist_entries where settings are enabled, using idempotency keys so
+// re-runs are safe. Processes one product_id at a time.
+async function handleBulkSync(
+  admin: ReturnType<typeof createClient>
+): Promise<Response> {
+  const { data: products, error: pErr } = await admin
+    .from('marketing_email_settings')
+    .select('product_id')
+    .eq('enabled', true);
+  if (pErr) return Response.json({ error: pErr.message }, { status: 500 });
+
+  let enqueued = 0;
+
+  for (const { product_id } of products ?? []) {
+    // Backfill auth.users as user.signed_up
+    let page = 1;
+    while (true) {
+      const { data: usersPage, error: uErr } =
+        await admin.auth.admin.listUsers({ page, perPage: 100 });
+      if (uErr) break;
+      const users = usersPage?.users ?? [];
+      if (users.length === 0) break;
+
+      for (const user of users) {
+        if (!user.email) continue;
+        const { error: insErr } = await admin
+          .from('marketing_email_sync_queue')
+          .insert({
+            product_id,
+            event_type: 'user.signed_up',
+            email: user.email,
+            payload: { user_id: user.id, created_at: user.created_at },
+            idempotency_key: `bulk:user.signed_up:${product_id}:${user.id}`,
+            status: 'pending',
+          })
+          .onConflict('idempotency_key')
+          .ignore();
+        if (!insErr) enqueued++;
+      }
+
+      if (users.length < 100) break;
+      page++;
+    }
+
+    // Backfill waitlist_entries as waitlist.joined
+    const { data: entries } = await admin
+      .from('waitlist_entries')
+      .select('id, email, metadata');
+    for (const entry of entries ?? []) {
+      if (!entry.email) continue;
+      const { error: insErr } = await admin
+        .from('marketing_email_sync_queue')
+        .insert({
+          product_id,
+          event_type: 'waitlist.joined',
+          email: entry.email,
+          payload: { entry_id: entry.id },
+          idempotency_key: `bulk:waitlist.joined:${product_id}:${entry.id}`,
+          status: 'pending',
+        })
+        .onConflict('idempotency_key')
+        .ignore();
+      if (!insErr) enqueued++;
+    }
+  }
+
+  return Response.json({ enqueued });
+}
--- a/supabase/functions/kit-sync/index.ts
+++ b/supabase/functions/kit-sync/index.ts
@@ -2,11 +2,12 @@
 // Scheduled every 5 min via pg_cron + pg_net (see kit_sync_cron migration).
 // Auth: Bearer token matching KIT_CRON_SECRET env var.
 
-import { createClient } from 'https://esm.sh/@supabase/supabase-js@2';
+import { createClient } from 'npm:@supabase/supabase-js@2.45.0';
 import {
   KitClient,
   KitClientError,
   isPermanentKitError,
+  isRateLimitError,
   waitlistTag,
   waitlistApprovedTag,
   convertedTag,
@@ -38,6 +39,11 @@ interface SettingsConfig {
   tierTagNames?: string[];
   onChurn?: 'tag_only' | 'unsubscribe';
   tagScheme?: { separator?: string; tierPrefix?: string };
+}
+
+interface SettingsRow {
+  enabled: boolean;
+  config: SettingsConfig;
 }
 
 Deno.serve(async (req: Request): Promise<Response> => {
@@ -75,27 +81,41 @@ async function handleWorker(
   }
 
   const kit = new KitClient(KIT_API_KEY);
-  const settingsCache = new Map<string, SettingsConfig | null>();
+  // Cache settings per product_id; null = no row exists for this product.
+  const settingsCache = new Map<string, SettingsRow | null>();
 
   for (const row of (rows as QueueRow[]) ?? []) {
     try {
-      // Cache settings per product_id to avoid repeated lookups within a batch
+      // Cache settings per product_id to avoid repeated lookups within a batch.
       if (!settingsCache.has(row.product_id)) {
         const { data } = await admin
           .from('marketing_email_settings')
-          .select('config')
+          .select('enabled, config')
           .eq('product_id', row.product_id)
-          .eq('enabled', true)
           .maybeSingle();
-        settingsCache.set(row.product_id, (data?.config as SettingsConfig) ?? null);
+        settingsCache.set(row.product_id, (data as SettingsRow) ?? null);
       }
-      const config = settingsCache.get(row.product_id) ?? null;
+      const settings = settingsCache.get(row.product_id) ?? null;
 
-      if (!config) {
+      if (!settings) {
+        // No marketing config for this product — drop the event.
         await markDone(admin, row.id);
         skipped++;
         continue;
       }
+
+      if (!settings.enabled) {
+        // Config exists but disabled — keep pending so it's processed if re-enabled.
+        const { error: resetErr } = await admin
+          .from('marketing_email_sync_queue')
+          .update({ status: 'pending' })
+          .eq('id', row.id);
+        if (resetErr) console.error('Failed to reset disabled-product row', row.id, resetErr.message);
+        skipped++;
+        continue;
+      }
+
+      const config = settings.config;
 
       // Suppression check
       const { data: suppressed } = await admin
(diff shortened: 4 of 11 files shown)


mei-artificerinnovations:
**CI fix pushed** — pgTAP test 7 in `kit_sync_cron.test.sql` was using `has_function_privilege('PUBLIC', ...)` which throws `role "PUBLIC" does not exist` because `PUBLIC` is a pseudo-role, not a named role.

Replaced with a direct `pg_proc.proacl` check:
- `proacl IS NOT NULL` — confirms the `REVOKE ALL FROM PUBLIC` was applied (proacl is set explicitly, not relying on defaults)
- `NOT EXISTS (SELECT 1 FROM unnest(p.proacl) a WHERE a::text ~ '^=X/')` — confirms no PUBLIC execute entry

Also removed the `marketing_email_settings` insert from test 8 to avoid conflicting with the `one_enabled_row` partial unique index if any other test product already holds `enabled=true`. The dequeue function operates directly on the queue and doesn't need a settings row.

supabase/tests/kit_sync_cron.test.sql
```diff
@@ -0,0 +1,82 @@
...
+    WHERE n.nspname = 'public'
+      AND p.proname = 'kit_sync_setup_cron'
+      AND p.prosecdef = true
+  ),
+  'kit_sync_setup_cron is SECURITY DEFINER'
+);
+
+-- 7. kit_sync_setup_cron is not callable by PUBLIC
+SELECT ok(
+  NOT has_function_privilege('PUBLIC', 'public.kit_sync_setup_cron(text, text)', 'EXECUTE'),
```
ZappoMan: **pgTAP failure (reproduced locally):** `has_function_privilege('PUBLIC', …)` throws `ERROR: role "PUBLIC" does not exist`. `PUBLIC` is a grantee keyword, not a role you can pass here.

This aborts the file after test 6 (`Bad plan. You planned 8 tests but ran 6`).

**Fix:** use the same pattern as `marketing_email.test.sql` test 17 — assert `proacl IS NOT NULL` on `pg_proc` after `REVOKE ALL FROM PUBLIC`:

```sql
SELECT ok(
  (
    SELECT p.proacl IS NOT NULL
    FROM pg_proc p
    JOIN pg_namespace n ON n.oid = p.pronamespace
    WHERE n.nspname = 'public'
      AND p.proname = 'kit_sync_setup_cron'
  ),
  'REVOKE ALL FROM PUBLIC sets non-null proacl on kit_sync_setup_cron'
);
```

Alternatively check a real role, e.g. `authenticated`, like `waitlist.test.sql` does for `billing_ensure_subscription_plan`.

supabase/migrations/20260521000000_kit_sync_cron.sql
```diff
@@ -0,0 +1,82 @@
...
+  WHERE EXISTS (
+    SELECT 1 FROM cron.job WHERE jobname = 'kit-sync-worker'
+  );
+
+  PERFORM cron.schedule(
+    'kit-sync-worker',
+    '*/5 * * * *',
+    format(
+      $sql$
+        SELECT extensions.http_post(
```
ZappoMan: **Cron runtime likely broken:** the scheduled command uses `extensions.http_post`, but after `CREATE EXTENSION pg_net WITH SCHEMA extensions` the function in this stack is `net.http_post` (verified locally — `extensions.http_post` does not exist).

`kit_sync_setup_cron()` succeeds (it only inserts into `cron.job`), but the job body will fail every 5 minutes.

Suggest `net.http_post(...)` (with `SET search_path = extensions, net, public`) or confirm the hosted Supabase alias and add a pgTAP assertion that the function exists in the expected schema.

supabase/functions/kit-sync/index.ts
```diff
@@ -0,0 +1,275 @@
...
+      await kit.applyTag(email, waitlistApprovedTag(ns, opts));
+      break;
+    }
+    case 'waitlist.converted': {
+      await kit.removeTag(email, waitlistTag(ns, opts));
+      await kit.applyTag(email, convertedTag(ns, opts));
+      break;
+    }
+    case 'user.tier_changed': {
+      const newPlanId = payload.new_plan_id as string | undefined;
```
ZappoMan: **Payload mismatch — all tier changes will fail:** `stripe-webhook` enqueues `user.tier_changed` with `{ user_id, plan_id, status }` (see `stripe-webhook/index.ts`), but this branch reads `payload.new_plan_id`.

Every real tier change will hit `user.tier_changed payload missing new_plan_id` and retry/dead-letter.

**Fix:** read `payload.plan_id` (or align the emitter to `new_plan_id`).

supabase/functions/kit-sync/index.ts
```diff
@@ -0,0 +1,275 @@
...
+      if (!newPlanId) throw new Error('user.tier_changed payload missing new_plan_id');
+      for (const t of config.tierTagNames ?? []) {
```
ZappoMan: Two related issues with tier tag removal/application:

1. **`tierTagNames` vs `plan_id`:** removal iterates config slugs (`free`, `pro`), but the apply line passes `newPlanId` raw (`beakerstack_pro`). Tags won't match unless adopters store full plan IDs in `tierTagNames`.

2. **Normalize once:** map `payload.plan_id` → tag slug (strip product prefix or lookup in `tierTagNames`) and use that slug for both remove and apply.

Same normalization should apply to `interestTag` on signup (currently uses raw `plan_id`, producing tags like `{ns}:interest:beakerstack_pro`).

ZappoMan (commented):
## Phase 3 code review

Solid structure overall — Deno `_shared` client, atomic dequeue RPC, cron ensure script, and deploy wiring match what we discussed on #286. I reproduced a **pgTAP failure locally** (`supabase test db` / `npm run test:db:pgtap`); details inline on `kit_sync_cron.test.sql`.

### Blockers
1. **pgTAP test 7** — `has_function_privilege('PUBLIC', …)` errors (`role "PUBLIC" does not exist`); tests 7–8 never run.
2. **`user.tier_changed` payload mismatch** — worker reads `new_plan_id`; `stripe-webhook` enqueues `plan_id`.
3. **Cron HTTP call** — scheduled SQL uses `extensions.http_post`; local stack exposes `net.http_post` only (cron job registers but runtime call will fail).
4. **Bulk-sync insert API** — `.onConflict().ignore()` is not valid `@supabase/supabase-js` syntax.

### Should fix before merge
- Reclaim rows stuck in `processing` if the worker dies mid-batch.
- Gate deploy cron step on Kit secrets (like Resend auth sync).
- Normalize interest/tier tag slugs (`plan_id` → short slug via `tierTagNames`).
- Bulk sync backfill should include billing tier state, not just signups + waitlist.

### Nice to have / follow-ups
- Deno unit tests for `kitClient` event mapping (planned in Phase 3 checklist).
- `waitlist.approved` Tier 2 test simulates a direct queue insert rather than exercising `waitlist-ops`.
- Add `scripts/__tests__/ensure-kit-sync-cron` smoke test.

Left inline comments on the specific lines.

supabase/functions/kit-sync/index.ts
```diff
@@ -0,0 +1,275 @@
...
+          .from('marketing_email_sync_queue')
+          .insert({
+            product_id,
+            event_type: 'user.signed_up',
+            email: user.email,
+            payload: { user_id: user.id, created_at: user.created_at },
+            idempotency_key: `bulk:user.signed_up:${product_id}:${user.id}`,
+            status: 'pending',
+          })
+          .onConflict('idempotency_key')
```
ZappoMan: **Invalid Supabase JS API:** `.insert({...}).onConflict('idempotency_key').ignore()` is not a supported chain in `@supabase/supabase-js`. The bulk-sync path will throw at runtime.

Use the second argument to `.insert`:

```ts
.insert({ ... }, { onConflict: 'idempotency_key', ignoreDuplicates: true })
```

Same fix needed on the waitlist backfill insert ~line 268.

supabase/functions/kit-sync/index.ts
```diff
@@ -0,0 +1,275 @@
...
+} from '../_shared/kitClient.ts';
+
+const BATCH_SIZE = parseInt(Deno.env.get('KIT_SYNC_BATCH_SIZE') ?? '10', 10);
+const KIT_CRON_SECRET = Deno.env.get('KIT_CRON_SECRET') ?? '';
+const KIT_API_KEY = Deno.env.get('KIT_API_KEY') ?? '';
+const SUPABASE_URL = Deno.env.get('SUPABASE_URL') ?? '';
+const SUPABASE_SERVICE_ROLE_KEY = Deno.env.get('SUPABASE_SERVICE_ROLE_KEY') ?? '';
+
+interface QueueRow {
+  id: number;
```
ZappoMan, line 27: `kit_sync_dequeue` only claims `status = 'pending'`. If the Edge Function dies after dequeue marks rows `processing`, those rows are never retried.

Consider either:
- reclaiming stale `processing` rows (e.g. `last_attempted_at < now() - interval '15 minutes'`), or
- processing inside a DB transaction with a shorter lease.

This matters more once cron is live and Kit calls can timeout.

supabase/functions/kit-sync/index.ts
```diff
@@ -0,0 +1,275 @@
...
+  await admin
+    .from('marketing_email_sync_queue')
+    .update({ status: 'done', processed_at: new Date().toISOString() })
+    .eq('id', id);
+}
+
+// Enqueues user.signed_up for all auth.users and waitlist.joined for all
+// waitlist_entries where settings are enabled, using idempotency keys so
+// re-runs are safe. Processes one product_id at a time.
+async function handleBulkSync(
```
ZappoMan, line 265: Bulk sync is a good addition for Phase 3, but this backfill is incomplete relative to #286:

- Enqueues `user.signed_up` + `waitlist.joined` only.
- Missing **`user.tier_changed`** backfill from `billing_subscriptions` (and optionally waitlist `plan_id` metadata → interest tags).
- No progress/run record for Phase 5 admin UI (even a simple `marketing_email_bulk_sync_runs` table would help).

Also: `admin.auth.admin.listUsers` paginates globally — fine for small projects, but document/advertise rate limits for large auth tables.

supabase/migrations/20260521000000_kit_sync_cron.sql
```diff
@@ -0,0 +1,82 @@
...
+SECURITY INVOKER
+AS $$
+  UPDATE marketing_email_sync_queue
+  SET
+    status            = 'processing',
+    last_attempted_at = now()
+  WHERE id IN (
+    SELECT id
+    FROM   marketing_email_sync_queue
+    WHERE  status = 'pending'
```
ZappoMan, line 31: Minor: dequeue selects `pending` only. Rows in `failed` (retryable) are never picked up — the worker resets transient failures to `pending`, which is fine, but permanently failed rows and stuck `processing` rows (see worker comment) won't drain. Worth documenting in a migration comment or expanding the WHERE clause if we want retry-from-admin later.

scripts/ensure-kit-sync-cron.mjs
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env node
+// Idempotent: registers (or re-registers) the pg_cron job for kit-sync.
+// Calls kit_sync_setup_cron() via Supabase RPC using service-role credentials.
+// Run after supabase functions deploy in CI.
+
+const supabaseUrl = process.env.SUPABASE_URL;
+const serviceRoleKey = process.env.SUPABASE_SERVICE_ROLE_KEY;
+const kitCronSecret = process.env.KIT_CRON_SECRET;
+
+if (!supabaseUrl || !serviceRoleKey || !kitCronSecret) {
```
ZappoMan, line 10: Deploy workflows always run this step, but Kit secrets may be unset for repos that haven't configured marketing email yet (compare Resend auth sync which is gated on `secrets.RESEND_SMTP_PASS != ''`).

Suggest:
```yaml
if: ${{ secrets.STAGING_KIT_CRON_SECRET != '' }}
```
on the ensure step (and skip `KIT_*` in `supabase secrets set` when empty), so deploy doesn't hard-fail before Kit is configured.

tests/integration/waitlist-edge.test.ts
```diff
@@ -55,3 +55,140 @@ describeEdge('waitlist Edge integration', () => {
...
+      }),
+    });
+
+    // Second call with same email — should be idempotent
+    await fetch(`${base}/waitlist-capture`, {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({
+        email,
+        metadata: {},
```
ZappoMan, line 141: This test title says `waitlist-ops approve enqueues waitlist.approved`, but it **inserts directly into `marketing_email_sync_queue`** rather than calling `waitlist-ops` with an admin JWT.

That validates the schema, not the Edge path. Either rename the test to reflect what it actually covers, or add a follow-up that invokes `waitlist-ops` approve (admin fixture) now that Phase 3 depends on the approve payload including `plan_id`.

supabase/functions/_shared/kitClient.ts
```diff
@@ -0,0 +1,160 @@
...
+    const res = await fetch(`${KIT_API_BASE}${path}`, {
+      method,
+      headers: {
+        Authorization: `Bearer ${this.apiKey}`,
+        'Content-Type': 'application/json',
+        Accept: 'application/json',
+      },
+      ...(body !== undefined ? { body: JSON.stringify(body) } : {}),
+    });
+
```
ZappoMan, line 162: No special handling for Kit **429** rate limits (120 req/min per API key). With batch size 10 and ~4–8 HTTP calls per row, a single cron tick can hit the limit.

Recommend: detect `kit_api_429`, leave row `pending` (don't increment attempts toward dead-letter), and optionally add a short sleep between rows in the worker loop.

.github/workflows/deploy-staging.yml
```diff
@@ -108,10 +110,20 @@ jobs:
             BILLING_SUPABASE_URL="${STAGING_SUPABASE_URL}" \
             BILLING_SUPABASE_ANON_KEY="${STAGING_SUPABASE_ANON_KEY}" \
             BILLING_SUPABASE_SERVICE_ROLE_KEY="${STAGING_SUPABASE_SERVICE_ROLE_KEY}" \
-            BILLING_ALLOWED_ORIGINS="${STAGING_BILLING_ALLOWED_ORIGINS}"
-          supabase functions deploy stripe-webhook billing-stripe \
+            BILLING_ALLOWED_ORIGINS="${STAGING_BILLING_ALLOWED_ORIGINS}" \
```
ZappoMan, line 113: Deploying `waitlist-capture`, `waitlist-ops`, and `kit-sync` together is fine, but note PR preview workflow (`.github/workflows/pr-preview-environment.yml`) still only deploys billing functions — preview branches won't get `kit-sync` or cron registration unless that's updated too.

supabase/functions/kit-sync/index.ts
```diff
@@ -0,0 +1,275 @@
...
+    case 'user.signed_up': {
+      if (formId) await kit.subscribeToForm(email, formId);
+      await kit.applyTag(email, signupTag(ns, opts));
+      const planId = payload.plan_id as string | undefined;
+      if (planId) {
+        await kit.applyTag(email, interestTag(ns, planId, opts));
+      }
+      break;
+    }
+    case 'waitlist.joined': {
```
rahul-artificerinnovations, line 202: **Bug: unknown `event_type` retries 5× instead of dead-lettering immediately.**

`processEvent` throws `new Error(...)` for unknown events (not a `KitClientError`), so `code` is `'unknown'`. `isPermanentKitError('unknown')` returns `false`, meaning the row burns through 5 attempts before it's dead-lettered.

Unknown events will never succeed — dead-letter immediately:
```typescript
default:
  throw new KitClientError(
    `Unknown event_type: ${event_type}`,
    'unknown_event_type'
  );
```
And add `'unknown_event_type'` (or `'unknown'`) to `PERMANENT_ERROR_CODES` in `kitClient.ts`.

supabase/functions/kit-sync/index.ts
```diff
@@ -0,0 +1,275 @@
...
+    let page = 1;
+    while (true) {
+      const { data: usersPage, error: uErr } =
+        await admin.auth.admin.listUsers({ page, perPage: 100 });
+      if (uErr) break;
+      const users = usersPage?.users ?? [];
+      if (users.length === 0) break;
+
+      for (const user of users) {
+        if (!user.email) continue;
```
rahul-artificerinnovations, line 287: **Unbounded fetch — add pagination.**

Unlike the `auth.users` loop above which pages 100 at a time, this fetches all `waitlist_entries` in one query. On a large table this will timeout or OOM the edge function.

```typescript
// Paginate like auth.users above
let ePage = 0;
while (true) {
  const { data: entries } = await admin
    .from('waitlist_entries')
    .select('id, email, metadata')
    .range(ePage * 100, (ePage + 1) * 100 - 1);
  if (!entries?.length) break;
  // ... process
  if (entries.length < 100) break;
  ePage++;
}
```

supabase/functions/kit-sync/index.ts
```diff
@@ -0,0 +1,275 @@
...
+        })
+        .eq('id', row.id);
+    }
+  }
+
+  return Response.json({ processed, skipped, failed });
+}
+
+async function processEvent(
+  kit: KitClient,
```
rahul-artificerinnovations, line 181: **Error update is fire-and-forget.** If this `.update()` call fails (network blip, timeout), the row stays as `'processing'` and is never retried. At minimum:
```typescript
const { error: updateErr } = await admin
  .from('marketing_email_sync_queue')
  .update({...})
  .eq('id', row.id);
if (updateErr) console.error('Failed to update error status for row', row.id, updateErr.message);
```

supabase/functions/kit-sync/index.ts
```diff
@@ -0,0 +1,275 @@
...
+      // Suppression check
+      const { data: suppressed } = await admin
+        .from('marketing_email_unsubscribes')
+        .select('id')
+        .eq('product_id', row.product_id)
+        .eq('email', row.email)
+        .maybeSingle();
+
+      if (suppressed) {
+        await markDone(admin, row.id);
```
rahul-artificerinnovations, line 129: **Nit: `markDone` should clear the `error` field** so a retried row that eventually succeeds doesn't retain the previous failure message:
```typescript
await admin
  .from('marketing_email_sync_queue')
  .update({ status: 'done', processed_at: new Date().toISOString(), error: null })
  .eq('id', id);
```

scripts/ensure-kit-sync-cron.mjs
```diff
@@ -0,0 +1,33 @@
...
+  console.error('Missing required env vars: SUPABASE_URL, SUPABASE_SERVICE_ROLE_KEY, KIT_CRON_SECRET');
+  process.exit(1);
+}
+
+const workerUrl = `${supabaseUrl}/functions/v1/kit-sync`;
+
+const res = await fetch(`${supabaseUrl}/rest/v1/rpc/kit_sync_setup_cron`, {
```
rahul-artificerinnovations, line 17: **No fetch timeout.** If `kit_sync_setup_cron` hangs, the deploy step stalls indefinitely. Add a timeout:
```js
const res = await fetch(`${supabaseUrl}/rest/v1/rpc/kit_sync_setup_cron`, {
  signal: AbortSignal.timeout(15_000),
  ...
});
```

tests/integration/waitlist-edge.test.ts
```diff
@@ -55,3 +55,140 @@ describeEdge('waitlist Edge integration', () => {
...
+    const { data: entry } = await sr
+      .from('waitlist_entries')
+      .insert({ email: email.toLowerCase(), metadata: {} })
+      .select('id')
+      .single();
+
+    expect(entry).not.toBeNull();
+
+    // Simulate what waitlist-ops approve does: enqueue waitlist.approved
+    const idempotencyKey = `waitlist.approved:${entry!.id}`;
```
rahul-artificerinnovations, line 171: **Nit: this test bypasses the actual HTTP path.** The comment acknowledges it, but worth noting: the new `entryPlanId` extraction in `waitlist-ops/index.ts` (the `approve` action) is not exercised here — you're inserting directly via service role. The `plan_id`-in-payload path only gets covered if/when a full admin-auth integration test is added.

rahul-artificerinnovations (changes requested):
## Review

Solid Phase 3 shipment overall — the queue/worker/cron/client architecture is clean, the permission model (SECURITY DEFINER for cron registration, SECURITY INVOKER for dequeue) is correct, and the idempotency handling looks right. A few things need attention before merge, including one that will cause silent wasted retries in production.

### Must-fix

1. **Unknown `event_type` is retried 5× instead of dead-lettered immediately** (see inline comment in `kit-sync/index.ts`). `new Error(...)` has `code === 'unknown'`, which `isPermanentKitError` returns false for, so the row burns through 5 attempts before being dead-lettered. An unrecognised event type will never succeed — it should be dead-lettered on the first attempt.

### Should-fix

2. **`handleBulkSync` fetches all `waitlist_entries` without pagination** — unlike the `auth.users` loop which paginates 100 at a time, `waitlist_entries` is a single unbounded fetch. On a table with tens of thousands of rows this will timeout or OOM the edge function.
3. **Error update in the catch block is fire-and-forget** — if the `.update()` call fails, the row stays as `'processing'` indefinitely. At minimum, log the error.
4. **Stuck `'processing'` rows** — if the edge function crashes mid-batch (OOM, timeout), all rows in that batch stay `'processing'` forever. A requeue sweep (`UPDATE ... SET status='pending' WHERE status='processing' AND last_attempted_at < now() - interval '10 minutes'`) in either the cron SQL or the worker would close this gap.

### Nits

5. `ensure-kit-sync-cron.mjs` has no fetch timeout — a hanging RPC call will stall the deploy step indefinitely. `AbortSignal.timeout(15000)` or similar.
6. `markDone` doesn't clear the `error` field — on retry success, the prior error text stays on the row, which is confusing during debugging.
7. The `waitlist-ops approve` integration test doesn't call the actual HTTP endpoint — it simulates the queue insertion directly. The new `entryPlanId` path in `waitlist-ops/index.ts` remains untested.
8. Double blank line between deploy steps in both workflow files (cosmetic).

The `PERFORM cron.unschedule(...) WHERE EXISTS (...)` pattern is fine — pg_cron's `cron.unschedule(text)` returns boolean, so this is valid plpgsql.

supabase/functions/kit-sync/index.ts
```diff
@@ -0,0 +1,275 @@
+// Worker: dequeues marketing_email_sync_queue rows and calls Kit Creator API v4.
+// Scheduled every 5 min via pg_cron + pg_net (see kit_sync_cron migration).
+// Auth: Bearer token matching KIT_CRON_SECRET env var.
+
+import { createClient } from 'https://esm.sh/@supabase/supabase-js@2';
```
sarah-artificerinnovations: **Significant — import inconsistency:** Every other Edge Function in this repo uses `npm:@supabase/supabase-js@2.45.0`. This one uses `esm.sh`, which resolves at runtime from a CDN and can return a different minor/patch than what's pinned everywhere else. Please switch to:

```ts
import { createClient } from 'npm:@supabase/supabase-js@2.45.0';
```

supabase/functions/kit-sync/index.ts
```diff
@@ -0,0 +1,275 @@
...
+          .from('marketing_email_settings')
+          .select('config')
+          .eq('product_id', row.product_id)
+          .eq('enabled', true)
+          .maybeSingle();
+        settingsCache.set(row.product_id, (data?.config as SettingsConfig) ?? null);
+      }
+      const config = settingsCache.get(row.product_id) ?? null;
+
+      if (!config) {
```
sarah-artificerinnovations: **Significant — silent data loss when settings are disabled:** If `marketing_email_settings` is disabled or missing (temporarily, mid-migration, or during an outage), every queued event gets permanently marked `done` and dropped.

These events should stay `pending` so they're retried once settings are re-enabled. Consider `continue` (skip, leave as pending) rather than `markDone`.

supabase/functions/kit-sync/index.ts
sarah-artificerinnovations: **Blocker — field name mismatch:** This reads `payload.new_plan_id`, but the Phase 2 stripe-webhook writer stores the field as `plan_id` (see `stripe-webhook/index.ts` line ~385, merged in PR #308):

```ts
{ user_id: row.user_id, plan_id: finalPlanId, status: finalStatus }
```

Because `new_plan_id` is always `undefined`, the guard on the next line throws `'user.tier_changed payload missing new_plan_id'` for every event, which hits the permanent-error dead-letter path. Every tier-change sync will silently fail.

Fix: read `payload.plan_id` to match what's written.

supabase/functions/kit-sync/index.ts
```diff
@@ -0,0 +1,275 @@
...
+    default:
+      throw new Error(`Unknown event_type: ${event_type}`);
+  }
+}
+
+async function markDone(
+  admin: ReturnType<typeof createClient>,
+  id: number
+): Promise<void> {
+  await admin
```
sarah-artificerinnovations: **Nit — `markDone` errors silently swallowed:** If the Supabase update fails (network blip, RLS change), the row stays as `processing` indefinitely and will never be retried. At minimum log the error so it shows up in Edge Function logs:

```ts
const { error } = await admin.from('marketing_email_sync_queue')
  .update({ status: 'done', processed_at: new Date().toISOString() })
  .eq('id', id);
if (error) console.error(`markDone failed for id ${id}:`, error);
```

supabase/functions/kit-sync/index.ts
```diff
@@ -0,0 +1,275 @@
...
+          .ignore();
+        if (!insErr) enqueued++;
+      }
+
+      if (users.length < 100) break;
+      page++;
+    }
+
+    // Backfill waitlist_entries as waitlist.joined
+    const { data: entries } = await admin
```
sarah-artificerinnovations: **Significant — bulk sync not paginated:** This fetches all `waitlist_entries` rows in a single query with no `limit`/`range`. On a large waitlist this will OOM the Edge Function (512 MB default). The `auth.users` sync path above is paginated — this one should be too. Consider a `range(offset, offset + BATCH_SIZE - 1)` loop matching the pattern used for users.

supabase/functions/_shared/kitClient.ts
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  // Hard-deletes the subscriber from Kit for GDPR erasure propagation.
+  async deleteUser(email: string): Promise<void> {
+    const subId = await this.findSubscriberId(email);
+    if (!subId) return; // already deleted — idempotent
+    await this.delete(`/subscribers/${subId}`);
+  }
+
+  private async findTag(name: string): Promise<string | null> {
```
sarah-artificerinnovations, line 123: **Nit — `findTag` doesn't paginate:** The Kit v4 `GET /tags` endpoint is paginated (default page size 500). If the account has more tags than one page, this will miss tags on subsequent pages and cause `applyTag`/`removeTag` to spuriously return `kit_tag_not_found`, dead-lettering events that should succeed. Worth adding a simple `?page=N` loop here.

scripts/ensure-kit-sync-cron.mjs
sarah-artificerinnovations, line 17: **Nit — no fetch timeout:** If Supabase is slow to respond this script will hang indefinitely in CI, blocking the deploy step. Consider wrapping with `AbortSignal.timeout(10_000)`:

```js
const res = await fetch(url, {
  signal: AbortSignal.timeout(10_000),
  ...
});
```

supabase/migrations/20260521000000_kit_sync_cron.sql
```diff
@@ -0,0 +1,82 @@
...
+  PERFORM cron.schedule(
+    'kit-sync-worker',
+    '*/5 * * * *',
+    format(
+      $sql$
+        SELECT extensions.http_post(
+          url     := %L,
+          headers := jsonb_build_object(
+            'Content-Type',  'application/json',
+            'Authorization', 'Bearer ' || %L
```
sarah-artificerinnovations: **Nit — `KIT_CRON_SECRET` stored in plain text in `cron.job`:** The `p_secret` value ends up in the `command` column of `cron.job`, which is readable by any role with access to that table. This is a known pg_cron limitation — worth a comment here documenting the trade-off and noting that access to `cron.job` should be restricted to `service_role`.

sarah-artificerinnovations (changes requested):
## PR #311 Review — kit-sync worker + cron scheduling

There's one blocker that will cause every `user.tier_changed` event to dead-letter immediately. A few significant data-integrity issues below that, plus some nits.

**Blocker:** `kit-sync/index.ts` reads `payload.new_plan_id` but the Phase 2 queue writer stores `plan_id`. These need to match before this ships.

Inline comments on everything else below.

supabase/functions/_shared/kitClient.ts
```diff
@@ -0,0 +1,160 @@
...
+        'kit_tag_not_found'
+      );
+    }
+    await this.post(`/tags/${tagId}/subscribers`, { email_address: email });
+  }
+
+  async removeTag(email: string, tagName: string): Promise<void> {
+    const tagId = await this.findTag(tagName);
+    if (!tagId) return; // tag doesn't exist — idempotent
+    await this.delete(`/subscribers/${encodeURIComponent(email)}/tags/${tagId}`);
```
diego-artificerinnovations: **🚨 Blocking — Wrong URL format: Kit v4 `DELETE /subscribers` requires subscriber ID, not email**

Kit Creator API v4's endpoint is `DELETE /subscribers/{id}/tags/{tag_id}` where `{id}` is a **numeric subscriber ID**, not an email address. Using `encodeURIComponent(email)` here will produce a 404 response from Kit for every `removeTag` call, which is in `PERMANENT_ERROR_CODES` → immediate dead-letter for every `waitlist.converted`, `user.tier_changed`, and `user.churned` event.

Fix: resolve the subscriber ID first (same pattern as `deleteUser`/`unsubscribeUser`), then use it in the path:

```typescript
async removeTag(email: string, tagName: string): Promise<void> {
  const tagId = await this.findTag(tagName);
  if (!tagId) return;
  const subId = await this.findSubscriberId(email);
  if (!subId) return; // subscriber not in Kit — idempotent
  await this.delete(`/subscribers/${subId}/tags/${tagId}`);
}
```

supabase/functions/kit-sync/index.ts
diego-artificerinnovations: **🚨 Blocking — Wrong payload key: Phase 2 sends `plan_id`, not `new_plan_id`**

`stripe-webhook/index.ts` (Phase 2) enqueues `user.tier_changed` with payload `{ user_id, plan_id, status }`. This line reads `payload.new_plan_id` — which will always be `undefined` — and then throws. Every real `user.tier_changed` event will retry 5× and dead-letter as `'failed'`.

Fix:
```typescript
const newPlanId = payload.plan_id as string | undefined;
```

supabase/functions/kit-sync/index.ts
```diff
@@ -0,0 +1,275 @@
...
+            event_type: 'user.signed_up',
+            email: user.email,
+            payload: { user_id: user.id, created_at: user.created_at },
+            idempotency_key: `bulk:user.signed_up:${product_id}:${user.id}`,
+            status: 'pending',
+          })
+          .onConflict('idempotency_key')
+          .ignore();
+        if (!insErr) enqueued++;
+      }
```
diego-artificerinnovations, line 302: **🚨 Blocking — `.onConflict().ignore()` is not a method on the Supabase JS v2 insert builder**

`insert()` returns a `PostgrestFilterBuilder` which has no `.onConflict()` method — that doesn't exist in Supabase JS v2. Calling this will throw a TypeError at runtime, aborting the entire bulk-sync action.

Fix: use `upsert` with `ignoreDuplicates: true` for DO NOTHING semantics:
```typescript
const { error: insErr } = await admin
  .from('marketing_email_sync_queue')
  .upsert(
    {
      product_id,
      event_type: 'user.signed_up',
      email: user.email,
      payload: { user_id: user.id, created_at: user.created_at },
      idempotency_key: `bulk:user.signed_up:${product_id}:${user.id}`,
      status: 'pending',
    },
    { onConflict: 'idempotency_key', ignoreDuplicates: true }
  );
```

Same fix applies to the `waitlist_entries` backfill loop below (~line 274).

supabase/functions/kit-sync/index.ts
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    // Backfill waitlist_entries as waitlist.joined
+    const { data: entries } = await admin
+      .from('waitlist_entries')
+      .select('id, email, metadata');
+    for (const entry of entries ?? []) {
+      if (!entry.email) continue;
+      const { error: insErr } = await admin
+        .from('marketing_email_sync_queue')
```
diego-artificerinnovations: **Medium — `waitlist_entries` backfill isn't paginated**

The `auth.users` backfill above uses `page`/`perPage: 100` pagination. This `waitlist_entries` fetch has no limit — for large tables it could OOM the Edge Function or hit the PostgREST row limit (default 1000).

Suggestion: use a range-based loop mirroring the users pagination, e.g. `from(0)` / `range(offset, offset + 99)`.

tests/integration/waitlist-edge.test.ts
```diff
@@ -55,3 +55,140 @@ describeEdge('waitlist Edge integration', () => {
...
+    const { data: row, error } = await sr
+      .from('marketing_email_sync_queue')
+      .select('event_type, status, email, product_id')
+      .eq('email', email.toLowerCase())
+      .eq('product_id', TEST_MKTG_PRODUCT_ID)
+      .eq('event_type', 'waitlist.joined')
+      .maybeSingle();
+
+    expect(error).toBeNull();
+    expect(row).not.toBeNull();
```
diego-artificerinnovations, line 116: **Medium — this test doesn't exercise the Edge function for `waitlist.approved`**

The comment acknowledges this, but it's worth naming the gap clearly for tracking: the test inserts the queue row directly via service role rather than calling `waitlist-ops approve` over HTTP. That means a regression in the `approve` action's `enqueueMarketingEmail` call wouldn't be caught here.

Suggest adding a `// TODO: Phase 3 follow-up — add a full-stack approve test once admin JWT helper is available in CI` comment, or opening a tracking issue, so it doesn't get forgotten.

supabase/migrations/20260521000000_kit_sync_cron.sql
```diff
@@ -0,0 +1,82 @@
...
+  UPDATE marketing_email_sync_queue
+  SET
+    status            = 'processing',
+    last_attempted_at = now()
+  WHERE id IN (
+    SELECT id
+    FROM   marketing_email_sync_queue
+    WHERE  status = 'pending'
+    ORDER  BY created_at
+    LIMIT  batch_size
```
diego-artificerinnovations, line 34: **Nit — ORDER BY doesn't match the partial index key**

The Phase 2 partial index is `(product_id, created_at) WHERE status IN ('pending', 'failed')`. The dequeue subquery uses `ORDER BY created_at` without `product_id` — Postgres can use the partial index but has to do a full scan + sort rather than a leading-column range scan.

Not blocking for batch-10 workloads, but `ORDER BY product_id, created_at` would get a cleaner index scan and is consistent with the index design intent.

diego-artificerinnovations (changes requested):
Good work overall — the architecture is clean (SKIP LOCKED dequeue, settings cache, suppression check, dead-letter on permanent errors) and the cron setup is well thought out. Found three blocking issues that need fixes before merge, plus two medium items.

**Blocking (must fix):**
1. `removeTag` in `kitClient.ts` uses email in the subscriber URL — Kit v4 requires a numeric subscriber ID. Every `waitlist.converted`, `user.tier_changed`, and `user.churned` event will 404 → immediate dead-letter.
2. `user.tier_changed` reads `payload.new_plan_id` but Phase 2 (`stripe-webhook`) sends `payload.plan_id`. Every real event will throw and retry 5× before being dead-lettered.
3. `.insert({}).onConflict('col').ignore()` doesn't exist in Supabase JS v2 — bulk-sync will throw TypeError at runtime.

**Medium:**
- `waitlist_entries` backfill isn't paginated (unlike the `auth.users` loop); will OOM or hit row limits at scale.
- `waitlist.approved` integration test inserts the queue row directly rather than going through the Edge function — the approval path isn't really tested.

**Nit:** dequeue `ORDER BY` doesn't match the partial index leading column (`product_id`).

Inline comments on all six items below.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — bulk-sync inserts now use `{ onConflict: 'idempotency_key', ignoreDuplicates: true }` as the second argument to `.insert()`, which is the correct Supabase JS v2 API.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — `processEvent` now reads `payload.plan_id` for both `user.signed_up` and `user.tier_changed`. If `plan_id` is missing on `user.tier_changed`, it throws a `KitClientError` with code `kit_api_400` to dead-letter immediately rather than retry.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — added `planIdToSlug(planId, tierTagNames)` helper that checks exact match first, then tries the last `_`-separated suffix (so `beakerstack_pro → pro`). `user.tier_changed` removes all tier tags then applies the resolved slug. `user.signed_up` applies an `interest:<slug>` tag when `plan_id` is present.

supabase/tests/kit_sync_cron.test.sql
mei-artificerinnovations: Fixed in `20260521000000_kit_sync_cron.sql` — test 7 now uses `SELECT p.proacl IS NOT NULL FROM pg_proc p JOIN pg_namespace n ON n.oid = p.pronamespace WHERE n.nspname = 'public' AND p.proname = 'kit_sync_setup_cron'`. `PUBLIC` is a pseudo-role and can't be passed to `has_function_privilege`; a non-null `proacl` is the correct indicator that `REVOKE FROM PUBLIC` was applied.

supabase/migrations/20260521000000_kit_sync_cron.sql
mei-artificerinnovations: Fixed — the cron job now uses `net.http_post` with `SET search_path = extensions, net, public`. I also took this further: the secret is no longer inlined in the job body at all (see Sarah's nit below — it's stored as a database GUC via `ALTER DATABASE ... SET app.kit_cron_secret = ...` and read at runtime via `current_setting('app.kit_cron_secret')`).

supabase/migrations/20260521000000_kit_sync_cron.sql
mei-artificerinnovations: `failed` rows are intentionally excluded from dequeue — they are dead-lettered (permanent errors or attempts ≥ 5) and require manual intervention. The comment in the migration now spells this out explicitly: `'failed' rows are intentionally excluded: they are dead-lettered (permanent errors or attempts >= 5) and require manual intervention.`

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — `kit_sync_dequeue` now also claims rows where `status = 'processing' AND last_attempted_at < now() - interval '15 minutes'`. These are rows where the Edge Function died after dequeue but before updating status.

tests/integration/waitlist-edge.test.ts
mei-artificerinnovations: Fixed — test renamed to `'waitlist.approved queue row has correct shape (DB-level; Edge auth complexity deferred)'` with an updated comment explaining why we're seeding directly instead of hitting the HTTP endpoint.

scripts/ensure-kit-sync-cron.mjs
mei-artificerinnovations: Fixed — the step now gates on a bash check: `[[ -z "${KIT_CRON_SECRET:-}" ]] && { echo "..."; exit 0; }`. This matches the pattern used elsewhere in the workflow for optional secrets (e.g. the CloudFront signing key step). `secrets.*` can't be used in step `if:` directly.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Acknowledged — the code comment now explicitly notes this: `-- Note: does not backfill user.tier_changed from billing_subscriptions — that is deferred to Phase 5 admin backfill UI.` Tier change history backfill needs the subscription history which is a larger operation, scoped to Phase 5.

.github/workflows/deploy-staging.yml
mei-artificerinnovations: Acknowledged — the `supabase secrets set` step already guards this: if `STAGING_KIT_API_KEY` or `STAGING_KIT_CRON_SECRET` are empty strings, Supabase CLI will set them to empty (harmless). The `ensure-kit-sync-cron.mjs` step now has a bash guard that skips if `KIT_CRON_SECRET` is unset, so repos without Kit configured won't fail.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — the error-path `.update()` is now awaited: `const { error: updateErr } = await admin.from(...).update(...)` with `if (updateErr) console.error(...)`. Same for the rate-limit reset path.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — `default` case now throws `new KitClientError(`Unknown event_type: ${event_type}`, 'unknown_event_type')`, and `'unknown_event_type'` is added to `PERMANENT_ERROR_CODES`. The worker dead-letters these immediately instead of retrying 5×.

supabase/functions/_shared/kitClient.ts
mei-artificerinnovations: Fixed — `isRateLimitError(code)` now returns true for `kit_api_429`. In the catch block, 429s reset the row to `pending` and update `last_attempted_at` without incrementing `attempts`, so the row is retried on the next cron tick without burning a retry slot.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — the `waitlist_entries` backfill is now paginated with `.range(ePage * 100, (ePage + 1) * 100 - 1)` in a `while (true)` loop that breaks when a page returns fewer than 100 rows, matching the pattern used for the `auth.users` loop above.

tests/integration/waitlist-edge.test.ts
mei-artificerinnovations: Fixed — test renamed to `'waitlist.approved queue row has correct shape (DB-level; Edge auth complexity deferred)'` with a comment explaining the constraint. The `waitlist-ops` approve path requires an admin JWT which isn't available in CI without significant scaffolding — calling it a DB-level shape test is accurate.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — the worker now distinguishes: if no row exists in `marketing_email_settings` → `markDone` (drop the event, no config for this product). If a row exists but `enabled = false` → reset to `pending` so the event is processed if the product is later re-enabled.

scripts/ensure-kit-sync-cron.mjs
mei-artificerinnovations: Fixed — `fetch(..., { signal: AbortSignal.timeout(15_000) })` added to the RPC call in `ensure-kit-sync-cron.mjs`.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — import is now `import { createClient } from 'npm:@supabase/supabase-js@2.45.0'` matching every other Edge Function in this repo.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — `markDone` now sets `error: null` in the update payload to clear any previous failure message. Also added error logging if the update itself fails.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — paginated with `.range(ePage * 100, (ePage + 1) * 100 - 1)` in a while loop, breaking when entries.length < 100.

supabase/functions/_shared/kitClient.ts
mei-artificerinnovations: Fixed — `findTag` now calls `/tags?name=...&per_page=1`. Tag names are unique in Kit, so a single result is sufficient and avoids pagination entirely.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — reads `payload.plan_id` (not `payload.new_plan_id`). For `user.tier_changed`, throws `KitClientError(..., 'kit_api_400')` if missing to dead-letter immediately.

scripts/ensure-kit-sync-cron.mjs
mei-artificerinnovations: Fixed — `AbortSignal.timeout(15_000)` added to the fetch call in `ensure-kit-sync-cron.mjs`.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — `markDone` now awaits and logs errors: `const { error } = await admin.from(...).update(...)` + `if (error) console.error(...)`. Also clears `error: null` so a successfully retried row doesn't keep the old failure message.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — `payload.plan_id` throughout. For `user.tier_changed`, throws `KitClientError(..., 'kit_api_400')` if the field is missing so it dead-letters immediately.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — bulk-sync inserts now use `.insert({ ... }, { onConflict: 'idempotency_key', ignoreDuplicates: true })`. This is the correct Supabase JS v2 signature.

supabase/migrations/20260521000000_kit_sync_cron.sql
mei-artificerinnovations: Fixed — `kit_sync_setup_cron` now stores the secret as a database GUC: `EXECUTE format('ALTER DATABASE %I SET app.kit_cron_secret = %L', current_database(), p_secret)`. The cron job body reads it at runtime via `current_setting('app.kit_cron_secret')`, so the literal secret never appears in `cron.job.command`.

supabase/functions/_shared/kitClient.ts
mei-artificerinnovations: Fixed — `removeTag` now calls `findSubscriberId(email)` first and constructs the path as `/subscribers/${subId}/tags/${tagId}`. If the subscriber isn't in Kit it returns early (idempotent). This was the critical bug causing 404 on every tag-removal event.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — paginated with `.range(ePage * 100, (ePage + 1) * 100 - 1)` in a while loop matching the auth.users pattern above.

tests/integration/waitlist-edge.test.ts
mei-artificerinnovations: Fixed — test renamed to make the coverage limitation explicit. Full Edge HTTP approval testing requires an admin JWT which is impractical in CI.

supabase/migrations/20260521000000_kit_sync_cron.sql
mei-artificerinnovations: Acknowledged — `ORDER BY created_at` is correct for ordering purposes (process oldest-first). The `FOR UPDATE SKIP LOCKED` in the subquery handles the concurrency safety. The partial index on `status = 'pending'` from Phase 2 still benefits the `WHERE status = 'pending'` half of the OR clause; the `status = 'processing'` recovery path is infrequent enough that a seq scan on that arm is fine.

mei-artificerinnovations:
Pushed fix commit `20ef18ab` addressing all feedback from Rahul, Sarah, Diego, and ZappoMan.

**Blockers fixed:**
- `removeTag` now looks up subscriber ID before DELETE (was using email in URL → 404 on every tag-removal event)
- `payload.plan_id` (not `new_plan_id`) for `user.tier_changed` and `user.signed_up`
- `.insert()` now uses `{ ignoreDuplicates: true }` (valid Supabase JS v2 API)

**Must-fix / significant:**
- Unknown `event_type` dead-letters immediately via `KitClientError(..., unknown_event_type)` + `PERMANENT_ERROR_CODES`
- Disabled settings: reset to `pending` (not dropped) so events survive a re-enable
- Stuck `processing` rows: `kit_sync_dequeue` now reclaims rows >15 min old
- Waitlist entries backfill: paginated with `.range()`
- Error updates: awaited + logged

**Nits + additional fixes:**
- `markDone`: logs errors, clears `error` field on success
- `findTag`: `&per_page=1` (tag names unique in Kit)
- `ensure-kit-sync-cron.mjs`: `AbortSignal.timeout(15_000)`
- Migration: `net.http_post` + correct `search_path`; cron secret stored as database GUC (`app.kit_cron_secret`) not inlined in `cron.job.command`
- pgTAP test 7: `proacl IS NOT NULL` check (PUBLIC is a pseudo-role)
- Integration test: renamed approve test to accurately describe what it covers
- CI: bash gate before `ensure-kit-sync-cron.mjs` when `KIT_CRON_SECRET` unset
- Import: `npm:@supabase/supabase-js@2.45.0`
- 429 rate limit: reset to `pending` without incrementing attempts

Replied inline to every comment. Ready for re-review.

sarah-artificerinnovations (approved):
Every item from my prior review is addressed — and then some.

**Blockers / significant:**
- `plan_id` / `new_plan_id` mismatch → fixed; worker now reads `payload.plan_id` with a comment tying it back to the stripe-webhook writer.
- `esm.sh` import → replaced with `npm:@supabase/supabase-js@2.45.0` to match the rest of the codebase.
- Disabled-settings data loss → the disabled branch now resets to `pending` and skips rather than marking done, so events are retried when settings come back on.
- `waitlist_entries` bulk sync pagination → now uses `range(ePage * 100, (ePage + 1) * 100 - 1)` matching the paginated `auth.users` path above it.

**Nits:**
- `markDone` now logs the Supabase error on failure instead of swallowing it.
- `ensure-kit-sync-cron.mjs` has `AbortSignal.timeout(15_000)` — CI won't hang.
- `findTag` uses `per_page=1` and a comment explaining that Kit tag names are unique, so a single result is sufficient. Clean solution.
- The `KIT_CRON_SECRET` is now stored as a database GUC (`ALTER DATABASE ... SET app.kit_cron_secret`) and read at runtime via `current_setting()`, rather than inlined into `cron.job.command`. The comment at the top of `kit_sync_setup_cron` explains the trade-off well.

A few things I noticed that are worth calling out as genuine improvements beyond my review:
- `kit_sync_dequeue` now reclaims rows stuck in `processing` for >15 min — handles the case where an Edge Function dies mid-batch.
- `planIdToSlug` handles Stripe plan IDs with product prefixes (e.g. `bhg_pro` → `pro`) cleanly.
- `unknown_event_type` is now a permanent error code so unknown events dead-letter immediately rather than burning all 5 retry attempts.

Ship it.

rahul-artificerinnovations (approved):
All previously raised issues are addressed in `20ef18ab`. Remaining sections (migration SQL, pgTAP tests, integration tests) look clean — approving.

**`20260521010000_marketing_email_plan_interest.sql`** — trigger rewrite correctly guards on `NOT FOUND` and empty email before inserting, conditionally includes `plan_id` only when present, idempotency key unchanged, `REVOKE FROM PUBLIC` applied.

**`kit_sync_cron.test.sql`** — 8 pgTAP assertions cover extension presence, SECURITY INVOKER / DEFINER properties, REVOKE-from-PUBLIC enforcement, and a functional dequeue test all inside `BEGIN`/`ROLLBACK`. The `proacl IS NOT NULL` check for test 7 is the standard way to confirm an explicit ACL override was applied.

**`waitlist-edge.test.ts`** — three new integration tests: happy-path enqueue, idempotency (double-submit → one row), and schema validation for `waitlist.approved`. The direct-seed approach for the approved test is honest about the CI limitation and still validates the queue schema. Cleanup in `afterAll` is complete.

Ship it. 🚢

diego-artificerinnovations (approved):
All three of my blocking issues are resolved correctly, and Rahul's/Sarah's items are in too. Full verification:

**Blocking fixes (all ✅)**
- `removeTag` now resolves subscriber ID via `findSubscriberId()` before the DELETE URL — correct Kit v4 path
- `user.tier_changed` reads `payload.plan_id` (matches what `stripe-webhook` enqueues); missing value now throws a `KitClientError` with code `kit_api_400` → immediate dead-letter via `PERMANENT_ERROR_CODES`
- Bulk-sync uses `insert({ ... }, { onConflict: 'idempotency_key', ignoreDuplicates: true })` — no more invalid `.onConflict().ignore()` chain; `waitlist_entries` backfill is now paginated with `.range()`

**Other improvements worth calling out**
- Settings cache now fetches `enabled` + `config` without pre-filtering on `enabled: true` — disabled-product rows are reset to `'pending'` instead of silently dropped. Much better failure mode.
- Rate limit handling: 429 resets to `'pending'` without burning an attempt — prevents needless dead-lettering during Kit API rate limit windows.
- Stuck `processing` row recovery (>15 min) in `kit_sync_dequeue` handles crashed worker scenarios cleanly.
- GUC-based secret storage (`ALTER DATABASE SET app.kit_cron_secret`) keeps the plaintext out of `cron.job.command`, which is readable by any DB user with access to that table.
- `markDone` now clears `error: null` on success and logs update failures.
- `findTag` adds `&per_page=1` — avoids fetching a full page of tags for a unique name lookup.
- CI cron step gated on secret presence; `AbortSignal.timeout(15_000)` in the ensure script.
- `planIdToSlug` normalizes Stripe plan IDs like `bhg_pro` → `pro` when `tierTagNames` is configured — good defensive DX.

**One nit to watch for Phase 4**
The `insert({ ... }, { onConflict, ignoreDuplicates })` idiom works in practice (23505 on duplicate is swallowed by the `if (!insErr)` check, so no double-inserts), but `upsert({ ... }, { onConflict, ignoreDuplicates })` is the documented Supabase JS v2 API for DO NOTHING semantics. Worth cleaning up when you're in that file again, but not a blocker.

Ready to merge.

rahul-artificerinnovations (approved):
All previously raised issues are addressed in `20ef18ab`. Remaining sections (migration SQL, pgTAP tests, integration tests) look clean — approving.

**`20260521010000_marketing_email_plan_interest.sql`** — trigger rewrite correctly guards on `NOT FOUND` and empty email before inserting, conditionally includes `plan_id` only when present, idempotency key unchanged, `REVOKE FROM PUBLIC` applied.

**`kit_sync_cron.test.sql`** — 8 pgTAP assertions cover extension presence, SECURITY INVOKER / DEFINER properties, REVOKE-from-PUBLIC enforcement, and a functional dequeue test all inside `BEGIN`/`ROLLBACK`. The `proacl IS NOT NULL` check for test 7 is the standard way to confirm an explicit ACL override was applied.

**`waitlist-edge.test.ts`** — three new integration tests: happy-path enqueue, idempotency (double-submit → one row), and schema validation for `waitlist.approved`. The direct-seed approach for the approved test is honest about the CI limitation and still validates the queue schema. Cleanup in `afterAll` is complete.

Ship it.

Issue #315, opened 2026-05-21
Kit Phase 3 follow-up: deploy secret handling, plan_id pipeline, operator setup (#311)


ZappoMan:
## Context

Follow-up from post-merge review of **PR #311** (`376dd91` on `develop`) — Kit Phase 3 worker, cron scheduling, and `plan_interest` propagation (#286).

PR #311 landed the core worker and addressed several pre-merge review items (pgTAP fix, `plan_id` in tier worker, `net.http_post`, stale `processing` reclaim, 429 handling, cron secret via DB GUC, deploy cron gating). The items below are **still blocking or high-risk** before treating Kit sync as production-ready.

---

### 1. Deploy overwrites Kit secrets with empty strings

`deploy-staging.yml` / `deploy-production.yml` always run:

```yaml
supabase secrets set ... KIT_API_KEY="${STAGING_KIT_API_KEY}" KIT_CRON_SECRET="${STAGING_KIT_CRON_SECRET}"
```

When GitHub secrets are unset, this passes **empty strings** on every deploy — potentially **clobbering** manually configured Supabase secrets and leaving `kit-sync` with a blank `KIT_API_KEY` (all queue rows dead-letter after retries).

The cron ensure step is already gated (`[[ -z "${KIT_CRON_SECRET}" ]] && exit 0`), but **`supabase secrets set` is not**.

**Fix:** Only include `KIT_API_KEY` / `KIT_CRON_SECRET` in `supabase secrets set` when the corresponding GitHub secret is non-empty (same pattern as Resend auth sync opt-in).

---

### 2. `plan_id` / interest-tag pipeline still incomplete

Phase 3 partially shipped `plan_id` propagation, but end-to-end intent tagging still does not work:

| Path | Gap |
|------|-----|
| Open signup (`SignupPage`) | Does not pass `{ plan_id }` in `auth.signUp` metadata when `?plan=` is present |
| Waitlist capture | Stores `plan_interest` (display name), not `plan_id`; `waitlist-ops approve` reads `metadata.plan_id` which is never set |
| `waitlist-capture` queue payload | Wraps metadata but worker `waitlist.joined` handler ignores `plan_id` / interest |
| OAuth signup | Intent metadata still not preserved through OAuth callback |

Auth trigger + worker interest tag on `user.signed_up` **will work only after** open-signup metadata is wired.

Refs: #286 product-review comment (intent vs tier tags), Phase 3 checklist item B.

---

[...]

PR #316, opened 2026-05-21
fix(kit-sync): address #315 blockers — secret gating, plan_id pipeline, config guard

mei-artificerinnovations:
## Summary

Addresses the three blockers in #315 raised after PR #311 (Kit Phase 3) merged.

### Fix 1 — Deploy does not clobber Kit secrets when GitHub secrets are unset

`supabase secrets set` previously always passed `KIT_API_KEY` and `KIT_CRON_SECRET`, even when the corresponding GitHub secrets were absent (empty string). On every deploy this would silently overwrite any manually-configured Supabase project secrets.

**Fix:** split the `secrets set` call — non-Kit secrets always set; Kit secrets only set when the GitHub secret is non-empty (same guard pattern as the cron ensure step).

### Fix 2 — plan_id flows end-to-end from signup/waitlist → queue → Kit interest tag

Four gaps closed:

| Layer | Before | After |
|---|---|---|
| `useAuth.signUp` | no metadata support | optional `{ data }` param forwarded to Supabase `auth.signUp` |
| `SignupPage` email signup | `auth.signUp(email, pw)` | passes `{ data: { plan_id } }` when `?plan=` is present |
| `SignupPage` waitlist metadata | `{ plan_interest: displayName }` only | adds `plan_id` (stable ID) so `waitlist-ops approve` can read `metadata.plan_id` |
| `waitlist-capture` queue payload | `{ metadata: {...} }` | hoists `plan_id` to top-level so worker doesn't need to dig into metadata |
| kit-sync `waitlist.joined` | form subscribe + waitlist tag only | adds interest tag when `payload.plan_id` present |
| kit-sync `waitlist.approved` | form subscribe + approved tag only | adds interest tag when `payload.plan_id` present |

> **OAuth gap:** `signInWithGoogle` cannot pass user metadata at OAuth initiation time (Supabase OAuth limitation). The postAuthPath already carries `?plan=...` so the billing redirect still works; Kit interest tagging via OAuth is deferred to a follow-up (noted in #315).

### Fix 3 — Config guard: missing namespace dead-letters with a clear error

When `marketing_email_settings.config.namespace` is unset, the worker previously reached the Kit API and got a confusing `kit_tag_not_found` error (permanent → dead-lettered). Now the worker checks namespace first and dead-letters with:
`marketing_email_settings.config missing required field namespace for product X`

### Known limitations (not in scope for this PR, tracked in #315)

- Bulk-sync does not backfill `user.tier_changed` from `billing_subscriptions` — deferred to Phase 5 admin UI
- OAuth signup does not propagate `plan_id` to user metadata — deferred

## Test plan

- [ ] Deploy staging: verify `supabase secrets set` run with no Kit GitHub secrets set does not overwrite existing Supabase secrets (check Supabase dashboard before/after)
- [ ] Sign up with `?plan=beakerstack_pro` via email — confirm `auth.users.raw_user_meta_data` contains `plan_id`
- [ ] Join waitlist with `?plan=beakerstack_pro` — confirm `waitlist_entries.metadata` has `plan_id` and `marketing_email_sync_queue` row has `payload.plan_id`
- [ ] Approve a waitlist entry with `plan_id` in metadata — confirm kit-sync applies `{ns}:interest:{slug}` tag in Kit dashboard
- [ ] Seed a `marketing_email_settings` row with `config = {}` (missing namespace) and trigger a queue row — confirm it dead-letters with the descriptive error

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -137,9 +137,11 @@ jobs:
             BILLING_SUPABASE_URL="${STAGING_SUPABASE_URL}" \
             BILLING_SUPABASE_ANON_KEY="${STAGING_SUPABASE_ANON_KEY}" \
             BILLING_SUPABASE_SERVICE_ROLE_KEY="${STAGING_SUPABASE_SERVICE_ROLE_KEY}" \
-            BILLING_ALLOWED_ORIGINS="${STAGING_BILLING_ALLOWED_ORIGINS}" \
-            KIT_API_KEY="${STAGING_KIT_API_KEY}" \
-            KIT_CRON_SECRET="${STAGING_KIT_CRON_SECRET}"
+            BILLING_ALLOWED_ORIGINS="${STAGING_BILLING_ALLOWED_ORIGINS}"
+          # Only set Kit secrets when the GitHub secrets are configured; passing empty
+          # strings would clobber any manually-configured Supabase project secrets.
+          [[ -n "${STAGING_KIT_API_KEY:-}" ]] && supabase secrets set KIT_API_KEY="${STAGING_KIT_API_KEY}"
+          [[ -n "${STAGING_KIT_CRON_SECRET:-}" ]] && supabase secrets set KIT_CRON_SECRET="${STAGING_KIT_CRON_SECRET}"
           supabase functions deploy stripe-webhook billing-stripe waitlist-capture waitlist-ops kit-sync \
             --project-ref "${STAGING_SUPABASE_PROJECT_REF}"
 
--- a/apps/web/src/pages/SignupPage.tsx
+++ b/apps/web/src/pages/SignupPage.tsx
@@ -24,6 +24,14 @@ import {
 } from '@beakerstack/waitlist/web';
 import { beakerstackWaitlistConfig } from '../waitlist/beakerstackWaitlistConfig';
 
+function signupPlanMetadata(sp: URLSearchParams): { plan_id: string } | undefined {
+  const planId = sp.get('plan');
+  if (!planId) return undefined;
+  const cfg = beakerstackBillingConfig.plans.find(p => p.id === planId);
+  if (!cfg || cfg.priceCents === 0) return undefined;
+  return { plan_id: planId };
+}
+
 function SignupPageContent() {
   const [email, setEmail] = useState('');
   const [password, setPassword] = useState('');
@@ -76,7 +84,8 @@ function SignupPageContent() {
     setError(null);
 
     try {
-      await auth.signUp(email, password);
+      const planMeta = signupPlanMetadata(searchParams);
+      await auth.signUp(email, password, planMeta ? { data: planMeta } : undefined);
       const {
         data: { session },
       } = await supabase.auth.getSession();
@@ -134,7 +143,9 @@ function SignupPageContent() {
     if (!planId) return undefined;
     const cfg = beakerstackBillingConfig.plans.find(p => p.id === planId);
     if (!cfg || cfg.priceCents === 0) return undefined;
-    return { [PLAN_INTEREST_METADATA_KEY]: cfg.displayName };
+    // Include both stable plan_id (for kit-sync interest tagging and waitlist-ops approve)
+    // and the display name (for human-readable metadata in the waitlist dashboard).
+    return { [PLAN_INTEREST_METADATA_KEY]: cfg.displayName, plan_id: planId };
   }, [isWaitlist, paidIntent, searchParams]);
 
   if (awaitingEmail) {
--- a/packages/shared/src/hooks/useAuth.ts
+++ b/packages/shared/src/hooks/useAuth.ts
@@ -71,7 +71,11 @@ export function useAuth(supabaseClient: SupabaseClient): AuthHookReturn {
     setLoading(false);
   };
 
-  const signUp = async (email: string, password: string): Promise<void> => {
+  const signUp = async (
+    email: string,
+    password: string,
+    options?: { data?: Record<string, unknown> }
+  ): Promise<void> => {
     setLoading(true);
     setError(null);
 
@@ -83,7 +87,10 @@ export function useAuth(supabaseClient: SupabaseClient): AuthHookReturn {
     const { error } = await supabaseClient.auth.signUp({
       email,
       password,
-      ...(emailRedirectTo ? { options: { emailRedirectTo } } : {}),
+      options: {
+        ...(emailRedirectTo ? { emailRedirectTo } : {}),
+        ...(options?.data ? { data: options.data } : {}),
+      },
     });
 
     setLoading(false);
--- a/packages/shared/src/types/auth.ts
+++ b/packages/shared/src/types/auth.ts
@@ -9,7 +9,7 @@ export interface AuthState {
 
 export interface AuthActions {
   signIn: (email: string, password: string) => Promise<void>;
-  signUp: (email: string, password: string) => Promise<void>;
+  signUp: (email: string, password: string, options?: { data?: Record<string, unknown> }) => Promise<void>;
   signOut: () => Promise<void>;
   signInWithGoogle: () => Promise<void>;
   requestPasswordReset: (email: string) => Promise<void>;
--- a/supabase/functions/kit-sync/index.ts
+++ b/supabase/functions/kit-sync/index.ts
@@ -117,6 +117,22 @@ async function handleWorker(
 
       const config = settings.config;
 
+      // Guard: namespace is required for all Kit tag operations. Dead-letter with a
+      // clear message so the operator knows exactly what to fix, rather than letting
+      // the row reach Kit API and fail with a confusing tag-not-found error.
+      if (!config.namespace) {
+        const { error: upErr } = await admin
+          .from('marketing_email_sync_queue')
+          .update({
+            status: 'failed',
+            error: `marketing_email_settings.config missing required field "namespace" for product "${row.product_id}" — insert or update the row in marketing_email_settings with a non-empty namespace`,
+          })
+          .eq('id', row.id);
+        if (upErr) console.error('Failed to dead-letter config-invalid row', row.id, upErr.message);
+        failed++;
+        continue;
+      }
+
       // Suppression check
       const { data: suppressed } = await admin
         .from('marketing_email_unsubscribes')
@@ -192,7 +208,7 @@ async function processEvent(
     case 'user.signed_up': {
       if (formId) await kit.subscribeToForm(email, formId);
       await kit.applyTag(email, signupTag(ns, opts));
-      const rawPlanId = payload.plan_id as string | undefined;
+      const rawPlanId = typeof payload.plan_id === 'string' && payload.plan_id ? payload.plan_id : undefined;
       if (rawPlanId) {
         const slug = planIdToSlug(rawPlanId, tierTagNames);
         await kit.applyTag(email, interestTag(ns, slug, opts));
@@ -202,11 +218,21 @@ async function processEvent(
     case 'waitlist.joined': {
       if (formId) await kit.subscribeToForm(email, formId);
       await kit.applyTag(email, waitlistTag(ns, opts));
+      const rawPlanId = typeof payload.plan_id === 'string' && payload.plan_id ? payload.plan_id : undefined;
+      if (rawPlanId) {
+        const slug = planIdToSlug(rawPlanId, tierTagNames);
+        await kit.applyTag(email, interestTag(ns, slug, opts));
+      }
       break;
     }
     case 'waitlist.approved': {
       if (formId) await kit.subscribeToForm(email, formId);
       await kit.applyTag(email, waitlistApprovedTag(ns, opts));
+      const rawPlanId = typeof payload.plan_id === 'string' && payload.plan_id ? payload.plan_id : undefined;
+      if (rawPlanId) {
+        const slug = planIdToSlug(rawPlanId, tierTagNames);
+        await kit.applyTag(email, interestTag(ns, slug, opts));
+      }
       break;
     }
     case 'waitlist.converted': {
@@ -216,7 +242,11 @@ async function processEvent(
     }
     case 'user.tier_changed': {
       // stripe-webhook enqueues with { user_id, plan_id, status }.
-      const rawPlanId = payload.plan_id as string | undefined;
+      if (tierTagNames.length === 0) throw new KitClientError(
+        `user.tier_changed requires tierTagNames in marketing_email_settings.config for product "${row.product_id}" — add tier slugs (e.g. ["pro","max"]) to config`,
+        'kit_api_400'
+      );
+      const rawPlanId = typeof payload.plan_id === 'string' && payload.plan_id ? payload.plan_id : undefined;
       if (!rawPlanId) throw new KitClientError(
         'user.tier_changed payload missing plan_id',
         'kit_api_400'
--- a/supabase/functions/waitlist-capture/index.ts
+++ b/supabase/functions/waitlist-capture/index.ts
@@ -82,12 +82,21 @@ Deno.serve(async req => {
   }
   const productId = envProductId || 'beakerstack';
 
+  const metadata = body.metadata ?? {};
+  // Hoist plan_id to the top level of the queue payload so the kit-sync worker
+  // can apply the interest tag without digging into the metadata envelope.
+  // Validate it's a non-empty string so a malformed client can't poison the queue.
+  const rawPlanId = metadata.plan_id;
+  const planId = typeof rawPlanId === 'string' && rawPlanId.trim() ? rawPlanId.trim() : null;
   await enqueueMarketingEmail(
     admin,
     productId,
     'waitlist.joined',
     email,
-    { metadata: body.metadata ?? {} },
+    {
+      metadata,
+      ...(planId ? { plan_id: planId } : {}),
+    },
     `waitlist.joined:${email}`
   );
 
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -137,9 +137,11 @@ jobs:
             BILLING_SUPABASE_URL="${PRODUCTION_SUPABASE_URL}" \
             BILLING_SUPABASE_ANON_KEY="${PRODUCTION_SUPABASE_ANON_KEY}" \
             BILLING_SUPABASE_SERVICE_ROLE_KEY="${PRODUCTION_SUPABASE_SERVICE_ROLE_KEY}" \
-            BILLING_ALLOWED_ORIGINS="${PRODUCTION_BILLING_ALLOWED_ORIGINS}" \
-            KIT_API_KEY="${PRODUCTION_KIT_API_KEY}" \
-            KIT_CRON_SECRET="${PRODUCTION_KIT_CRON_SECRET}"
+            BILLING_ALLOWED_ORIGINS="${PRODUCTION_BILLING_ALLOWED_ORIGINS}"
+          # Only set Kit secrets when the GitHub secrets are configured; passing empty
+          # strings would clobber any manually-configured Supabase project secrets.
+          [[ -n "${PRODUCTION_KIT_API_KEY:-}" ]] && supabase secrets set KIT_API_KEY="${PRODUCTION_KIT_API_KEY}"
+          [[ -n "${PRODUCTION_KIT_CRON_SECRET:-}" ]] && supabase secrets set KIT_CRON_SECRET="${PRODUCTION_KIT_CRON_SECRET}"
           supabase functions deploy stripe-webhook billing-stripe waitlist-capture waitlist-ops kit-sync \
             --project-ref "${PRODUCTION_SUPABASE_PROJECT_REF}"
 


.github/workflows/deploy-staging.yml
```diff
@@ -110,9 +110,11 @@ jobs:
             BILLING_SUPABASE_URL="${STAGING_SUPABASE_URL}" \
             BILLING_SUPABASE_ANON_KEY="${STAGING_SUPABASE_ANON_KEY}" \
             BILLING_SUPABASE_SERVICE_ROLE_KEY="${STAGING_SUPABASE_SERVICE_ROLE_KEY}" \
-            BILLING_ALLOWED_ORIGINS="${STAGING_BILLING_ALLOWED_ORIGINS}" \
-            KIT_API_KEY="${STAGING_KIT_API_KEY}" \
-            KIT_CRON_SECRET="${STAGING_KIT_CRON_SECRET}"
+            BILLING_ALLOWED_ORIGINS="${STAGING_BILLING_ALLOWED_ORIGINS}"
+          # Only set Kit secrets when the GitHub secrets are configured; passing empty
+          # strings would clobber any manually-configured Supabase project secrets.
+          [[ -n "${STAGING_KIT_API_KEY:-}" ]] && supabase secrets set KIT_API_KEY="${STAGING_KIT_API_KEY}"
```
ZappoMan, line 143: ✅ **#315 blocker 1 addressed.** Splitting Kit secrets into conditional `supabase secrets set` calls prevents empty GitHub secrets from overwriting manually configured Supabase project secrets. Same pattern should stay in sync on production (looks good there too).

Minor: each Kit secret is a separate CLI invocation — that's fine (Supabase merges secrets), but worth noting in deploy docs so operators know partial Kit config (API key only, no cron secret) is supported.

apps/web/src/pages/SignupPage.tsx
```diff
@@ -76,7 +76,8 @@ function SignupPageContent() {
     setError(null);
 
     try {
-      await auth.signUp(email, password);
+      const planId = searchParams.get('plan');
+      await auth.signUp(email, password, planId ? { data: { plan_id: planId } } : undefined);
```
ZappoMan: ⚠️ **#315 blocker 2 — mostly, but validate `plan_id` here.**

This passes **any** non-empty `?plan=` query param into `raw_user_meta_data`, including invalid IDs or the free plan. Waitlist capture metadata (lines 133–141) correctly restricts to a **paid public plan** from `beakerstackBillingConfig`.

Recommend reusing the same guard for email signup, e.g. only pass metadata when `waitlistCaptureMetadata` would include `plan_id`, or extract a shared helper:

```ts
function signupPlanMetadata(searchParams: URLSearchParams) {
  const planId = searchParams.get('plan');
  if (!planId) return undefined;
  const cfg = beakerstackBillingConfig.plans.find(p => p.id === planId);
  if (!cfg || cfg.priceCents === 0 || !cfg.isPublic) return undefined;
  return { plan_id: planId };
}
```

Without this, a crafted `?plan=beakerstack_free` or `?plan=garbage` URL produces interest tags or queue metadata we explicitly said should not happen (#286 intent vs tier distinction).

supabase/functions/waitlist-capture/index.ts
```diff
@@ -82,12 +82,19 @@ Deno.serve(async req => {
   }
   const productId = envProductId || 'beakerstack';
 
+  const metadata = body.metadata ?? {};
+  // Hoist plan_id to the top level of the queue payload so the kit-sync worker
+  // can apply the interest tag without digging into the metadata envelope.
+  const planId = (metadata.plan_id as string | undefined) ?? null;
```
ZappoMan: ✅ **#315 blocker 2 — waitlist path addressed.** Hoisting `plan_id` to the top-level payload is the right shape for the worker. Client-side waitlist form already sends `plan_id` in metadata after the SignupPage change.

Optional hardening: coerce/validate `plan_id` is a non-empty string before hoisting (ignore non-string metadata values).

supabase/functions/kit-sync/index.ts
```diff
@@ -117,6 +117,22 @@ async function handleWorker(
...
 
+      // Guard: namespace is required for all Kit tag operations. Dead-letter with a
+      // clear message so the operator knows exactly what to fix, rather than letting
+      // the row reach Kit API and fail with a confusing tag-not-found error.
+      if (!config.namespace) {
+        const { error: upErr } = await admin
+          .from('marketing_email_sync_queue')
+          .update({
+            status: 'failed',
+            error: `marketing_email_settings.config missing required field "namespace" for product "${row.product_id}" — set it in the admin or via ensure-kit-sync-settings script`,
```
ZappoMan: ⚠️ **#315 blocker 3 — partial fix.**

Namespace guard is helpful, but #315 also called out missing **`kitFormId`** and **`tierTagNames`** — those still produce confusing `kit_tag_not_found` permanent dead-letters when unset.

Also: the error references **`ensure-kit-sync-settings` script**, which doesn't exist in this PR or the repo. That will mislead operators.

Suggestions:
1. Extend the guard (or a single `validateKitConfig()`) to check `namespace`, `kitFormId`, and non-empty `tierTagNames` before any Kit calls.
2. Remove the script reference until it exists, or add the script/docs in this PR / a immediate follow-up.
3. Consider listing required Kit dashboard tags in a `docs/recipes/kit-first-30-minutes.md` snippet (#315 acceptance criteria).

supabase/functions/kit-sync/index.ts
```diff
@@ -202,11 +218,21 @@ async function processEvent(
     case 'waitlist.joined': {
       if (formId) await kit.subscribeToForm(email, formId);
       await kit.applyTag(email, waitlistTag(ns, opts));
+      const rawPlanId = payload.plan_id as string | undefined;
```
ZappoMan: ✅ **#315 blocker 2 — worker side addressed.** Interest tag on `waitlist.joined` mirrors `user.signed_up` and uses `planIdToSlug()` — good.

Same pattern on `waitlist.approved` (below) closes the approve path that reads `metadata.plan_id` from the waitlist entry.

packages/shared/src/hooks/useAuth.ts
```diff
@@ -83,7 +87,10 @@ export function useAuth(supabaseClient: SupabaseClient): AuthHookReturn {
     const { error } = await supabaseClient.auth.signUp({
       email,
       password,
-      ...(emailRedirectTo ? { options: { emailRedirectTo } } : {}),
+      options: {
```
ZappoMan, line 90: Looks correct for forwarding signup metadata. One subtle change from before: `options: { ... }` is **always** passed now (even when empty `{}`), whereas previously `options` was omitted entirely when there was no `emailRedirectTo`. Supabase should tolerate this, but if you see regressions in signup confirm emails, this is the diff to check.

Consider adding a unit test that `signUp(email, pw, { data: { plan_id: 'x' } })` forwards `options.data` to `auth.signUp`.

packages/shared/src/types/auth.ts
```diff
@@ -9,7 +9,7 @@ export interface AuthState {
 
 export interface AuthActions {
   signIn: (email: string, password: string) => Promise<void>;
-  signUp: (email: string, password: string) => Promise<void>;
+  signUp: (email: string, password: string, options?: { data?: Record<string, unknown> }) => Promise<void>;
```
ZappoMan, line 12: Type surface updated correctly. `useAuth.native.ts` still implements the old 2-arg `signUp` — that's structurally compatible with the optional third param, but mobile won't pass `plan_id` until wired. Fine if web-only for now; worth a one-line comment in the PR or a tracked follow-up if mobile signup gets `?plan=` support.

ZappoMan (commented):
## Review vs #315

Good targeted follow-up — the three **blockers from #315 are mostly addressed**, but I would not consider #315 fully closed until the gaps called out inline are resolved (especially operator setup and email signup `plan_id` validation).

### #315 blocker coverage

| #315 item | Status | Notes |
|-----------|--------|-------|
| **1. Deploy clobbering Kit secrets** | ✅ **Fixed** | Conditional `supabase secrets set` for `KIT_*` matches the cron ensure pattern. |
| **2. `plan_id` → interest tag pipeline** | ⚠️ **Mostly fixed** | Email signup metadata, waitlist metadata, capture hoist, and worker handlers for `waitlist.joined` / `waitlist.approved` look correct. OAuth still deferred (acknowledged). Email signup passes **any** `?plan=` value without the paid-plan validation used on the waitlist path — see inline. |
| **3. Operator enable path** | ⚠️ **Partial** | Namespace guard is a good start, but `kitFormId` / `tierTagNames` still fail at Kit with `kit_tag_not_found`. Error message references `ensure-kit-sync-settings` script that **does not exist** in this PR or repo. No setup script/docs yet (#315 asked for one). |

### #315 high-priority items (expected out of scope)

| Item | Status |
|------|--------|
| **4. Hosted cron smoke test** | Not in PR — still manual before first enable |
| **5. Bulk-sync billing tier backfill** | Correctly deferred; PR documents limitation |

### Other gaps

- **No automated tests** for the new behavior (SignupPage metadata, waitlist-capture hoist, kit-sync interest handlers, config guard).
- **`useAuth.native.ts`** unchanged — fine for web-first, but mobile email signup won't propagate `plan_id` if that path is added later.

### Verdict

**Approve directionally** — merge is reasonable for blocker #1 and the core of #2. Recommend a quick follow-up (or amendments here) for: validated `plan_id` on email signup, config guard completeness / error message, and at least one integration or unit test before marking #315 done.

mei-artificerinnovations:
**Heads up — conflict with #317 on `SignupPage.tsx`.**

Diego's PR #317 (fix/312-signup-pending-ux) also modifies `handleSignup` in `SignupPage.tsx`: it removes the `else { navigate('/dashboard') }` branch so all no-session signups hit the `awaitingEmail` pending UI. This PR doesn't have that fix.

Updating this branch now to incorporate both changes so the merged `handleSignup` has:
- `plan_id` passed to `auth.signUp` (this PR)
- No silent `/dashboard` redirect for no-session cases (PR #317's fix)

Whoever merges second (or if they land together) needs the version below.

.github/workflows/deploy-staging.yml
mei-artificerinnovations: Agreed — will add a doc note about partial Kit config (API key only, no cron secret) being supported. Tracking as a follow-up so it does not block this PR.

apps/web/src/pages/SignupPage.tsx
mei-artificerinnovations: Fixed in commit 4e440b8. Extracted a `signupPlanMetadata()` helper (line 27) that mirrors the same guard used in `waitlistCaptureMetadata`: finds the plan in `beakerstackBillingConfig`, rejects if `priceCents === 0` (free tier), returns `undefined` otherwise. The `handleSignup` call now uses `planMeta ? { data: planMeta } : undefined`, so invalid or free-plan IDs no longer reach `raw_user_meta_data`.

supabase/functions/waitlist-capture/index.ts
mei-artificerinnovations: Added the string guard in 4e440b8: `typeof rawPlanId === 'string' && rawPlanId.trim() ? rawPlanId.trim() : null`. Non-string metadata values (number, boolean, object) are now dropped silently rather than cast into the payload.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Three changes in 4e440b8:

1. **Script reference removed.** Error message now reads: "insert or update the row in marketing_email_settings with a non-empty namespace" — actionable without referencing a nonexistent script.

2. **`tierTagNames` guard for `user.tier_changed`.** Added a dead-letter check (`tierTagNames.length === 0`) before the tier swap logic. Without configured tier slugs, the worker cannot know which tags to remove, so it fails fast with a clear config error rather than producing a `kit_tag_not_found` retry loop.

3. **`kitFormId`** — intentionally not guarded globally. It is optional by design (gated with `if (formId)` at each call site); absent means no form subscription but tags still apply. A blanket guard would break products that use tag-only flows.

On `docs/recipes/kit-first-30-minutes.md`: can open a follow-up issue if that is an acceptance criterion for this PR or a separate task.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Confirmed — the same pattern now has `typeof payload.plan_id === 'string'` guards in all three cases (`user.signed_up`, `waitlist.joined`, `waitlist.approved`) as of 4e440b8.

packages/shared/src/hooks/useAuth.ts
mei-artificerinnovations: Good catch on the always-passed `options`. That is intentional — it avoids the spread conditional becoming a maintenance trap as more fields get added. Supabase accepts an empty options object fine. Unit test for `signUp` forwarding `options.data` is a reasonable ask; adding to the test follow-up ticket rather than holding this PR since there are no existing `useAuth` unit tests in the repo to extend.

packages/shared/src/types/auth.ts
mei-artificerinnovations: Correct — `useAuth.native.ts` still has the 2-arg signature which is structurally compatible with the optional third param. Mobile signup does not currently support `?plan=` so this is intentionally web-only for now. Can track mobile plan intent as a separate issue when that path is needed.

supabase/functions/waitlist-capture/index.ts
```diff
@@ -82,12 +82,19 @@ Deno.serve(async req => {
...
   await enqueueMarketingEmail(
     admin,
     productId,
     'waitlist.joined',
     email,
-    { metadata: body.metadata ?? {} },
+    {
+      metadata,
+      ...(planId ? { plan_id: planId } : {}),
+    },
```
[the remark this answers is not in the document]
mei-artificerinnovations: Fixed in 4e440b8 — `typeof rawPlanId === 'string' && rawPlanId.trim()` guard added before hoisting.

supabase/functions/kit-sync/index.ts
```diff
@@ -202,11 +218,21 @@ async function processEvent(
...
+      if (rawPlanId) {
+        const slug = planIdToSlug(rawPlanId, tierTagNames);
+        await kit.applyTag(email, interestTag(ns, slug, opts));
+      }
```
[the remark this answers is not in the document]
mei-artificerinnovations: Fixed in 4e440b8 — all three plan_id reads in processEvent now use `typeof payload.plan_id === 'string' && payload.plan_id` instead of a cast.

supabase/functions/kit-sync/index.ts
```diff
@@ -202,11 +218,21 @@ async function processEvent(
...
       break;
     }
     case 'waitlist.approved': {
       if (formId) await kit.subscribeToForm(email, formId);
       await kit.applyTag(email, waitlistApprovedTag(ns, opts));
+      const rawPlanId = payload.plan_id as string | undefined;
+      if (rawPlanId) {
+        const slug = planIdToSlug(rawPlanId, tierTagNames);
+        await kit.applyTag(email, interestTag(ns, slug, opts));
+      }
```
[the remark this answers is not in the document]
mei-artificerinnovations: Fixed in 4e440b8 — same typeof string guard applied here.

supabase/functions/kit-sync/index.ts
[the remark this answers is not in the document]
mei-artificerinnovations: Fixed in 4e440b8 — script reference removed. Message now points to the actual fix path (updating the row in marketing_email_settings).

apps/web/src/pages/SignupPage.tsx
```diff
@@ -76,7 +76,8 @@ function SignupPageContent() {
...
       const {
```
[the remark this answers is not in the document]
mei-artificerinnovations: Fixed in 4e440b8 — `signupPlanMetadata()` validates against `beakerstackBillingConfig.plans` and rejects free/unknown IDs before they reach auth.signUp.

packages/shared/src/hooks/useAuth.ts
```diff
@@ -83,7 +87,10 @@ export function useAuth(supabaseClient: SupabaseClient): AuthHookReturn {
...
+        ...(emailRedirectTo ? { emailRedirectTo } : {}),
+        ...(options?.data ? { data: options.data } : {}),
+      },
     });
```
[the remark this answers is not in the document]
mei-artificerinnovations: Noted — tracking as a test follow-up. No existing useAuth unit tests in the repo to extend right now; will cover in the test-coverage issue.

apps/web/src/pages/SignupPage.tsx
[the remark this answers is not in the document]
mei-artificerinnovations: Tracked in the same test follow-up. The signupPlanMetadata helper is pure and easy to unit-test independently; will include that in the coverage pass.

.github/workflows/deploy-staging.yml
ZappoMan: Created tracking issue: #325 (Kit Phase 3 post-#316 follow-ups — operator setup, tests, cron verification). No follow-up ticket existed before this re-review.

supabase/functions/kit-sync/index.ts
ZappoMan: Created tracking issue: #325 (Kit Phase 3 post-#316 follow-ups — operator setup, tests, cron verification). No follow-up ticket existed before this re-review.

packages/shared/src/hooks/useAuth.ts
ZappoMan: Created tracking issue: #325 (Kit Phase 3 post-#316 follow-ups — operator setup, tests, cron verification). No follow-up ticket existed before this re-review.

packages/shared/src/hooks/useAuth.ts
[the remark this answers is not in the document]
ZappoMan: Created tracking issue: #325 (Kit Phase 3 post-#316 follow-ups — operator setup, tests, cron verification). No follow-up ticket existed before this re-review.

apps/web/src/pages/SignupPage.tsx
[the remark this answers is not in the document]
ZappoMan: Created tracking issue: #325 (Kit Phase 3 post-#316 follow-ups — operator setup, tests, cron verification). No follow-up ticket existed before this re-review.

supabase/functions/kit-sync/index.ts
ZappoMan: ⚠️ **Re-review regression:** `waitlistCaptureMetadata` (line 146) still omits `plan_id` — only `plan_interest` is stored. First commit had both; this rebase dropped `plan_id` from waitlist entry metadata, breaking the approve → queue → interest tag path. Email signup is fixed via `signupPlanMetadata()`; waitlist path still needs `plan_id: planId` restored here.

.github/workflows/deploy-staging.yml
```diff
@@ -137,9 +110,11 @@ jobs:
...
```
ZappoMan, line 143: ✅ Kit secret gating resolves #315 blocker 1.

🚨 **Same PR also removes unrelated `develop` deploy steps** (rebase regression): job `if: push to develop`, Resend `sync-supabase-auth-config.sh` step, and header comments. Rebase onto current `develop` and keep only these `[[ -n ]]` Kit lines.

apps/web/src/pages/SignupPage.tsx
```diff
@@ -24,6 +24,14 @@ import {
 } from '@beakerstack/waitlist/web';
 import { beakerstackWaitlistConfig } from '../waitlist/beakerstackWaitlistConfig';
 
+function signupPlanMetadata(sp: URLSearchParams): { plan_id: string } | undefined {
```
ZappoMan, line 27: ✅ Resolves prior inline comment on unvalidated email signup `plan_id`.

Optional: also reject non-`isPublic` plan IDs to match `resolvePostAuthDestination()`.

See thread reply on waitlist metadata — `waitlistCaptureMetadata` below still needs `plan_id` restored.

ZappoMan (commented):
## Re-review vs #315 (commit `28a63a8`)

Most prior inline items are **resolved**. Two **new blockers** appeared in the rebase.

### Prior inline items — status

| Item | Status |
|------|--------|
| Deploy Kit secret gating | ✅ Resolved |
| Email signup `plan_id` validation (`signupPlanMetadata`) | ✅ Resolved |
| Waitlist capture `plan_id` hoist + string guard | ✅ Resolved |
| Worker interest tags + `typeof plan_id` guards | ✅ Resolved |
| Namespace config guard (no phantom script ref) | ✅ Resolved |
| `tierTagNames` guard on `user.tier_changed` | ✅ Resolved |
| `useAuth` metadata forwarding | ✅ Resolved |
| Native `useAuth` parity | ⏭️ #325 |
| Automated tests | ⏭️ **#325** (created; linked in threads) |
| Operator setup docs/script | ⏭️ **#325** |
| Deploy partial-Kit-config doc | ⏭️ **#325** (linked) |
| OAuth `plan_id` | ⏭️ #325 |

### Follow-up ticket

No issue existed when Mei replied "tracking as follow-up." Created **[#325](https://github.com/Artificer-Innovations/BeakerStack/issues/325)** and linked it in those threads.

### New blockers (rebase — fix before merge)

1. **`deploy-staging.yml` / `deploy-production.yml`** — drops Resend auth sync step (`sync-supabase-auth-config.sh`) and job-level branch guards that exist on current `develop` (#317 email automation). Rebase onto current `develop` and keep only the Kit `[[ -n ]]` secret lines.
2. **`waitlistCaptureMetadata`** — `plan_id` missing from waitlist entry metadata again (see reply on waitlist thread). Breaks approve → interest tag path.

### #315 closure

Blockers 1–2 addressed in code (modulo waitlist metadata regression). Blocker 3 partially done; operator docs in #325. Cron verification + bulk-sync remain in #325.

supabase/functions/waitlist-capture/index.ts
ZappoMan: ⚠️ **Re-review note (SignupPage, not this file):** `waitlistCaptureMetadata` still omits `plan_id` from waitlist entry metadata — only `plan_interest` is stored. That breaks `waitlist-ops approve` reading `metadata.plan_id`. Capture hoist + string guard here are fine; client must send `plan_id` in metadata first.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed in 080468d — `waitlistCaptureMetadata` now returns `{ [PLAN_INTEREST_METADATA_KEY]: cfg.displayName, plan_id: planId }` again. Was dropped when the rebase took develop's base which did not have this change; restored now.

.github/workflows/deploy-staging.yml
mei-artificerinnovations: Fixed in 080468d — took develop's version of both workflows as the base and applied only the Kit `[[ -n ]]` guard lines on top. Header comments, develop/main `if:` conditions, and the #298 Resend sync step are all preserved.

supabase/functions/waitlist-capture/index.ts
mei-artificerinnovations: Agreed — the hoist in waitlist-capture is correct; the client-side `waitlistCaptureMetadata` is what sends `plan_id` in the metadata. That field is restored in 080468d.

.github/workflows/deploy-staging.yml
mei-artificerinnovations: Noted — follow-ups tracked in #325.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Noted, tracked in #325.

packages/shared/src/hooks/useAuth.ts
mei-artificerinnovations: Noted, tracked in #325.

packages/shared/src/hooks/useAuth.ts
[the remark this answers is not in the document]
mei-artificerinnovations: Noted, tracked in #325.

apps/web/src/pages/SignupPage.tsx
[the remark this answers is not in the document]
mei-artificerinnovations: Noted, tracked in #325.

apps/web/src/pages/SignupPage.tsx
mei-artificerinnovations: Noted, tracked in #325.

Issue #312, opened 2026-05-21
Post-signup UX: show email confirmation pending state instead of redirecting home


ZappoMan:
## Context

Signup email confirmation is now enabled (`enable_confirmations = true` in `supabase/config.toml`, deployed via auth config sync). After a user submits the email/password signup form, Supabase creates the account but does **not** establish a session until they click the confirmation link in email.

**Observed behavior today:** On web, after clicking **Create account**, the app often navigates to `/dashboard` (or the user lands on the home page) while **not signed in**. That feels like a failed signup and does not tell the user what to do next.

**Expected behavior:** Show a clear **success / pending** state: “Check your email and click the confirmation link,” with optional next steps (sign in after confirming, resend link, etc.).

Related: email confirmation enablement and templates (#291 / PR #298). This issue is specifically about **client post-signup UX**, not SMTP or Supabase config.

---

## Business needs

1. **Reduce confusion and support burden** — Users should understand that signup succeeded and that email verification is the next step, not that the product rejected them or logged them out.
2. **Match production auth posture** — With confirmations required, the product demo and forks should model the same trustworthy flow users see on staging/production.
3. **Preserve billing / plan intent** — Users who sign up from `/signup?plan=…` still need a path to complete checkout after they confirm (post-auth redirect already exists for paid intent; general signup should not skip the pending state).
4. **Consistent cross-platform story** — Web and mobile should not diverge on what happens immediately after email signup when no session is returned.

---

### Web — `apps/web/src/pages/SignupPage.tsx`

After `auth.signUp()`:

- If `getSession()` returns a session → navigate to `postAuthPath` (dashboard or billing).
- Else if **paid plan intent** → set `awaitingEmail` and show inline **“Check your email”** UI (already implemented).
- Else → **`navigate('/dashboard')`** ← this is the problematic path when confirmations are on and there is no session.

The pending UI (`awaitingEmail`) is only shown for paid-intent signups. Plain `/signup` takes the dashboard redirect branch (see test: `navigates to dashboard after signup when no session and no paid plan intent`).

### Web — confirmation link

`/auth/confirm` (`AuthConfirmPage.tsx`) verifies OTP and redirects (e.g. dashboard). Post-auth redirect keys in `postAuthRedirect.ts` support resuming billing after confirmation for paid intent.

### Mobile — `apps/mobile/src/screens/SignupScreen.tsx`

Always `navigation.navigate('Dashboard')` after `signUp`, with no pending-email state.

[...]

PR #317, opened 2026-05-21
fix(#312): show email confirmation pending state after signup

diego-artificerinnovations:
## Summary

Fixes #312. After `auth.signUp()`, Supabase may not create a session if email confirmation is required. The old code silently redirected users to `/dashboard` (web) or `Dashboard` (mobile) as if signup succeeded — which looked like a broken login.

- **Web (`SignupPage.tsx`):** Remove the `else { navigate('/dashboard') }` branch. All no-session signups now show the existing `awaitingEmail` pending UI. Billing-specific copy ("we'll take you to billing") is conditionalized on `paidIntent`.
- **Mobile (`SignupScreen.tsx`):** Call `supabase.auth.getSession()` after signup; navigate to new `SignupPendingScreen` when no session, `Dashboard` when session is present.
- **New `SignupPendingScreen.tsx`:** Displays email, instruction copy, and "Already confirmed? Sign in" link. Registered in `AppNavigator` and typed in `RootStackParamList`.
- **Tests:** Web test rewritten to expect pending UI (not navigate to dashboard); two mobile tests split to cover both session / no-session paths.

## Test plan

- [ ] Plain signup on web with email confirmation on → should see "Check your email" screen, no navigation to dashboard
- [ ] Paid-plan signup on web with no session → should see "Check your email" with billing copy
- [ ] Session-returning signup on web → should navigate to dashboard as before
- [ ] Mobile signup with email confirmation on → should navigate to SignupPending screen with email displayed
- [ ] Mobile signup returning a session → should navigate to Dashboard
- [ ] "Already confirmed? Sign in" link on SignupPendingScreen navigates to Login
- [ ] All existing tests pass: `npm run test` in `apps/web` (18/18 SignupPage tests) and `apps/mobile` (231/231 tests)

### DIFF
--- a/apps/mobile/__tests__/screens/SignupScreen.test.tsx
+++ b/apps/mobile/__tests__/screens/SignupScreen.test.tsx
@@ -22,8 +22,13 @@ jest.mock('expo-constants', () => ({
 }));
 
 // Mock supabase
+const mockGetSession = jest.fn();
 jest.mock('../../src/lib/supabase', () => ({
-  supabase: {} as SupabaseClient,
+  supabase: {
+    auth: {
+      getSession: (...args: unknown[]) => mockGetSession(...args),
+    },
+  } as unknown as SupabaseClient,
 }));
 
 // Mock AppHeader
@@ -129,6 +134,7 @@ const renderWithProviders = (
 describe('SignupScreen', () => {
   beforeEach(() => {
     jest.clearAllMocks();
+    mockGetSession.mockResolvedValue({ data: { session: null }, error: null });
     jest.spyOn(Alert, 'alert').mockImplementation(() => {});
   });
 
@@ -221,7 +227,17 @@ describe('SignupScreen', () => {
     });
   });
 
-  it('handles successful signup', async () => {
+  it('navigates to Dashboard when signup returns a session', async () => {
+    const mockSession = {
+      access_token: 'mock-token',
+      refresh_token: 'mock-refresh',
+      expires_in: 3600,
+      expires_at: Date.now() + 3600000,
+      token_type: 'bearer',
+      user: { id: 'test-user-id', email: 'test@example.com' } as any,
+    };
+    mockGetSession.mockResolvedValueOnce({ data: { session: mockSession }, error: null });
+
     const mockClient = createMockSupabaseClient();
     const { getByPlaceholderText, getByText } = renderWithProviders(
       <SignupScreen navigation={mockNavigation} />,
@@ -234,7 +250,6 @@ describe('SignupScreen', () => {
 
     await waitFor(() => {
       const submitButton = getByText('Create Account');
-
       fireEvent.changeText(emailInput, 'test@example.com');
       fireEvent.changeText(passwordInput, 'password123');
       fireEvent.changeText(confirmPasswordInput, 'password123');
@@ -250,6 +265,34 @@ describe('SignupScreen', () => {
 
     await waitFor(() => {
       expect(mockNavigate).toHaveBeenCalledWith('Dashboard');
+    });
+  });
+
+  it('navigates to SignupPending when signup returns no session', async () => {
+    mockGetSession.mockResolvedValueOnce({ data: { session: null }, error: null });
+
+    const mockClient = createMockSupabaseClient();
+    const { getByPlaceholderText, getByText } = renderWithProviders(
+      <SignupScreen navigation={mockNavigation} />,
+      mockClient
+    );
+
+    const emailInput = getByPlaceholderText('Email address');
+    const passwordInput = getByPlaceholderText('Password');
+    const confirmPasswordInput = getByPlaceholderText('Confirm password');
+
+    await waitFor(() => {
+      const submitButton = getByText('Create Account');
+      fireEvent.changeText(emailInput, 'test@example.com');
+      fireEvent.changeText(passwordInput, 'password123');
+      fireEvent.changeText(confirmPasswordInput, 'password123');
+      fireEvent.press(submitButton);
+    });
+
+    await waitFor(() => {
+      expect(mockNavigate).toHaveBeenCalledWith('SignupPending', {
+        email: 'test@example.com',
+      });
     });
   });
 
--- a/apps/mobile/src/navigation/AppNavigator.tsx
+++ b/apps/mobile/src/navigation/AppNavigator.tsx
@@ -13,6 +13,7 @@ import BillingNavigator from '../navigation/BillingNavigator';
 import ForgotPasswordScreen from '../screens/ForgotPasswordScreen';
 import AuthCallbackScreen from '../screens/AuthCallbackScreen';
 import ResetPasswordScreen from '../screens/ResetPasswordScreen';
+import SignupPendingScreen from '../screens/SignupPendingScreen';
 import { useFeatureFlags } from '../config/featureFlags';
 import { navigationRef } from './navigationRef';
 import type { RootStackParamList } from './types';
@@ -68,6 +69,7 @@ export const AppNavigator = () => {
           component={BillingNavigator}
           options={{ headerShown: false }}
         />
+        <Stack.Screen name='SignupPending' component={SignupPendingScreen} />
       </Stack.Navigator>
     </NavigationContainer>
   );
--- a/apps/mobile/src/navigation/AppNavigator.tsx
+++ b/apps/mobile/src/navigation/AppNavigator.tsx
@@ -69,7 +69,11 @@ export const AppNavigator = () => {
           component={BillingNavigator}
           options={{ headerShown: false }}
         />
-        <Stack.Screen name='SignupPending' component={SignupPendingScreen} />
+        <Stack.Screen
+          name='SignupPending'
+          component={SignupPendingScreen}
+          options={{ title: 'Confirm your email' }}
+        />
       </Stack.Navigator>
     </NavigationContainer>
   );
--- a/apps/mobile/src/screens/SignupPendingScreen.tsx
+++ b/apps/mobile/src/screens/SignupPendingScreen.tsx
@@ -1,0 +1,72 @@
+import React from 'react';
+import {
+  View,
+  Text,
+  TouchableOpacity,
+  StyleSheet,
+  SafeAreaView,
+} from 'react-native';
+import { NativeStackScreenProps } from '@react-navigation/native-stack';
+import { colors } from '@beakerstack/shared/theme/colors';
+import type { RootStackParamList } from '../navigation/types';
+
+type Props = NativeStackScreenProps<RootStackParamList, 'SignupPending'>;
+
+export default function SignupPendingScreen({ route, navigation }: Props) {
+  const { email } = route.params;
+
+  return (
+    <SafeAreaView style={styles.container}>
+      <View style={styles.content}>
+        <Text style={styles.heading}>Check your email</Text>
+        <Text style={styles.body}>
+          We sent a confirmation link to{' '}
+          <Text style={styles.emailHighlight}>{email}</Text>. Click the link in
+          that email to finish creating your account.
+        </Text>
+        <TouchableOpacity
+          style={styles.linkButton}
+          onPress={() => navigation.navigate('Login')}
+        >
+          <Text style={styles.linkText}>Already confirmed? Sign in</Text>
+        </TouchableOpacity>
+      </View>
+    </SafeAreaView>
+  );
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    backgroundColor: colors.background,
+  },
+  content: {
+    flex: 1,
+    justifyContent: 'center',
+    paddingHorizontal: 24,
+  },
+  heading: {
+    fontSize: 24,
+    fontWeight: '600',
+    color: colors.text,
+    marginBottom: 12,
+  },
+  body: {
+    fontSize: 15,
+    color: colors.textSecondary,
+    lineHeight: 22,
+    marginBottom: 24,
+  },
+  emailHighlight: {
+    fontWeight: '600',
+    color: colors.text,
+  },
+  linkButton: {
+    alignSelf: 'flex-start',
+  },
+  linkText: {
+    fontSize: 15,
+    color: colors.primary,
+    fontWeight: '500',
+  },
+});
--- a/apps/mobile/src/screens/SignupPendingScreen.tsx
+++ b/apps/mobile/src/screens/SignupPendingScreen.tsx
@@ -38,7 +38,7 @@ export default function SignupPendingScreen({ route, navigation }: Props) {
 const styles = StyleSheet.create({
   container: {
     flex: 1,
-    backgroundColor: colors.background,
+    backgroundColor: colors.pageBg,
   },
   content: {
     flex: 1,
@@ -48,25 +48,25 @@ const styles = StyleSheet.create({
   heading: {
     fontSize: 24,
     fontWeight: '600',
-    color: colors.text,
+    color: colors.textPrimary,
     marginBottom: 12,
   },
   body: {
     fontSize: 15,
-    color: colors.textSecondary,
+    color: colors.textSubtle,
     lineHeight: 22,
     marginBottom: 24,
   },
   emailHighlight: {
     fontWeight: '600',
-    color: colors.text,
+    color: colors.textPrimary,
   },
   linkButton: {
     alignSelf: 'flex-start',
   },
   linkText: {
     fontSize: 15,
-    color: colors.primary,
+    color: colors.brand,
     fontWeight: '500',
   },
 });
--- a/apps/mobile/src/screens/SignupPendingScreen.tsx
+++ b/apps/mobile/src/screens/SignupPendingScreen.tsx
@@ -13,7 +13,7 @@ import type { RootStackParamList } from '../navigation/types';
 type Props = NativeStackScreenProps<RootStackParamList, 'SignupPending'>;
 
 export default function SignupPendingScreen({ route, navigation }: Props) {
-  const email = route.params?.email ?? '';
+  const email = route?.params?.email ?? '';
 
   return (
     <SafeAreaView style={styles.container}>
--- a/apps/mobile/src/screens/SignupPendingScreen.tsx
+++ b/apps/mobile/src/screens/SignupPendingScreen.tsx
@@ -13,7 +13,7 @@ import type { RootStackParamList } from '../navigation/types';
 type Props = NativeStackScreenProps<RootStackParamList, 'SignupPending'>;
 
 export default function SignupPendingScreen({ route, navigation }: Props) {
-  const { email } = route.params;
+  const email = route.params?.email ?? '';
 
   return (
     <SafeAreaView style={styles.container}>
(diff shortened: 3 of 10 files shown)


apps/web/src/pages/SignupPage.tsx
```diff
@@ -84,14 +84,14 @@ function SignupPageContent() {
...
-        );
-        setAwaitingEmail(true);
       } else {
-        navigate('/dashboard', { replace: true });
+        if (paidIntent && postAuthPath !== '/dashboard') {
+          localStorage.setItem(
+            POST_AUTH_REDIRECT_KEY,
+            serializePostAuthRedirectPayload(postAuthPath)
+          );
+        }
```
ZappoMan, line 93: Nice consolidation — single no-session path with conditional `localStorage` stash matches the plan we discussed. This fixes the bug without touching session/OAuth/error paths.

apps/web/src/pages/SignupPage.tsx
```diff
@@ -148,9 +148,11 @@ function SignupPageContent() {
             </h2>
             <p className='mt-3 text-sm text-gray-600 dark:text-gray-300'>
               We sent a confirmation link to <strong>{email}</strong>. Click the
-              link in that email to finish creating your account — we&apos;ll
-              take you to billing to complete your plan when you&apos;re signed
-              in.
+              link in that email to finish creating your account
+              {paidIntent && postAuthPath !== '/dashboard'
+                ? " — we’ll take you to billing to complete your plan when you’re signed in"
```
ZappoMan: Nit: this string uses curly/smart apostrophes (`we'll`, `you're`) while the rest of this file uses `&apos;` in JSX strings. Consider matching the existing convention for consistency, or split into two full sentences to avoid the em-dash fragment.

apps/mobile/src/screens/SignupScreen.tsx
```diff
@@ -24,6 +24,7 @@ type RootStackParamList = {
   Login: undefined;
   Signup: undefined;
   Dashboard: undefined;
+  SignupPending: { email: string };
```
ZappoMan: Since you're adding `SignupPending` here, consider importing the shared `RootStackParamList` from `../navigation/types` instead of maintaining a local duplicate. Keeps navigator registration and screen navigation types in sync.

apps/web/src/pages/__tests__/SignupPage.test.tsx
```diff
@@ -489,10 +489,33 @@ describe('SignupPage', () => {
     await user.click(screen.getByRole('button', { name: /create account/i }));
 
     await waitFor(() => {
-      expect(mockNavigate).toHaveBeenCalledWith('/dashboard', {
-        replace: true,
-      });
+      expect(screen.getByText('Check your email')).toBeInTheDocument();
+      expect(screen.getByText(/confirm@example\.com/)).toBeInTheDocument();
     });
+    expect(mockNavigate).not.toHaveBeenCalled();
```
ZappoMan, line 495: Good assertion that navigation does not fire. Optional: also assert `localStorage.getItem(POST_AUTH_REDIRECT_KEY)` is null here to lock in the "no stash for plain signup" behavior — the paid-intent test already covers the positive case.

apps/mobile/src/screens/SignupScreen.tsx
```diff
@@ -75,7 +76,12 @@ export default function SignupScreen({ navigation }: Props) {
 
     try {
       await auth.signUp(email, password);
-      navigation.navigate('Dashboard');
+      const { data: { session } } = await supabase.auth.getSession();
```
ZappoMan, line 72: Using `getSession()` after `signUp()` matches the web implementation — good for consistency. Minor note: Supabase also returns `{ data: { session } }` directly from `signUp()`; either approach works.

apps/mobile/src/screens/SignupPendingScreen.tsx
```diff
@@ -0,0 +1,72 @@
...
+        </Text>
+        <TouchableOpacity
+          style={styles.linkButton}
+          onPress={() => navigation.navigate('Login')}
+        >
+          <Text style={styles.linkText}>Already confirmed? Sign in</Text>
+        </TouchableOpacity>
+      </View>
+    </SafeAreaView>
+  );
```
ZappoMan, line 38: No dedicated test file for this screen. Not strictly required for v1, but a lightweight render test asserting heading, email param, and Login navigation would match the coverage added for `SignupScreen`.

apps/web/src/pages/__tests__/SignupPage.test.tsx
```diff
@@ -489,10 +489,33 @@ describe('SignupPage', () => {
...
+  });
+
+  it('does not show billing copy in pending UI for plain signup', async () => {
```
ZappoMan: These two tests share nearly identical setup. Consider merging into one test that asserts both the email display and absence of billing copy, or extracting a small helper.

apps/mobile/src/screens/SignupPendingScreen.tsx
```diff
@@ -0,0 +1,72 @@
...
+  SafeAreaView,
+} from 'react-native';
+import { NativeStackScreenProps } from '@react-navigation/native-stack';
+import { colors } from '@beakerstack/shared/theme/colors';
+import type { RootStackParamList } from '../navigation/types';
+
+type Props = NativeStackScreenProps<RootStackParamList, 'SignupPending'>;
+
+export default function SignupPendingScreen({ route, navigation }: Props) {
+  const email = route?.params?.email ?? '';
```
ZappoMan: `NativeStackScreenProps` already types `route.params.email` as required — the `route?.params?.email` optional chaining shouldn't be necessary in production. If this was added for tests, proper navigation props in a dedicated test would be cleaner than defensive runtime guards.

apps/mobile/__tests__/screens/SignupScreen.test.tsx
```diff
@@ -22,8 +22,13 @@ jest.mock('expo-constants', () => ({
 }));
 
 // Mock supabase
+const mockGetSession = jest.fn();
```
ZappoMan, line 25: Good approach splitting the module-level `mockGetSession` from the AuthProvider client mock — `SignupScreen` calls `supabase.auth.getSession()` on the singleton import, so this correctly intercepts the post-signup check.

apps/mobile/src/screens/SignupPendingScreen.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  return (
+    <SafeAreaView style={styles.container}>
```
ZappoMan, line 21: Every other auth screen (`SignupScreen`, `LoginScreen`, `ForgotPasswordScreen`, etc.) renders `<AppHeader supabaseClient={supabase} />`. This screen is missing it, so users lose the nav header after signup. Recommend adding AppHeader for parity with web's pending view.

apps/mobile/__tests__/screens/SignupScreen.test.tsx
```diff
@@ -253,6 +268,34 @@ describe('SignupScreen', () => {
     });
   });
 
+  it('navigates to SignupPending when signup returns no session', async () => {
```
ZappoMan, line 271: Coverage looks correct for both branches. Pre-existing note: when `getSession()` returns a session, the `useEffect` on `auth.user` may also fire `navigation.reset` to Dashboard — not introduced here, but worth watching in manual QA.

ZappoMan (commented):
## Review summary

Solid fix for #312. Diego's implementation matches the plan we discussed and correctly addresses the root cause on both platforms.

### What works well
- **Web:** Consolidated no-session branch — all email signups without a session show the existing `awaitingEmail` UI; billing redirect stash is preserved only for paid intent.
- **Web copy:** Conditional billing sentence avoids the awkward em-dash-only fragment for plain signup.
- **Mobile:** `getSession()` check + new `SignupPendingScreen` mirrors web behavior.
- **Tests:** Web test rewritten correctly (18/18 pass); mobile tests split for session/no-session paths (10/10 pass). Mock strategy for `supabase.auth.getSession` is correct.

### Suggested before/after merge (non-blocking)
1. Add `AppHeader` to `SignupPendingScreen` — every other auth screen has it; web pending view does too.
2. Drop unnecessary `route?.params?.` optional chaining — types already guarantee params.
3. Import shared `RootStackParamList` in `SignupScreen` instead of local duplicate.
4. Minor: smart apostrophe consistency in billing copy string.

### Follow-up (out of scope)
- Mobile `emailRedirectTo` + native confirm deep link (confirmation emails currently land on web `/auth/confirm`)
- Short docs note in `EMAIL_TEMPLATES.md` per issue AC

**Verdict:** Approve — nothing blocking merge from my side. Left inline comments on the specific lines.

apps/mobile/src/screens/SignupScreen.tsx
```diff
@@ -24,6 +24,7 @@ type RootStackParamList = {
   Login: undefined;
```
mei-artificerinnovations: Pre-existing issue: this local `RootStackParamList` duplicates the canonical one in `navigation/types.ts`. You correctly updated both, but the right fix long-term is to import from `'../navigation/types'` and remove the local declaration. Out of scope here — worth a follow-up ticket.

apps/mobile/src/navigation/AppNavigator.tsx
```diff
@@ -68,6 +69,7 @@ export const AppNavigator = () => {
           component={BillingNavigator}
           options={{ headerShown: false }}
         />
+        <Stack.Screen name='SignupPending' component={SignupPendingScreen} />
```
mei-artificerinnovations: Nit: without `options` this renders the route name (`"SignupPending"`) as the header title. Consider `options={{ title: 'Confirm your email' }}` or `options={{ headerShown: false }}`.

mei-artificerinnovations (approved):
Core logic is correct and the test split (session vs no-session) is clean. Two things to resolve before merge.

### ⚠️ Conflict with #316 (`SignupPage.tsx`)

PR #316 (my kit-sync plan_id pipeline fix) also modifies `apps/web/src/pages/SignupPage.tsx` — specifically `handleSignup`. It adds `plan_id` to the `auth.signUp` call but **did not pick up this PR's fix** (it still has the silent `navigate('/dashboard')` for the no-session case). The two PRs need to land with a merged handleSignup:

```typescript
const planId = searchParams.get('plan');
await auth.signUp(email, password, planId ? { data: { plan_id: planId } } : undefined);
const { data: { session } } = await supabase.auth.getSession();
if (session) {
  clearPostAuthRedirectKeys();
  navigate(postAuthPath, { replace: true });
} else {
  if (paidIntent && postAuthPath !== '/dashboard') {
    localStorage.setItem(POST_AUTH_REDIRECT_KEY, serializePostAuthRedirectPayload(postAuthPath));
  }
  setAwaitingEmail(true);
}
```

I'll update #316 with this merged version once we know merge order. Either way one of us needs to pick up the other's change before Brad merges.

### Nit: `SignupPending` needs a header title in `AppNavigator`

Without `options`, RN Stack will show the literal string `"SignupPending"` as the header title. Either add `options={{ title: 'Confirm your email' }}` or `options={{ headerShown: false }}` if the screen handles its own header.

apps/web/src/pages/SignupPage.tsx
diego-artificerinnovations: Thanks — agreed that consolidating to a single no-session path was the cleanest approach.

apps/web/src/pages/SignupPage.tsx
diego-artificerinnovations: Fixed — split the billing copy into its own separate `<p>` element so both sentences use `&apos;` entities in JSX text, matching the rest of the file. No more inline string concatenation.

apps/mobile/src/screens/SignupScreen.tsx
diego-artificerinnovations: Done — removed the local `RootStackParamList` definition and now import the shared one from `'../navigation/types'`. The local `SignupScreenNavigationProp` type still references the same canonical list.

apps/web/src/pages/__tests__/SignupPage.test.tsx
diego-artificerinnovations: Added — the merged test now includes `expect(window.localStorage.getItem(POST_AUTH_REDIRECT_KEY)).toBeNull()` to lock in the no-stash behavior for plain signups.

apps/mobile/src/screens/SignupScreen.tsx
diego-artificerinnovations: Agreed — `signUp()` does return a session in its response. Using `getSession()` afterward keeps this consistent with the web implementation (`SignupPage.tsx` does the same), so the session-check behavior is uniform across platforms.

apps/mobile/src/screens/SignupPendingScreen.tsx
diego-artificerinnovations: Added `__tests__/screens/SignupPendingScreen.test.tsx` with tests covering: renders heading and email from route params, and navigates to Login when pressing the sign-in link.

apps/web/src/pages/__tests__/SignupPage.test.tsx
diego-artificerinnovations: Merged — the two tests are now a single case: `'shows pending email confirmation UI after plain signup without billing copy or redirect stash'`. Asserts heading, email, no navigation, no billing text, and localStorage null in one flow.

apps/mobile/src/screens/SignupPendingScreen.tsx
diego-artificerinnovations: Removed the optional chaining — `route.params.email` is now used directly as typed. Fixed the root cause instead: added `jest.mock('../src/screens/SignupPendingScreen', ...)` in `App.test.tsx` and `AppNavigator.test.tsx` so the real component is never rendered without proper route props in those integration suites.

apps/mobile/__tests__/screens/SignupScreen.test.tsx
diego-artificerinnovations: Thanks — yes, the split was necessary because `signUp()` is called via the auth context client while `getSession()` is called on the singleton import, so they need separate mock surfaces.

apps/mobile/src/screens/SignupPendingScreen.tsx
diego-artificerinnovations: Added — `SignupPendingScreen` now renders `<AppHeader supabaseClient={supabase} />` matching the other auth screens.

apps/mobile/__tests__/screens/SignupScreen.test.tsx
diego-artificerinnovations: Noted — the `useEffect` that fires on `auth.user` is pre-existing behavior and out of scope here. Will flag for manual QA.

apps/mobile/src/screens/SignupScreen.tsx
diego-artificerinnovations: Agreed — the local duplicate is removed in this PR (replaced with `import type { RootStackParamList } from '../navigation/types'`). The broader cleanup to remove it from any other screens that might still have it is worth a follow-up ticket.

apps/mobile/src/navigation/AppNavigator.tsx
diego-artificerinnovations: Added `options={{ title: 'Confirm your email' }}` — avoids the route name leaking as the native header title.

apps/web/src/pages/SignupPage.tsx
```diff
@@ -148,9 +148,11 @@ function SignupPageContent() {
...
+                : ''}
+              .
```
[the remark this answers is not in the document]
diego-artificerinnovations: Fixed — billing copy is now a separate `<p>` element after the main paragraph, so the period is always attached cleanly and all apostrophes use `&apos;` entities in JSX text.

apps/mobile/src/screens/SignupScreen.tsx
```diff
@@ -24,6 +24,7 @@ type RootStackParamList = {
...
   Signup: undefined;
   Dashboard: undefined;
+  SignupPending: { email: string };
 };
```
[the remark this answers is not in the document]
diego-artificerinnovations: Removed — now imports `RootStackParamList` from `'../navigation/types'` and the local duplicate is gone.

apps/mobile/src/screens/SignupScreen.tsx
```diff
@@ -75,7 +76,12 @@ export default function SignupScreen({ navigation }: Props) {
 
     try {
       await auth.signUp(email, password);
-      navigation.navigate('Dashboard');
+      const { data: { session } } = await supabase.auth.getSession();
+      if (session) {
+        navigation.navigate('Dashboard');
+      } else {
+        navigation.navigate('SignupPending', { email });
+      }
```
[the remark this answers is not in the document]
diego-artificerinnovations: Valid observation. The web implementation (`SignupPage.tsx`) uses the same `supabase` singleton for `getSession()` post-signup, so this keeps both platforms consistent. The test suite correctly handles this by mocking the singleton import separately from the AuthProvider client. A refactor to expose `getSession` through the auth context is worth considering in a future pass — tracking as a follow-up.

apps/mobile/__tests__/screens/SignupScreen.test.tsx
ZappoMan: Tracked as follow-up: #319

apps/mobile/src/screens/SignupScreen.tsx
ZappoMan: Tracked as follow-up: #318

apps/mobile/src/screens/SignupScreen.tsx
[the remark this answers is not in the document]
ZappoMan: Agree this is worth fixing separately — tracked as #321. Not blocking #317 merge since production uses the same singleton, but unifying through auth context would simplify tests and remove the dual-client footgun.

ZappoMan (approved):
## Re-review (post feedback commit e4da5c8)

Diego addressed all review feedback from the first pass. Re-reviewed the latest diff and ran tests locally.

### Verified fixes
- Web: consolidated no-session branch + split billing copy into separate `<p>` with `&apos;` entities
- Web tests: merged pending test + `localStorage` null assertion (17/17 pass)
- Mobile: shared `RootStackParamList` import, `AppHeader` on pending screen, screen title option, `SignupPendingScreen` tests, navigator mocks
- CI: all checks green; PR is mergeable

### Follow-up issues filed
| Thread | Issue |
|--------|-------|
| RootStackParamList cleanup (other screens) | #318 |
| SignupScreen useEffect double navigation | #319 |
| Native mobile confirm deep link + `emailRedirectTo` | #320 |
| Auth context vs singleton `getSession` | #321 |
| Docs for post-signup pending UX | #322 |

Linked #318–#321 on the relevant inline threads.

### Blockers
**None for this PR.** Ready to merge.

**Coordination note (not a git conflict):** PR #316 also touches `SignupPage.tsx` (adds `plan_id` to signup metadata). Whichever lands second should rebase — functionally both changes belong in the same `handleSignup`. Git merge state is currently clean.

ZappoMan:
Additional follow-ups from the re-review (no specific inline thread):

- **#320** — Mobile native email confirmation (`emailRedirectTo` + deep-link `verifyOtp` screen). Pending UI in this PR is correct; confirm-link completion still happens on web.
- **#322** — Docs note in `EMAIL_TEMPLATES.md` for adopters enabling confirmations (optional AC from #312).

PR #340, opened 2026-05-22
feat(#286): kit-webhook Edge Function + user.deleted GDPR wiring (Phase 4)

mei-artificerinnovations:
## Summary

- **`kit-webhook` Edge Function** — inbound Kit Creator webhook handler with HMAC-SHA256 signature validation (constant-time comparison), dual rate limiting (pre-HMAC 200/hr + post-HMAC 60/hr per IP), handles `subscriber.unsubscribed` (Kit v3) and `subscriber.subscriber_unsubscribe` (Kit v4); routes unsubscribes to `marketing_email_unsubscribes`; returns 503 when `KIT_WEBHOOK_SECRET` is not configured
- **`user.deleted` GDPR wiring** — AFTER DELETE trigger on `auth.users` enqueues `user.deleted` into `marketing_email_sync_queue`; `kit-sync` adds a GDPR suppression bypass so erasure calls proceed even if the email appears in the unsubscribes list
- **Migration `20260522000001`** — extends `event_type` CHECK constraint, adds trigger + `_marketing_email_on_user_deleted()` function, adds `kit_webhook_check_rate_limit` RPC (reuses `waitlist_rate_limits` table with hour-truncated window)
- **pgTAP tests** — 14 assertions covering CHECK constraint, trigger existence, trigger behavior (idempotent, NULL email, disabled product), rate limit RPC (within/exceeds/independent-IP/window), and queue payload shape
- **Deploy wiring** — `KIT_WEBHOOK_SECRET` secret set in staging + production workflows; `kit-webhook` added to `supabase functions deploy` list

## Notes

- Audit row deferred to Phase 5: `admin_audit_log.actor_user_id` is NOT NULL — system-actor support requires a schema change (see TODO comment in webhook handler)
- `user.deleted` dispatch in `kit-sync` calls `kit.deleteUser(email)` — GDPR erasure best-effort per Brad's guidance

## Test plan

- [ ] `supabase db test` passes (14 pgTAP assertions in `marketing_email_phase4.test.sql`)
- [ ] CI green on staging deploy
- [ ] Set `STAGING_KIT_WEBHOOK_SECRET` GitHub secret → verify Kit dashboard can send a test event and it routes correctly
- [ ] Verify `supabase functions deploy` includes `kit-webhook`

Closes #286 Phase 4

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -121,6 +121,7 @@ jobs:
           STAGING_BILLING_ALLOWED_ORIGINS: ${{ secrets.STAGING_BILLING_ALLOWED_ORIGINS }}
           STAGING_KIT_API_KEY: ${{ secrets.STAGING_KIT_API_KEY }}
           STAGING_KIT_CRON_SECRET: ${{ secrets.STAGING_KIT_CRON_SECRET }}
+          STAGING_KIT_WEBHOOK_SECRET: ${{ secrets.STAGING_KIT_WEBHOOK_SECRET }}
         run: |
           set -euo pipefail
           cd supabase
@@ -139,7 +140,8 @@ jobs:
           # strings would clobber any manually-configured Supabase project secrets.
           [[ -n "${STAGING_KIT_API_KEY:-}" ]] && supabase secrets set KIT_API_KEY="${STAGING_KIT_API_KEY}"
           [[ -n "${STAGING_KIT_CRON_SECRET:-}" ]] && supabase secrets set KIT_CRON_SECRET="${STAGING_KIT_CRON_SECRET}"
-          supabase functions deploy stripe-webhook billing-stripe waitlist-capture waitlist-ops kit-sync \
+          [[ -n "${STAGING_KIT_WEBHOOK_SECRET:-}" ]] && supabase secrets set KIT_WEBHOOK_SECRET="${STAGING_KIT_WEBHOOK_SECRET}"
+          supabase functions deploy stripe-webhook billing-stripe waitlist-capture waitlist-ops kit-sync kit-webhook \
             --project-ref "${STAGING_SUPABASE_PROJECT_REF}"
 
       - name: Ensure kit-sync cron job (staging)
--- a/supabase/functions/kit-sync/index.ts
+++ b/supabase/functions/kit-sync/index.ts
@@ -152,7 +152,9 @@ async function handleWorker(
         .eq('email', row.email)
         .maybeSingle();
 
-      if (suppressed) {
+      // user.deleted is a GDPR erasure — it must run even if the user previously
+      // unsubscribed via Kit webhook. Suppression only governs marketing sends.
+      if (suppressed && row.event_type !== 'user.deleted') {
         await markDone(admin, row.id);
         skipped++;
         continue;
@@ -307,6 +309,12 @@ async function processEvent(
       if ((config.onChurn ?? 'tag_only') === 'unsubscribe') {
         await kit.unsubscribeUser(email);
       }
+      break;
+    }
+    case 'user.deleted': {
+      // GDPR erasure: hard-delete the subscriber from Kit. Bypasses suppression
+      // check above so this always runs regardless of unsubscribe status.
+      await kit.deleteUser(email);
       break;
     }
     default:
--- a/supabase/functions/kit-sync/index.ts
+++ b/supabase/functions/kit-sync/index.ts
@@ -123,10 +123,10 @@ async function handleWorker(
 
       const config = settings.config;
 
-      // Guard: namespace is required for all Kit tag operations. Dead-letter with a
-      // clear message so the operator knows exactly what to fix, rather than letting
-      // the row reach Kit API and fail with a confusing tag-not-found error.
-      if (!config.namespace) {
+      // Guard: namespace is required for tag operations, but not for GDPR erasure
+      // (deleteUser only needs API key + email). Carve out user.deleted so it is
+      // never dead-lettered by a missing namespace.
+      if (!config.namespace && row.event_type !== 'user.deleted') {
         const { error: upErr } = await admin
           .from('marketing_email_sync_queue')
           .update({
--- a/supabase/functions/kit-webhook/index.ts
+++ b/supabase/functions/kit-webhook/index.ts
@@ -41,7 +41,7 @@ async function verifyKitSignature(
   secret: string
 ): Promise<boolean> {
   if (!sigHeader?.startsWith('sha256=')) return false;
-  const expected = sigHeader.slice(7);
+  const expected = sigHeader.slice(7).toLowerCase();
   const key = await crypto.subtle.importKey(
     'raw',
     new TextEncoder().encode(secret),
@@ -68,12 +68,11 @@ async function checkRateLimit(
   bucketKey: string,
   limit: number
 ): Promise<boolean> {
-  const windowStart = new Date();
-  windowStart.setMinutes(0, 0, 0); // hour-truncated window
-
+  // Pass raw timestamp — the RPC applies date_trunc('hour', ...) server-side in UTC,
+  // avoiding any TZ skew from the Edge runtime's local clock.
   const { data, error } = await admin.rpc('kit_webhook_check_rate_limit', {
     p_bucket_key: bucketKey,
-    p_window_start: windowStart.toISOString(),
+    p_window_start: new Date().toISOString(),
     p_limit: limit,
   });
 
@@ -101,7 +100,9 @@ Deno.serve(async req => {
     return jsonResponse({ error: 'webhook_not_configured' }, 503);
   }
 
-  const admin = createClient(SUPABASE_URL, SUPABASE_SERVICE_ROLE_KEY);
+  const admin = createClient(SUPABASE_URL, SUPABASE_SERVICE_ROLE_KEY, {
+    auth: { persistSession: false },
+  });
   const ip = clientIp(req);
 
   // Pre-HMAC rate limit — caps request flood before spending CPU on signature verification.
--- a/supabase/functions/kit-webhook/index.ts
+++ b/supabase/functions/kit-webhook/index.ts
@@ -1,0 +1,207 @@
+// Inbound Kit Creator webhook handler.
+// Validates HMAC-SHA256 signature, rate-limits by IP, records unsubscribes.
+//
+// Env vars required:
+//   KIT_WEBHOOK_SECRET  — Kit dashboard → Webhooks → Signing secret
+//   SUPABASE_URL
+//   SUPABASE_SERVICE_ROLE_KEY
+
+import { createClient } from 'npm:@supabase/supabase-js@2.45.0';
+
+const KIT_WEBHOOK_SECRET = Deno.env.get('KIT_WEBHOOK_SECRET') ?? '';
+const SUPABASE_URL = Deno.env.get('SUPABASE_URL') ?? '';
+const SUPABASE_SERVICE_ROLE_KEY =
+  Deno.env.get('SUPABASE_SERVICE_ROLE_KEY') ?? '';
+
+// Pre-HMAC cap: limits raw request flood before signature verification (200/hr per IP).
+const RAW_RATE_LIMIT = 200;
+// Post-HMAC cap: limits valid-signature traffic (60/hr per IP).
+const SIGNED_RATE_LIMIT = 60;
+
+function jsonResponse(body: unknown, status: number): Response {
+  return new Response(JSON.stringify(body), {
+    status,
+    headers: { 'Content-Type': 'application/json' },
+  });
+}
+
+function clientIp(req: Request): string {
+  const cf = req.headers.get('cf-connecting-ip');
+  if (cf) return cf.trim();
+  const real = req.headers.get('x-real-ip');
+  if (real) return real.trim();
+  const fwd = req.headers.get('x-forwarded-for');
+  if (fwd) return fwd.split(',').at(-1)?.trim() ?? 'unknown';
+  return 'unknown';
+}
+
+async function verifyKitSignature(
+  body: ArrayBuffer,
+  sigHeader: string | null,
+  secret: string
+): Promise<boolean> {
+  if (!sigHeader?.startsWith('sha256=')) return false;
+  const expected = sigHeader.slice(7);
+  const key = await crypto.subtle.importKey(
+    'raw',
+    new TextEncoder().encode(secret),
+    { name: 'HMAC', hash: 'SHA-256' },
+    false,
+    ['sign']
+  );
+  const mac = await crypto.subtle.sign('HMAC', key, body);
+  const actual = Array.from(new Uint8Array(mac))
+    .map(b => b.toString(16).padStart(2, '0'))
+    .join('');
+  // Constant-time string comparison to avoid timing oracles.
+  if (actual.length !== expected.length) return false;
+  let diff = 0;
+  for (let i = 0; i < actual.length; i++) {
+    diff |= actual.charCodeAt(i) ^ expected.charCodeAt(i);
+  }
+  return diff === 0;
+}
+
+/** Upsert a rate limit bucket; returns true if request is within the limit. */
+async function checkRateLimit(
+  admin: ReturnType<typeof createClient>,
+  bucketKey: string,
+  limit: number
+): Promise<boolean> {
+  const windowStart = new Date();
+  windowStart.setMinutes(0, 0, 0); // hour-truncated window
+
+  const { data, error } = await admin.rpc('kit_webhook_check_rate_limit', {
+    p_bucket_key: bucketKey,
+    p_window_start: windowStart.toISOString(),
+    p_limit: limit,
+  });
+
+  if (error) {
+    // On DB error, fail open (allow) to avoid blocking legitimate Kit deliveries.
+    console.error('kit-webhook rate limit check failed', error.message);
+    return true;
+  }
+  return data as boolean;
+}
+
+Deno.serve(async req => {
+  if (req.method !== 'POST') {
+    return jsonResponse({ error: 'method_not_allowed' }, 405);
+  }
+
+  if (!SUPABASE_URL || !SUPABASE_SERVICE_ROLE_KEY) {
+    return jsonResponse({ error: 'server_misconfigured' }, 500);
+  }
+
+  // Reject immediately when webhook secret is not configured — prevents silent
+  // pass-through of unauthenticated requests during incomplete deploys. Kit
+  // will retry on 503 once the secret is set.
+  if (!KIT_WEBHOOK_SECRET) {
+    return jsonResponse({ error: 'webhook_not_configured' }, 503);
+  }
+
+  const admin = createClient(SUPABASE_URL, SUPABASE_SERVICE_ROLE_KEY);
+  const ip = clientIp(req);
+
+  // Pre-HMAC rate limit — caps request flood before spending CPU on signature verification.
+  const withinRawLimit = await checkRateLimit(
+    admin,
+    `kit-webhook-raw:${ip}`,
+    RAW_RATE_LIMIT
+  );
+  if (!withinRawLimit) {
+    return jsonResponse({ error: 'rate_limited' }, 429);
+  }
+
+  // Read raw bytes before any parsing — HMAC is computed over the raw body.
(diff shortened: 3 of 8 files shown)


supabase/functions/kit-webhook/index.ts
```diff
@@ -0,0 +1,207 @@
...
+}
+
+/** Upsert a rate limit bucket; returns true if request is within the limit. */
+async function checkRateLimit(
+  admin: ReturnType<typeof createClient>,
+  bucketKey: string,
+  limit: number
+): Promise<boolean> {
+  const windowStart = new Date();
+  windowStart.setMinutes(0, 0, 0); // hour-truncated window
```
ZappoMan: **Minor — hour window computed in Edge runtime timezone**

`setMinutes(0, 0, 0)` uses the Deno host TZ. The RPC already applies `date_trunc('hour', p_window_start)` server-side. Safer to pass `new Date().toISOString()` and let SQL truncate exclusively, or truncate in UTC explicitly — otherwise buckets can split across two hour keys.

supabase/migrations/20260522000001_marketing_email_user_deleted.sql
```diff
@@ -0,0 +1,97 @@
...
+    (v_product_id, 'user.deleted', v_email,
+     'user.deleted:' || OLD.id::text,
+     jsonb_build_object('user_id', OLD.id),
+     'pending')
+  ON CONFLICT (idempotency_key) DO NOTHING;
+
+  RETURN OLD;
+END;
+$$;
+
```
ZappoMan, line 61: **Blocker — missing `REVOKE ALL FROM PUBLIC` on trigger function**

`_marketing_email_on_user_signed_up()` revokes public execute in `20260520200000_marketing_email_v1.sql`. This migration should match:

```sql
REVOKE ALL ON FUNCTION public._marketing_email_on_user_deleted() FROM PUBLIC;
```

Add after the `$$;` and before `CREATE OR REPLACE TRIGGER`. Without it, the SECURITY DEFINER function may be executable by unintended roles.

supabase/functions/kit-sync/index.ts
```diff
@@ -152,7 +152,9 @@ async function handleWorker(
         .eq('email', row.email)
         .maybeSingle();
 
-      if (suppressed) {
+      // user.deleted is a GDPR erasure — it must run even if the user previously
+      // unsubscribed via Kit webhook. Suppression only governs marketing sends.
+      if (suppressed && row.event_type !== 'user.deleted') {
```
ZappoMan, line 157: **Blocker — namespace guard (unchanged lines ~129) blocks GDPR erasure**

This PR correctly bypasses suppression for `user.deleted`, but the **namespace guard above** (unchanged in this diff, ~line 129) still dead-letters all events when `config.namespace` is missing.

`deleteUser()` only needs Kit API key + email — no tags/namespace. Please carve out erasure:

```ts
if (!config.namespace && row.event_type !== 'user.deleted') { ... }
```

Or move the namespace check into tag-bearing `processEvent` cases only.

supabase/functions/kit-webhook/index.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  const { data, error } = await admin.rpc('kit_webhook_check_rate_limit', {
+    p_bucket_key: bucketKey,
+    p_window_start: windowStart.toISOString(),
+    p_limit: limit,
+  });
+
+  if (error) {
+    // On DB error, fail open (allow) to avoid blocking legitimate Kit deliveries.
```
ZappoMan, line 80: **Minor — rate limit fails open on DB errors**

Returning `true` on RPC failure disables rate limiting during DB outages. Reasonable for webhook deliverability; worth documenting. Alternative: fail closed (429) after valid HMAC — Kit retries.

supabase/functions/kit-webhook/index.ts
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  // Reject immediately when webhook secret is not configured — prevents silent
+  // pass-through of unauthenticated requests during incomplete deploys. Kit
+  // will retry on 503 once the secret is set.
+  if (!KIT_WEBHOOK_SECRET) {
+    return jsonResponse({ error: 'webhook_not_configured' }, 503);
+  }
+
+  const admin = createClient(SUPABASE_URL, SUPABASE_SERVICE_ROLE_KEY);
```
ZappoMan: **Nit — match `kit-sync` Supabase client options**

`kit-sync` uses `createClient(..., { auth: { persistSession: false } })`. Worth matching here for consistency.

supabase/functions/kit-webhook/index.ts
```diff
@@ -0,0 +1,207 @@
...
+      string,
+      unknown
+    >;
+  } catch {
+    return jsonResponse({ error: 'invalid_json' }, 400);
+  }
+
+  // Kit v3 uses 'subscriber.unsubscribed'; Kit v4 uses 'subscriber.subscriber_unsubscribe'.
+  // Handle both — first live staging delivery will confirm the name in info logs.
+  const eventType = payload['type'] as string | undefined;
```
ZappoMan, line 153: **Verify on first staging delivery — payload shape**

Good dual-name handling + logging. Kit v4 registration docs use nested `event.name`; deliveries may not use top-level `type`. Confirm field path on first real webhook and extend parser if needed. Consider lowercasing signature hex before compare.

supabase/functions/kit-webhook/index.ts
```diff
@@ -0,0 +1,207 @@
...
+      .order('product_id')
+      .limit(1)
+      .maybeSingle();
+
+    if (!settings) {
+      // Marketing email is disabled — nothing to record, acknowledge to Kit.
+      return jsonResponse({ ok: true }, 200);
+    }
+
+    const { error: insertErr } = await admin
```
ZappoMan, line 185: **Nit — idempotent insert**

Handling `23505` is fine. Optional: `.upsert(..., { onConflict: 'product_id,email', ignoreDuplicates: true })`.

.github/workflows/deploy-staging.yml
```diff
@@ -139,7 +140,8 @@ jobs:
           # strings would clobber any manually-configured Supabase project secrets.
           [[ -n "${STAGING_KIT_API_KEY:-}" ]] && supabase secrets set KIT_API_KEY="${STAGING_KIT_API_KEY}"
           [[ -n "${STAGING_KIT_CRON_SECRET:-}" ]] && supabase secrets set KIT_CRON_SECRET="${STAGING_KIT_CRON_SECRET}"
-          supabase functions deploy stripe-webhook billing-stripe waitlist-capture waitlist-ops kit-sync \
+          [[ -n "${STAGING_KIT_WEBHOOK_SECRET:-}" ]] && supabase secrets set KIT_WEBHOOK_SECRET="${STAGING_KIT_WEBHOOK_SECRET}"
+          supabase functions deploy stripe-webhook billing-stripe waitlist-capture waitlist-ops kit-sync kit-webhook \
```
ZappoMan, line 144: **Deploy wiring looks good** — conditional `KIT_WEBHOOK_SECRET` + `kit-webhook` in deploy list mirrors existing Kit secrets pattern. Note: PR preview still only deploys billing functions (`deploy-preview-billing-functions.sh`), so preview won't exercise this endpoint until that script grows (not a regression from this PR).

supabase/tests/marketing_email_phase4.test.sql
```diff
@@ -0,0 +1,274 @@
+-- pgTAP tests for Phase 4: user.deleted CHECK constraint, AFTER DELETE trigger,
+-- and kit_webhook_check_rate_limit RPC.
+-- Pattern: BEGIN / SELECT plan(N) / assertions / SELECT * FROM finish() / ROLLBACK
+
+BEGIN;
+
+SELECT plan(14);
```
ZappoMan, line 7: **Nice pgTAP coverage** — CHECK, trigger, idempotency, disabled/null-email, rate-limit RPC, payload shape. Follow-up (non-blocking): Tier 2 HTTP integration + Deno signature tests from approved plan — fine to defer if tracked on #286.

ZappoMan (changes requested):
## Phase 4 review (#340)

Solid implementation overall — HMAC validation, dual rate limits, suppression bypass, trigger wiring, and pgTAP coverage match the approved Phase 4 plan. Staging/production deploy wiring looks correct.

**Requesting two fixes before merge** (see inline):
1. `REVOKE ALL` on `_marketing_email_on_user_deleted()` — parity with the signup trigger
2. Skip the namespace guard for `user.deleted` — GDPR `deleteUser` does not need tags/namespace

**Non-blocking follow-ups:** hour-window computation in Edge, fail-open rate limiting, setup docs, integration/Deno tests (acceptable to defer if tracked).

Happy to re-review quickly once the blockers land.

supabase/functions/kit-webhook/index.ts
mei-artificerinnovations: Fixed — removed the `setMinutes(0,0,0)` client-side truncation entirely. Now passing `new Date().toISOString()` raw and letting the RPC's `date_trunc('hour', p_window_start)` handle truncation server-side in UTC.

supabase/migrations/20260522000001_marketing_email_user_deleted.sql
mei-artificerinnovations: Fixed — added `REVOKE ALL ON FUNCTION public._marketing_email_on_user_deleted() FROM PUBLIC;` immediately after the function body, matching the pattern from the v1 migration.

supabase/functions/kit-sync/index.ts
mei-artificerinnovations: Fixed — namespace guard is now `if (!config.namespace && row.event_type !== 'user.deleted')`, matching the suppression bypass. GDPR erasure proceeds regardless of namespace.

supabase/functions/kit-webhook/index.ts
mei-artificerinnovations: Acknowledged — fail-open is intentional for webhook deliverability (Kit retries on 5xx but not on 429). Added a clearer comment in the code explaining the tradeoff.

supabase/functions/kit-webhook/index.ts
mei-artificerinnovations: Fixed — `createClient` now passes `{ auth: { persistSession: false } }` to match kit-sync.

supabase/functions/kit-webhook/index.ts
mei-artificerinnovations: Noted — tracked on #286 for first staging delivery confirmation. Will extend the parser to handle nested `event.name` if Kit v4 deliveries use that shape. Also fixed: `expected` is now lowercased before the constant-time compare.

supabase/functions/kit-webhook/index.ts
mei-artificerinnovations: Keeping the 23505 swallow for now — it's explicit and makes the error path visible. Will consider the upsert pattern in a follow-up if we add more insert sites.

.github/workflows/deploy-staging.yml
mei-artificerinnovations: Agreed — PR preview is out of scope here. Tracked on #286 for when the preview billing functions script is extended.

supabase/tests/marketing_email_phase4.test.sql
mei-artificerinnovations: Tracked on #286 for Phase 5. The pgTAP suite covers the DB layer; Deno sig tests and HTTP integration tests are the logical next layer.

supabase/migrations/20260522000001_marketing_email_user_deleted.sql
```diff
@@ -0,0 +1,99 @@
...
+REVOKE ALL ON FUNCTION public._marketing_email_on_user_deleted() FROM PUBLIC;
+
```
ZappoMan, line 63: ✅ **Confirmed fixed** (`dff776f`) — `REVOKE ALL ON FUNCTION public._marketing_email_on_user_deleted() FROM PUBLIC` matches the signup trigger pattern. Blocker resolved.

supabase/functions/kit-sync/index.ts
```diff
@@ -123,10 +123,10 @@ async function handleWorker(
...
       const config = settings.config;
 
-      // Guard: namespace is required for all Kit tag operations. Dead-letter with a
-      // clear message so the operator knows exactly what to fix, rather than letting
-      // the row reach Kit API and fail with a confusing tag-not-found error.
-      if (!config.namespace) {
+      // Guard: namespace is required for tag operations, but not for GDPR erasure
+      // (deleteUser only needs API key + email). Carve out user.deleted so it is
+      // never dead-lettered by a missing namespace.
+      if (!config.namespace && row.event_type !== 'user.deleted') {
```
ZappoMan, line 129: ✅ **Confirmed fixed** (`dff776f`) — namespace guard now carves out `user.deleted` so GDPR `deleteUser` is not dead-lettered when `config.namespace` is missing. Blocker resolved.

supabase/functions/kit-webhook/index.ts
```diff
@@ -0,0 +1,208 @@
...
+}
+
+async function verifyKitSignature(
+  body: ArrayBuffer,
+  sigHeader: string | null,
+  secret: string
+): Promise<boolean> {
+  if (!sigHeader?.startsWith('sha256=')) return false;
+  const expected = sigHeader.slice(7).toLowerCase();
+  const key = await crypto.subtle.importKey(
```
ZappoMan, line 45: ✅ **Confirmed fixed** (`dff776f`) — lowercasing expected signature hex before constant-time compare. Good hardening beyond the original review note.

supabase/functions/kit-webhook/index.ts
```diff
@@ -0,0 +1,208 @@
...
+  admin: ReturnType<typeof createClient>,
+  bucketKey: string,
+  limit: number
+): Promise<boolean> {
+  // Pass raw timestamp — the RPC applies date_trunc('hour', ...) server-side in UTC,
+  // avoiding any TZ skew from the Edge runtime's local clock.
+  const { data, error } = await admin.rpc('kit_webhook_check_rate_limit', {
+    p_bucket_key: bucketKey,
+    p_window_start: new Date().toISOString(),
+    p_limit: limit,
```
ZappoMan, line 76: ✅ **Confirmed fixed** (`dff776f`) — passes raw `toISOString()` to the RPC and lets SQL `date_trunc('hour', ...)` handle UTC truncation. Avoids Edge runtime TZ skew.

supabase/functions/kit-webhook/index.ts
```diff
@@ -0,0 +1,208 @@
...
+  // Reject immediately when webhook secret is not configured — prevents silent
+  // pass-through of unauthenticated requests during incomplete deploys. Kit
+  // will retry on 503 once the secret is set.
+  if (!KIT_WEBHOOK_SECRET) {
+    return jsonResponse({ error: 'webhook_not_configured' }, 503);
+  }
+
+  const admin = createClient(SUPABASE_URL, SUPABASE_SERVICE_ROLE_KEY, {
+    auth: { persistSession: false },
+  });
```
ZappoMan, line 105: ✅ **Confirmed fixed** (`dff776f`) — `{ auth: { persistSession: false } }` now matches `kit-sync`.

ZappoMan (approved):
## Re-review — blockers resolved ✅

Both requested fixes landed in `dff776f`. CI green (`db-tests`, `supabase-test-report`, `integration`).

**Still open (non-blocking — see report in PR thread):**
- Rate-limit fail-open on DB RPC error (acceptable tradeoff)
- Setup docs paragraph for Kit webhook URL/events
- Tier 2 HTTP + Deno signature tests (defer OK if tracked on #286)
- PR-preview deploy of `kit-webhook` (same gap as waitlist/kit-sync today)
- Live Kit payload field verification (`type` vs nested `event.name`) — confirm on first staging delivery

Approve from my side once the above deferrals are acceptable to the team.

supabase/migrations/20260522000001_marketing_email_user_deleted.sql
```diff
@@ -0,0 +1,97 @@
...
+     'pending')
+  ON CONFLICT (idempotency_key) DO NOTHING;
+
+  RETURN OLD;
+END;
+$$;
+
+CREATE OR REPLACE TRIGGER marketing_email_on_user_deleted
+  AFTER DELETE ON auth.users
+  FOR EACH ROW EXECUTE FUNCTION public._marketing_email_on_user_deleted();
```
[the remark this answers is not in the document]
ZappoMan: **Status: partially fixed / not blocking**

- **REVOKE:** ✅ Fixed in `dff776f` — `REVOKE ALL ON FUNCTION public._marketing_email_on_user_deleted() FROM PUBLIC;` is now present (line 63), matching the signup trigger pattern.
- **`CREATE OR REPLACE TRIGGER`:** ⏳ Left as-is — valid on PG14+ (Supabase PG15). pgTAP/CI passes with this migration. Repo’s first-install triggers elsewhere use plain `CREATE TRIGGER`; if we want idempotent re-apply semantics we can switch to `DROP TRIGGER IF EXISTS …; CREATE TRIGGER …` in a follow-up, but not required for merge.

supabase/migrations/20260522000001_marketing_email_user_deleted.sql
```diff
@@ -0,0 +1,97 @@
...
+  p_bucket_key text,
+  p_window_start timestamptz,
+  p_limit integer
+)
+RETURNS boolean
+LANGUAGE plpgsql
+VOLATILE
+SECURITY DEFINER
+SET search_path = public
+AS $$
```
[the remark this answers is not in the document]
ZappoMan: **Status: pending follow-up (#286)**

Still `SET search_path = public` without `pg_catalog`. Valid hardening — other marketing-email SECURITY DEFINER functions use `public, pg_catalog`. Recommend a one-line fix before or immediately after merge:

```sql
SET search_path = public, pg_catalog
```

Tracked on [#286 Phase 4 completion report](https://github.com/Artificer-Innovations/BeakerStack/issues/286#issuecomment-4522775097).

supabase/migrations/20260522000001_marketing_email_user_deleted.sql
```diff
@@ -0,0 +1,97 @@
...
+DECLARE
+  v_count integer;
+BEGIN
+  INSERT INTO public.waitlist_rate_limits (bucket_key, window_start, count)
+  VALUES (p_bucket_key, date_trunc('hour', p_window_start), 1)
+  ON CONFLICT (bucket_key, window_start)
+  DO UPDATE SET count = public.waitlist_rate_limits.count + 1
+  RETURNING count INTO v_count;
```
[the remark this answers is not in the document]
ZappoMan: **Status: pending follow-up (#286)**

Agree — `clientIp()` currently passes the raw header value into `kit-webhook-raw:${ip}` / `kit-webhook:${ip}` with no length cap. Waitlist RPC truncates with `left(trim(p_client_ip), 64)`.

Recommend mirroring that defensively in either:
- `kit-webhook/index.ts` (truncate/normalize IP before building bucket key), **or**
- `kit_webhook_check_rate_limit` (`left(trim(p_bucket_key), …)` before insert)

Low priority for v1 (HMAC + rate limit still apply), but worth a fast follow-up to prevent bucket-key bloat. Tracked on #286.

supabase/tests/marketing_email_phase4.test.sql
```diff
@@ -0,0 +1,274 @@
...
+
+-- ── Test fixtures ────────────────────────────────────────────────────────────
+
+-- Enable marketing email for one product so trigger fires.
+INSERT INTO public.marketing_email_settings
+  (product_id, enabled, provider, config)
+VALUES
+  ('beakerstack', true, 'kit', '{"namespace":"bstack-test","form_id":"form-1"}')
+ON CONFLICT (product_id) DO UPDATE
+  SET enabled = true, provider = 'kit', config = '{"namespace":"bstack-test","form_id":"form-1"}'::jsonb;
```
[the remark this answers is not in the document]
ZappoMan: **Status: pending follow-up (nit)**

Still uses `form_id` in the pgTAP fixture; production config shape uses `kitFormId` (see `kit-sync` / integration tests). This test only exercises the DELETE trigger (doesn’t read form config), so behavior is unaffected.

Recommend renaming to `kitFormId` in the fixture for copy-paste consistency — optional cleanup in this PR or a tiny follow-up.

PR #361, opened 2026-05-23
Add Kit setup automation and shared CI deploy wiring

ZappoMan:
## Summary
- Add `npm run setup:kit` and a `kit` phase in `setup:full` to collect shared `KIT_*` secrets, merge them into local env files, and optionally sync to GitHub.
- Auto-provision Kit unsubscribe webhooks via `ensure-kit-webhook-endpoint.mjs` when Supabase URLs are available (setup + CI).
- Wire shared Kit secrets through staging, production, and PR preview edge deploys, including waitlist + Kit functions and `kit-sync` cron registration in preview.

## Test plan
- [x] `node --test scripts/__tests__/setup-kit.test.mjs scripts/__tests__/setup-kit-script.test.mjs scripts/__tests__/ensure-kit-webhook.test.mjs scripts/__tests__/setup-manifest-github-sync.test.mjs`
- [x] `npm run docs:actions-secrets`
- [ ] Run `npm run setup:kit -- --dry-run` locally to verify prompts/guidance
- [ ] After merging: set `KIT_*` GitHub secrets and confirm staging/preview deploy steps register webhooks + cron


mei-artificerinnovations (changes requested):
## PR #361 review — Kit setup automation and shared CI deploy

The architecture is solid. Three concerns before merging.

---

### Blocking

#### 1. `normalizeKitWebhookUrl` doesn't collapse mid-path double slashes

In all three deploy workflows, `KIT_WEBHOOK_URL` is built by GitHub Actions expression concatenation:

```yaml
KIT_WEBHOOK_URL: ${{ secrets.PRODUCTION_SUPABASE_URL }}/functions/v1/kit-webhook
```

If the GitHub secret has a trailing slash (`https://proj.supabase.co/`), the resulting URL is `https://proj.supabase.co//functions/v1/kit-webhook`. `normalizeKitWebhookUrl` only strips trailing slashes:

```js
return String(url || '').trim().replace(/\/+$/, '');
```

So the double slash in the path survives. Kit's API would register a webhook with a malformed URL. When the same step runs again, `findKitWebhookByUrl` normalizes both sides the same way — so the double-slash URL would be found and matched — but the registered URL in Kit's dashboard would look wrong and could break Kit's delivery.

The library already has `kitWebhookUrlFromSupabaseUrl` that builds a clean URL from a Supabase URL. The fix is to have `ensure-kit-webhook-endpoint.mjs` accept `SUPABASE_URL` (and derive the webhook path) rather than `KIT_WEBHOOK_URL` (the pre-built URL). Alternatively, update `normalizeKitWebhookUrl` to collapse consecutive slashes in the path:

```js
return String(url || '').trim().replace(/\/+$/, '').replace(/([^:])\/\/+/g, '$1/');
```

Either fix removes the latent dependency on the secret's format.

#### 2. `ensure-kit-webhook-endpoint.mjs` — `KIT_WEBHOOK_URL` and `kitWebhookUrlFromSupabaseUrl` diverge

Related to #1: `ensure-kit-webhook-endpoint.mjs` accepts `KIT_WEBHOOK_URL` directly from the environment, but `setup-kit.mjs` / `setup-kit.mjs lib` always derive the webhook URL from `SUPABASE_URL` via `kitWebhookUrlFromSupabaseUrl`. This means the CI path and the setup path use different construction logic. The CI path could register a URL that the setup path would not match (e.g., if the URL is constructed differently across environments).

Simplest fix: have the CI entry point accept `KIT_SUPABASE_URL` instead of `KIT_WEBHOOK_URL`, derive internally with `kitWebhookUrlFromSupabaseUrl`, and remove the `KIT_WEBHOOK_URL` env variable from all three workflow steps. This aligns the CI and setup paths.

---

### Non-blocking

#### 3. `setup-kit-script.test.mjs` — hardcoded string literal instead of constant

```js
assert.equal(acc.SETUP_KIT_SKIPPED, 'true');
```

Should be:

```js
import { SETUP_KIT_SKIPPED_ENV } from '../lib/setup-kit.mjs';
// ...
assert.equal(acc[SETUP_KIT_SKIPPED_ENV], 'true');
```

Functionally correct today since `SETUP_KIT_SKIPPED_ENV === 'SETUP_KIT_SKIPPED'`, but if the constant is renamed the test passes silently with the wrong assertion.

#### 4. `collectKitEnvKeys` makes live Kit API calls on re-run when credentials are present

```js
if (hasKitCredentials(acc)) {
  // ...
  await ensureKitWebhooksForTiers(acc, { dryRun, logInfo, logWarn });
  return;
}
```

Every time `setup:full` runs with valid credentials, this path calls the Kit API to check webhook registration. It's idempotent and `--dry-run` suppresses it, but it will silently make external network calls during re-runs. The log line just says "Kit credentials already present" with no indication that an API call is about to happen. Adding a log line before the call (`logInfo('Verifying Kit webhook registration...')`) would remove the surprise.

#### 5. `run_stripe_billing` scope output now gates non-billing functions

The "Deploy preview edge functions" step now triggers on `run_stripe_billing == 'true' || run_supabase_reset == 'true'`. Correct behavior — but the scope output name `run_stripe_billing` now also controls whether waitlist and Kit functions are deployed. Future readers of the scope logic may be confused about why a billing flag governs Kit deploy. Worth noting in a comment or renaming the output to `run_edge_functions` if that's feasible without a larger refactor.

---

### What's done well

- The `ensureKitWebhook` / `findKitWebhookByUrl` / `listKitWebhooks` with cursor pagination is clean and handles the idempotency correctly.
- Test coverage is thorough — 174 lines covering pagination, URL matching, create vs. find, and all the library helpers.
- `continue-on-error: true` on all three webhook ensure steps is correct — Kit registration is never deploy-critical.
- The `--yes` / `--dry-run` / `--skip-kit` matrix is complete and consistent across standalone and `setup:full` use.
- The `RESOLVED_KIT_*` fallback pattern (shared first, legacy `STAGING_KIT_*` / `PRODUCTION_KIT_*` second) in staging/production deploys is a clean migration path.

.github/workflows/pr-preview-environment.yml
```diff
@@ -279,12 +294,28 @@ jobs:
...
           chmod +x ./scripts/pr-preview/deploy-preview-billing-functions.sh
           ./scripts/pr-preview/deploy-preview-billing-functions.sh
 
+      - name: Ensure kit-sync cron job (preview)
+        if: >-
+          needs.preview-scope.outputs.run_stripe_billing == 'true' ||
+          needs.preview-scope.outputs.run_supabase_reset == 'true'
+        continue-on-error: true
+        env:
+          SUPABASE_URL: ${{ secrets.PREVIEW_SUPABASE_URL }}
```
diego-artificerinnovations, line 312: Preview cron is missing the legacy `PREVIEW_KIT_CRON_SECRET` fallback that staging and production both have. Staging:

```yaml
KIT_CRON_SECRET: ${{ secrets.KIT_CRON_SECRET }}
STAGING_KIT_CRON_SECRET: ${{ secrets.STAGING_KIT_CRON_SECRET }}
```
```bash
RESOLVED_KIT_CRON_SECRET="${KIT_CRON_SECRET:-${STAGING_KIT_CRON_SECRET:-}}"
[[ -z "${RESOLVED_KIT_CRON_SECRET}" ]] && { echo "KIT_CRON_SECRET not set — skipping"; exit 0; }
KIT_CRON_SECRET="${RESOLVED_KIT_CRON_SECRET}" node scripts/ensure-kit-sync-cron.mjs
```

`deploy-preview-billing-functions.sh` already resolves the fallback when pushing secrets into Supabase, so the secret gets set correctly — but then the cron step skips because it never sees `PREVIEW_KIT_CRON_SECRET`. Fix matches the staging/production pattern.

scripts/setup-kit.mjs
```diff
@@ -0,0 +1,311 @@
...
+    ghCtx,
+    { dryRun, interactive: true, githubRepoOverride: Boolean(repoOverride) }
+  );
+  if (!proceed) {
+    logInfo('Skipped GitHub Kit secret sync.');
+    return;
+  }
+
+  for (const [name, value] of Object.entries(secrets)) {
+    ghSecretSetSync(repo, name, value, dryRun);
```
diego-artificerinnovations: `ghSecretSetSync` return value is discarded. On failure it emits a warning, but `logInfo` here runs unconditionally — user sees "GitHub secret set: KIT_API_KEY" right after the failure warning.

```js
const status = ghSecretSetSync(repo, name, value, dryRun);
if (status === 0) {
  logInfo(dryRun ? `[dry-run] would set secret ${name}` : `GitHub secret set: ${name}`);
}
```

diego-artificerinnovations (changes requested):
## Review

The overall design is clean. The Kit phase follows the exact same pattern as the Stripe phase (`phaseKit` / `phaseStripe`, `SETUP_KIT_SKIPPED_ENV` / `SETUP_STRIPE_SKIPPED_ENV`), the `ensureKitWebhook` abstraction is well-layered, the pagination loop is correct, and the fallback to legacy `STAGING_KIT_*` / `PRODUCTION_KIT_*` secrets is handled properly in staging and production. One blocking issue in the preview workflow; a couple of non-blocking things below.

---

### Blocking

**Preview cron missing legacy secret fallback — `pr-preview-environment.yml`**

See inline comment. The "Ensure kit-sync cron job (preview)" step reads only `${{ secrets.KIT_CRON_SECRET }}` with no fallback to `PREVIEW_KIT_CRON_SECRET`. Every parallel step — staging cron, production cron, and `deploy-preview-billing-functions.sh` — handles the `KIT_CRON_SECRET ?: TIER_KIT_CRON_SECRET` fallback. Preview cron is the odd one out.

An adopter who already has `PREVIEW_KIT_CRON_SECRET` set (but not the new shared `KIT_CRON_SECRET`) would get the Kit secret pushed into Supabase correctly (the shell script handles it), but the cron job would not be registered — silently skipped.

---

### Non-blocking

**`syncKitGithubSecrets` logs false positive on `gh` failure — `scripts/setup-kit.mjs`**

See inline comment. `ghSecretSetSync` returns a non-zero status and emits a warning on failure, but the caller discards the return value and unconditionally logs `"GitHub secret set: ${name}"`. On a `gh` auth failure or network error, the user sees a warning and then immediately "GitHub secret set: KIT_API_KEY" — confusing.

Fix:
```js
const status = ghSecretSetSync(repo, name, value, dryRun);
if (status === 0) {
  logInfo(dryRun ? `[dry-run] would set secret ${name}` : `GitHub secret set: ${name}`);
}
```

**Missing test: `collectKitEnvKeys` webhook-secret blank rollback**

The path in `lib/setup-kit.mjs` where `apiKey` is entered but `webhookSecret` is left blank deletes `acc.KIT_API_KEY` and `acc.KIT_CRON_SECRET` (the rollback on lines 465–472). This is the most surprising behavior in the module and has no test. Worth adding — e.g., secretAnswers of `['api-key', '', '']` should result in no Kit keys in `acc`.

**Style nit: duplicate import in `setup-kit-script.test.mjs`**

```js
import { pickKitGithubPayload } from '../setup-kit.mjs';
import { phaseKit } from '../setup-kit.mjs';
```
Merge into one import statement.

scripts/lib/setup-kit.mjs
```diff
@@ -0,0 +1,253 @@
...
+      );
+  const webhookSecret =
+    webhookSecretRaw.trim() || String(acc.KIT_WEBHOOK_SECRET ?? '').trim();
+  if (!webhookSecret) {
+    logWarn(
+      'No webhook signing secret entered — inbound Kit unsubscribes will not verify until KIT_WEBHOOK_SECRET is set.'
+    );
+    delete acc.KIT_API_KEY;
+    delete acc.KIT_CRON_SECRET;
+    return;
```
rahul-artificerinnovations, line 302: **Blocking: partial abort leaves no `SETUP_KIT_SKIPPED_ENV` signal.**

When the user provides API key + cron secret but no webhook secret, the function deletes both keys from `acc` and returns — but never sets `acc[SETUP_KIT_SKIPPED_ENV] = 'true'`. Downstream, the GitHub phase sees Kit keys missing from `acc` and from GitHub, emits the "1 Kit key(s) still missing" warning, and may prompt the user during `collectMissingGithubCiEnvIntoAcc` — even though the user deliberately skipped at this step.

```js
if (!webhookSecret) {
  logWarn(
    'No webhook signing secret entered — inbound Kit unsubscribes will not verify until KIT_WEBHOOK_SECRET is set.'
  );
  delete acc.KIT_API_KEY;
  delete acc.KIT_CRON_SECRET;
  acc[SETUP_KIT_SKIPPED_ENV] = 'true'; // add this line
  return;
}
```

The `SETUP_KIT_SKIPPED_ENV` constant is already in scope (imported at the top of `scripts/lib/setup-kit.mjs`).

scripts/__tests__/setup-kit-script.test.mjs
```diff
@@ -0,0 +1,29 @@
...
+    KIT_API_KEY: 'k',
+    KIT_CRON_SECRET: 'c',
+    KIT_WEBHOOK_SECRET: 'w',
+  });
+});
+
+test('phaseKit with skipKit sets SETUP_KIT_SKIPPED', async () => {
+  const acc = {};
+  const rl = { question: async () => '', close() {} };
+  await phaseKit({ skipKit: true }, rl, acc);
```
rahul-artificerinnovations, line 27: **Non-blocking: string literal instead of imported constant.**

```js
assert.equal(acc.SETUP_KIT_SKIPPED, 'true');
```

`SETUP_KIT_SKIPPED_ENV` is not imported in this file, so the test hard-codes the current constant value (`'SETUP_KIT_SKIPPED'`). If the constant is ever renamed, this test silently continues to pass while the production code breaks. Use the constant:

```js
import { SETUP_KIT_SKIPPED_ENV } from '../lib/setup-kit.mjs';
// ...
assert.equal(acc[SETUP_KIT_SKIPPED_ENV], 'true');
```

Note also: the two `import { ... } from '../setup-kit.mjs'` lines at the top of this file can be merged into one.

rahul-artificerinnovations (changes requested):
The overall architecture is solid: `ensureKitWebhook` is properly idempotent, the paginated list-before-create pattern is correct, `kitWebhookUrlFromSupabaseUrl` uses `new URL()` rather than string hacks, and the `RESOLVED_KIT_*` fallback pattern in CI is clean. Two blocking issues before merge.

## Blocking

**1. CI webhook URL construction can produce double-slash, bypassing idempotency**

All three workflows build `KIT_WEBHOOK_URL` by concatenating a secret with a path:

```yaml
# deploy-production.yml, deploy-staging.yml, pr-preview-environment.yml
KIT_WEBHOOK_URL: ${{ secrets.PRODUCTION_SUPABASE_URL }}/functions/v1/kit-webhook
```

If the stored secret has a trailing slash — a common entry error — the URL becomes `https://proj.supabase.co//functions/v1/kit-webhook`. `normalizeKitWebhookUrl` only strips *trailing* slashes (`.replace(/\/+$/, '')`); it does not collapse internal `//`. So `findKitWebhookByUrl` normalizes the input to `https://proj.supabase.co//functions/v1/kit-webhook`, compares against Kit-stored `https://proj.supabase.co/functions/v1/kit-webhook`, finds no match, and `ensureKitWebhook` creates a new webhook on every deploy — registering a permanently broken URL.

The library already has the right tool: `kitWebhookUrlFromSupabaseUrl` parses via `new URL()` and reconstructs a clean path, immune to trailing-slash secrets. Two fixes are acceptable:

**Option A** — Pass `SUPABASE_URL` (the raw project URL) as a separate env var and derive the webhook URL inside `ensure-kit-webhook-endpoint.mjs`:
```js
const url = process.env.KIT_WEBHOOK_URL
  || kitWebhookUrlFromSupabaseUrl(process.env.SUPABASE_URL || '');
```
Then in CI just pass `SUPABASE_URL: ${{ secrets.PRODUCTION_SUPABASE_URL }}` and drop the manual concatenation.

**Option B** — Extend `normalizeKitWebhookUrl` to also collapse internal double-slashes after the scheme:
```js
export function normalizeKitWebhookUrl(url) {
  return String(url || '')
    .trim()
    .replace(/([^:])(\/\/+)/g, '$1/')
    .replace(/\/+$/, '');
}
```
Either option also fixes the same latent issue in `deploy-preview-billing-functions.sh`, which constructs `KIT_WEBHOOK_URL` with the same pattern.

**2. `collectKitEnvKeys` partial abort leaves no `SETUP_KIT_SKIPPED_ENV` signal** — see inline comment on `scripts/lib/setup-kit.mjs`.

## Non-blocking

**3. `setup-kit-script.test.mjs` uses string literal instead of imported constant** — see inline comment.

**4. Double guidance in interactive `setup-full` flow**

In `setup-full.mjs`, the kit phase first calls `printKitPhaseReadinessBriefing(logCtx)` followed by a "Press Enter..." pause, then dispatches to `phaseKit` → `collectKitEnvKeys`, which calls `printKitSetupGuidance(logInfo)` again (line 416 of `scripts/lib/setup-kit.mjs`). The user sees two successive blocks of Kit setup instructions with an Enter prompt between them. Consider: either suppress the inner `printKitSetupGuidance` call when invoked from `setup-full` (via a `skipGuidance` opt), or fold the briefing into `collectKitEnvKeys` and remove `printKitPhaseReadinessBriefing` from the `setup-full` dispatch site.

scripts/lib/setup-kit.mjs
rahul-artificerinnovations, line 302: **Blocking: partial abort leaves no `SETUP_KIT_SKIPPED_ENV` signal.**

When the user provides API key + cron secret but no webhook secret, the function deletes both keys from `acc` and returns — but never sets `acc[SETUP_KIT_SKIPPED_ENV] = 'true'`. Downstream, the GitHub phase sees Kit keys missing from `acc` and from GitHub, emits the "1 Kit key(s) still missing" warning, and may prompt the user during `collectMissingGithubCiEnvIntoAcc` — even though the user deliberately skipped at this step.

```js
if (!webhookSecret) {
  logWarn(
    'No webhook signing secret entered — inbound Kit unsubscribes will not verify until KIT_WEBHOOK_SECRET is set.'
  );
  delete acc.KIT_API_KEY;
  delete acc.KIT_CRON_SECRET;
  acc[SETUP_KIT_SKIPPED_ENV] = 'true'; // add this line
  return;
}
```

The `SETUP_KIT_SKIPPED_ENV` constant is already in scope (imported at the top of `scripts/lib/setup-kit.mjs`).

scripts/__tests__/setup-kit-script.test.mjs
rahul-artificerinnovations, line 27: **Non-blocking: string literal instead of imported constant.**

```js
assert.equal(acc.SETUP_KIT_SKIPPED, 'true');
```

`SETUP_KIT_SKIPPED_ENV` is not imported in this file, so the test hard-codes the current constant value (`'SETUP_KIT_SKIPPED'`). If the constant is ever renamed, this test silently continues to pass while the production code breaks. Use the constant:

```js
import { SETUP_KIT_SKIPPED_ENV } from '../lib/setup-kit.mjs';
// ...
assert.equal(acc[SETUP_KIT_SKIPPED_ENV], 'true');
```

Note also: the two `import { ... } from '../setup-kit.mjs'` lines at the top of this file can be merged into one.

rahul-artificerinnovations (changes requested):
The overall architecture is solid: `ensureKitWebhook` is properly idempotent, the paginated list-before-create pattern is correct, `kitWebhookUrlFromSupabaseUrl` uses `new URL()` rather than string hacks, and the `RESOLVED_KIT_*` fallback pattern in CI is clean. Two blocking issues before merge.

## Blocking

**1. CI webhook URL construction can produce double-slash, bypassing idempotency**

All three workflows build `KIT_WEBHOOK_URL` by concatenating a secret with a path:

```yaml
# deploy-production.yml, deploy-staging.yml, pr-preview-environment.yml
KIT_WEBHOOK_URL: ${{ secrets.PRODUCTION_SUPABASE_URL }}/functions/v1/kit-webhook
```

If the stored secret has a trailing slash — a common entry error — the URL becomes `https://proj.supabase.co//functions/v1/kit-webhook`. `normalizeKitWebhookUrl` only strips *trailing* slashes (`.replace(/\/+$/, '')`); it does not collapse internal `//`. So `findKitWebhookByUrl` normalizes the input to `https://proj.supabase.co//functions/v1/kit-webhook`, compares against Kit-stored `https://proj.supabase.co/functions/v1/kit-webhook`, finds no match, and `ensureKitWebhook` creates a new webhook on every deploy — registering a permanently broken URL.

The library already has the right tool: `kitWebhookUrlFromSupabaseUrl` parses via `new URL()` and reconstructs a clean path, immune to trailing-slash secrets. Two fixes are acceptable:

**Option A** — Pass `SUPABASE_URL` (the raw project URL) as a separate env var and derive the webhook URL inside `ensure-kit-webhook-endpoint.mjs`:
```js
const url = process.env.KIT_WEBHOOK_URL
  || kitWebhookUrlFromSupabaseUrl(process.env.SUPABASE_URL || '');
```
Then in CI just pass `SUPABASE_URL: ${{ secrets.PRODUCTION_SUPABASE_URL }}` and drop the manual concatenation.

**Option B** — Extend `normalizeKitWebhookUrl` to also collapse internal double-slashes after the scheme:
```js
export function normalizeKitWebhookUrl(url) {
  return String(url || '')
    .trim()
    .replace(/([^:])(\/\/+)/g, '$1/')
    .replace(/\/+$/, '');
}
```
Either option also fixes the same latent issue in `deploy-preview-billing-functions.sh`, which constructs `KIT_WEBHOOK_URL` with the same pattern.

**2. `collectKitEnvKeys` partial abort leaves no `SETUP_KIT_SKIPPED_ENV` signal** — see inline comment on `scripts/lib/setup-kit.mjs`.

## Non-blocking

**3. `setup-kit-script.test.mjs` uses string literal instead of imported constant** — see inline comment.

**4. Double guidance in interactive `setup-full` flow**

In `setup-full.mjs`, the kit phase first calls `printKitPhaseReadinessBriefing(logCtx)` followed by a "Press Enter..." pause, then dispatches to `phaseKit` → `collectKitEnvKeys`, which calls `printKitSetupGuidance(logInfo)` again (line 416 of `scripts/lib/setup-kit.mjs`). The user sees two successive blocks of Kit setup instructions with an Enter prompt between them. Consider: either suppress the inner `printKitSetupGuidance` call when invoked from `setup-full` (via a `skipGuidance` opt), or fold the briefing into `collectKitEnvKeys` and remove `printKitPhaseReadinessBriefing` from the `setup-full` dispatch site.

sarah-artificerinnovations:
## Code review — PR #361 (ZappoMan)

Solid overall structure. The lib/script separation is clean: `ensure-kit-webhook.mjs` is a well-factored pure API client, `setup-kit.mjs` / `lib/setup-kit.mjs` follow the Stripe phase pattern correctly, and the CI wiring (shared `KIT_*` with per-env fallback) is the right model. The test coverage is thorough. One data-loss bug needs fixing before merge.

---

### 🔴 One blocker

**`collectKitEnvKeys` deletes pre-existing credentials when the webhook secret prompt is skipped**

When `hasKitCredentials(acc)` is false (e.g. the user has `KIT_API_KEY` in env but no `KIT_WEBHOOK_SECRET` yet), the function enters the prompt path and writes `acc.KIT_API_KEY = apiKey`. If the user then skips or cancels at the webhook secret prompt, the function does:

```js
if (!webhookSecret) {
  logWarn('No webhook signing secret entered...');
  delete acc.KIT_API_KEY;     // ← deletes a key the user may have had before
  delete acc.KIT_CRON_SECRET; // ← deletes an auto-generated key
  return;
}
```

This destroys `KIT_API_KEY` from `acc` — including the value that was already in the user's env file before the function ran. When the write phase later flushes `acc` to disk, the previously valid API key is gone. The auto-generated `KIT_CRON_SECRET` is also deleted even though it was freshly minted by this call.

The common user scenario: developer has their Kit API key but hasn't created the webhook signing secret yet. They run `setup:kit`, hit the webhook prompt, choose to skip for now — and lose their API key from the env file. Recoverable (they can re-enter it), but surprising and incorrect behavior.

**Fix:** track which keys were *newly set* by this call and only delete those on rollback, not keys that already existed:

```js
const wasPresent = { KIT_API_KEY: Boolean(acc.KIT_API_KEY?.trim()), KIT_CRON_SECRET: Boolean(acc.KIT_CRON_SECRET?.trim()) };

// ... set acc.KIT_API_KEY, acc.KIT_CRON_SECRET ...

if (!webhookSecret) {
  if (!wasPresent.KIT_API_KEY) delete acc.KIT_API_KEY;
  if (!wasPresent.KIT_CRON_SECRET) delete acc.KIT_CRON_SECRET;
  return;
}
```

Or more simply: never delete. Just return without writing `KIT_WEBHOOK_SECRET` — the caller already guards writes with `if (v) out[key] = v` in `pickKitGithubPayload`, so partial credentials won't reach GitHub.

---

### 🟡 Three design gaps

**1. `KIT_WEBHOOK_URL` string concatenation is fragile**

In all three workflow files:

```yaml
KIT_WEBHOOK_URL: ${{ secrets.STAGING_SUPABASE_URL }}/functions/v1/kit-webhook
```

If the stored secret has a trailing slash (not uncommon — Supabase dashboard copies URLs without, but humans add them), the resulting URL is `https://abc.supabase.co//functions/v1/kit-webhook`. `normalizeKitWebhookUrl` only strips trailing slashes, not mid-path double slashes, so `findKitWebhookByUrl` won't match a webhook registered with the clean URL, causing spurious re-creates.

Fix in `ensure-kit-webhook-endpoint.mjs`: pass `KIT_WEBHOOK_URL` through `kitWebhookUrlFromSupabaseUrl` (which handles both raw project URLs and custom base-path URLs) rather than trusting the env var to be clean:

```js
import { kitWebhookUrlFromSupabaseUrl } from './lib/ensure-kit-webhook.mjs';

const rawUrl = process.env.KIT_WEBHOOK_URL;
const url = rawUrl.includes('/functions/v1/kit-webhook')
  ? rawUrl          // already a full webhook URL
  : kitWebhookUrlFromSupabaseUrl(rawUrl); // derive from Supabase URL
```

Or simpler: just normalize mid-path slashes in `normalizeKitWebhookUrl`:

```js
return String(url || '').trim().replace(/([^:])//+/g, '$1/').replace(//+$/, '');
```

**2. `kitRequest` has no 429 retry**

`ensureKitWebhooksForTiers` calls the Kit API three times in sequence (Preview → Staging → Production). If a deployment fires after a recent setup run, the list-then-create flow doubles the API calls (one GET + possible POST per tier). Kit's API is rate-limited; a 429 today throws immediately and the CI step silently passes (via `continue-on-error: true`) with the webhook missing.

Add a retry with exponential backoff on 429:

```js
const retryAfter = res.headers.get('Retry-After');
const delay = retryAfter ? Number(retryAfter) * 1000 : 2000;
// wait and retry once
```

Or at minimum, log the 429 clearly so an operator knows to re-run the step rather than wondering why webhook delivery is failing.

**3. `stringifyDotEnv` in `setup-kit.mjs` produces different output than `setup-full.mjs`**

`setup-kit.mjs` defines its own:

```js
function stringifyDotEnv(obj) {
  return `${Object.entries(obj).map(([k, v]) => `${k}="${escapeDotEnvDoubleQuotedValue(v)}"`).join('\n')}\n`;
}
```

`setup-full.mjs` has a different version: adds a `# Generated by` header comment, sorts keys alphabetically, and quotes selectively based on whether the value contains spaces or `#`. If a user runs `setup:kit` standalone after `setup-full`, `.env.local` gets rewritten without the header, without sorting, and with all values double-quoted — surprising output for a cosmetic no-op.

Extract the serializer to a shared helper in `scripts/lib/setup-dotenv.mjs` (it already has `parseDotEnv` and `escapeDotEnvDoubleQuotedValue`) so both scripts produce consistent output.

---

### 🔵 Minor

**Double import in `setup-kit-script.test.mjs`:**

```js
import { pickKitGithubPayload } from '../setup-kit.mjs';
import { phaseKit } from '../setup-kit.mjs';
```

Merge into one import.

**`which()` is duplicated.** Both `setup-kit.mjs` and `setup-full.mjs` implement identical `which()` helpers. Extract to `scripts/lib/setup-shell.mjs` or wherever other shared CLI utilities live.

**`listKitWebhooks` is called once per tier.** In a repo with all three tiers configured, `ensureKitWebhooksForTiers` fires three separate paginated list fetches. If the Kit account has many webhooks across pages, this amplifies API calls 3×. Consider fetching the full list once and passing it to a `findWebhookInList(list, url)` helper.

scripts/lib/ensure-kit-webhook.mjs
```diff
@@ -0,0 +1,167 @@
...
+  'subscriber.subscriber_unsubscribe';
+
+/**
+ * @param {string} url
+ * @returns {string}
+ */
+export function normalizeKitWebhookUrl(url) {
+  return String(url || '')
+    .trim()
+    .replace(/\/+$/, '');
```
mei-artificerinnovations: This strips trailing slashes but not mid-path double slashes. In all three CI workflow steps, `KIT_WEBHOOK_URL` is built by YAML expression concatenation:

```yaml
KIT_WEBHOOK_URL: ${{ secrets.PRODUCTION_SUPABASE_URL }}/functions/v1/kit-webhook
```

If the secret has a trailing slash, the evaluated string is `https://proj.supabase.co//functions/v1/kit-webhook`. That double slash survives this normalize call and gets registered with Kit's API. Kit stores the malformed URL in its webhook list, and depending on Kit's URL matching, delivery can break.

**Fix option A** — add a path-segment collapse here:

```js
return String(url || '')
  .trim()
  .replace(/([^:])//+/g, '$1/') // collapse double slashes outside protocol
  .replace(//+$/, '');
```

**Fix option B** — have `ensure-kit-webhook-endpoint.mjs` accept `KIT_SUPABASE_URL` and call `kitWebhookUrlFromSupabaseUrl` (which already builds a clean URL from a Supabase base URL). This aligns the CI path with the setup path.

scripts/ensure-kit-webhook-endpoint.mjs
```diff
@@ -0,0 +1,49 @@
...
+ * Note: Kit does not return the webhook signing secret via API. Set KIT_WEBHOOK_SECRET
+ * in GitHub / Supabase separately (Kit dashboard → Webhooks → Signing secret).
+ */
+import process from 'node:process';
+
+import { ensureKitWebhook } from './lib/ensure-kit-webhook.mjs';
+
+async function main() {
+  const apiKey = process.env.KIT_API_KEY;
+  const url = process.env.KIT_WEBHOOK_URL;
```
mei-artificerinnovations: Accepts a pre-built URL from env, but every CI caller constructs it by concatenating a Supabase URL secret with `/functions/v1/kit-webhook` in a YAML expression (production line 118, staging line 118, preview line 275). If the secret has a trailing slash, the URL reaching here has a double slash that `normalizeKitWebhookUrl` won't fix.

The library already exports `kitWebhookUrlFromSupabaseUrl`, which is what `setup-kit.mjs` uses. Accepting `KIT_SUPABASE_URL` here and deriving the webhook path internally would unify the two paths and eliminate the secret-format dependency:

```js
import { ensureKitWebhook, kitWebhookUrlFromSupabaseUrl } from './lib/ensure-kit-webhook.mjs';

async function main() {
  const apiKey = process.env.KIT_API_KEY;
  const supabaseUrl = process.env.KIT_SUPABASE_URL;
  const url = supabaseUrl
    ? kitWebhookUrlFromSupabaseUrl(supabaseUrl)
    : process.env.KIT_WEBHOOK_URL; // legacy fallback
  if (!apiKey || !url) { ... }
```

Then the YAML steps pass `KIT_SUPABASE_URL: ${{ secrets.PRODUCTION_SUPABASE_URL }}` instead of the pre-concatenated URL.

.github/workflows/deploy-production.yml
```diff
@@ -111,6 +111,16 @@ jobs:
       - name: Sync edge shared packages
         run: npm run functions:sync-shared
 
+      - name: Ensure Kit webhook endpoint (production)
+        continue-on-error: true
+        env:
+          KIT_API_KEY: ${{ secrets.KIT_API_KEY }}
+          KIT_WEBHOOK_URL: ${{ secrets.PRODUCTION_SUPABASE_URL }}/functions/v1/kit-webhook
```
mei-artificerinnovations: YAML expression concatenation: if `PRODUCTION_SUPABASE_URL` has a trailing slash in the GitHub secret (e.g. `https://proj.supabase.co/`), this evaluates to `https://proj.supabase.co//functions/v1/kit-webhook`. The double slash survives `normalizeKitWebhookUrl` (which only strips trailing slashes, not mid-path duplicates) and is registered with Kit's API as-is.

Preferred fix: pass the raw Supabase URL and let `ensure-kit-webhook-endpoint.mjs` derive the path via `kitWebhookUrlFromSupabaseUrl`:

```yaml
KIT_SUPABASE_URL: ${{ secrets.PRODUCTION_SUPABASE_URL }}
# Remove KIT_WEBHOOK_URL
```

Same issue applies in `deploy-staging.yml` line 118 and `pr-preview-environment.yml` line 275.

.github/workflows/deploy-staging.yml
```diff
@@ -111,6 +111,16 @@ jobs:
       - name: Sync edge shared packages
         run: npm run functions:sync-shared
 
+      - name: Ensure Kit webhook endpoint (staging)
+        continue-on-error: true
+        env:
+          KIT_API_KEY: ${{ secrets.KIT_API_KEY }}
+          KIT_WEBHOOK_URL: ${{ secrets.STAGING_SUPABASE_URL }}/functions/v1/kit-webhook
```
mei-artificerinnovations: Same double-slash risk as `deploy-production.yml` line 118: if `STAGING_SUPABASE_URL` has a trailing slash in the secret, the concatenated `KIT_WEBHOOK_URL` has a malformed double slash that `normalizeKitWebhookUrl` won't collapse. Pass `KIT_SUPABASE_URL: ${{ secrets.STAGING_SUPABASE_URL }}` and derive the webhook path in the script.

.github/workflows/pr-preview-environment.yml
```diff
@@ -265,9 +265,24 @@ jobs:
...
 
-      - name: Deploy preview billing edge functions
+      - name: Ensure Kit webhook endpoint (preview)
+        if: >-
+          needs.preview-scope.outputs.run_stripe_billing == 'true' ||
+          needs.preview-scope.outputs.run_supabase_reset == 'true'
+        continue-on-error: true
+        env:
+          KIT_API_KEY: ${{ secrets.KIT_API_KEY }}
+          KIT_WEBHOOK_URL: ${{ secrets.PREVIEW_SUPABASE_URL }}/functions/v1/kit-webhook
```
mei-artificerinnovations: Same double-slash risk as production and staging: `PREVIEW_SUPABASE_URL` trailing slash → `PREVIEW_SUPABASE_URL//functions/v1/kit-webhook` → malformed URL registered with Kit. Pass `KIT_SUPABASE_URL: ${{ secrets.PREVIEW_SUPABASE_URL }}` and have `ensure-kit-webhook-endpoint.mjs` derive the path via `kitWebhookUrlFromSupabaseUrl`.

scripts/__tests__/setup-kit-script.test.mjs
```diff
@@ -0,0 +1,29 @@
...
+  assert.equal(acc.SETUP_KIT_SKIPPED, 'true');
```
mei-artificerinnovations: Hardcoded string literal `'SETUP_KIT_SKIPPED'` instead of the exported constant. Functionally correct today since `SETUP_KIT_SKIPPED_ENV === 'SETUP_KIT_SKIPPED'`, but if the constant is renamed this assertion silently passes while checking the wrong key.

```js
import { SETUP_KIT_SKIPPED_ENV } from '../lib/setup-kit.mjs';
// ...
assert.equal(acc[SETUP_KIT_SKIPPED_ENV], 'true');
```

scripts/lib/setup-kit.mjs
```diff
@@ -0,0 +1,253 @@
...
+        'Kit webhook signing secret (KIT_WEBHOOK_SECRET, from Kit dashboard): '
+      );
+  const webhookSecret =
+    webhookSecretRaw.trim() || String(acc.KIT_WEBHOOK_SECRET ?? '').trim();
+  if (!webhookSecret) {
+    logWarn(
+      'No webhook signing secret entered — inbound Kit unsubscribes will not verify until KIT_WEBHOOK_SECRET is set.'
+    );
+    delete acc.KIT_API_KEY;
+    delete acc.KIT_CRON_SECRET;
```
sarah-artificerinnovations: 🔴 **Data-loss bug — deletes pre-existing credentials on rollback**

When `hasKitCredentials(acc)` is false but `acc.KIT_API_KEY` already holds a valid value (e.g. the user ran `setup-full` earlier and their env file already has `KIT_API_KEY`), execution reaches line 221 and writes `acc.KIT_API_KEY = apiKey`. If the user then hits Enter at the webhook secret prompt, control reaches here and `delete acc.KIT_API_KEY` removes the value from `acc`. When the write phase later flushes `acc`, the previously-valid key is gone from `.env.local`.

Same for `KIT_CRON_SECRET`: it was freshly auto-generated on line 228–231 but is also deleted here before it was ever persisted.

**Fix — track what was newly set:**
```js
const wasPresent = {
  KIT_API_KEY: Boolean(String(acc.KIT_API_KEY ?? '').trim()),
  KIT_CRON_SECRET: Boolean(String(acc.KIT_CRON_SECRET ?? '').trim()),
};

// ... lines 221, 228-231 set acc.KIT_API_KEY and acc.KIT_CRON_SECRET ...

if (!webhookSecret) {
  logWarn('No webhook signing secret entered...');
  if (!wasPresent.KIT_API_KEY) delete acc.KIT_API_KEY;
  if (!wasPresent.KIT_CRON_SECRET) delete acc.KIT_CRON_SECRET;
  return;
}
```

Or simpler: don't delete at all. Just return without writing `KIT_WEBHOOK_SECRET`. The caller already guards partial credential writes via `pickKitGithubPayload`'s `if (v) out[key] = v` filter, so an incomplete set won't reach GitHub Actions secrets.

scripts/lib/ensure-kit-webhook.mjs
sarah-artificerinnovations: 🟡 **`normalizeKitWebhookUrl` only strips trailing slashes — mid-path double slashes are not fixed**

All three workflow files construct `KIT_WEBHOOK_URL` via string concatenation:
```yaml
KIT_WEBHOOK_URL: ${{ secrets.STAGING_SUPABASE_URL }}/functions/v1/kit-webhook
```

If the stored secret has a trailing slash (common when users copy the URL from the Supabase dashboard), the result is `https://abc.supabase.co//functions/v1/kit-webhook`. This regex only removes trailing slashes; the double slash at `//{functions}` survives. When `findKitWebhookByUrl` later compares this against webhooks registered with the clean URL, `normalizeKitWebhookUrl(w.target_url) === normalized` returns `false`, and the CI step creates a duplicate webhook on every deploy.

**Fix — add a mid-path deduplication pass:**
```js
export function normalizeKitWebhookUrl(url) {
  return String(url || '')
    .trim()
    .replace(/([^:])//+/g, '$1/')  // collapse mid-path double slashes
    .replace(//+$/, '');
}
```

The `([^:])` lookbehind preserves the `://` in the protocol.

.github/workflows/deploy-staging.yml
sarah-artificerinnovations: 🟡 **Trailing-slash hazard — see note on `normalizeKitWebhookUrl`**

If `STAGING_SUPABASE_URL` is stored with a trailing slash, this produces `https://…//functions/v1/kit-webhook`. `normalizeKitWebhookUrl` does not collapse mid-path double slashes, so `findKitWebhookByUrl` won't match a webhook registered under the clean URL and will spuriously re-create it on every deploy. Same issue applies to `STRIPE_WEBHOOK_URL` on line 106.

Either fix `normalizeKitWebhookUrl` (see comment there) or normalize the secret value before concatenation — though note GitHub Actions has no built-in trim filter, so the fix belongs in `normalizeKitWebhookUrl`.

.github/workflows/deploy-production.yml
sarah-artificerinnovations: 🟡 Same trailing-slash hazard as `deploy-staging.yml` line 118 — `PRODUCTION_SUPABASE_URL` with a trailing slash produces a double-slash URL that `normalizeKitWebhookUrl` does not fix. Fix: collapse mid-path double slashes in `normalizeKitWebhookUrl`.

.github/workflows/pr-preview-environment.yml
sarah-artificerinnovations: 🟡 Same trailing-slash hazard as the staging and production workflows — `PREVIEW_SUPABASE_URL` with a trailing slash produces a double-slash URL that `normalizeKitWebhookUrl` does not fix.

scripts/lib/ensure-kit-webhook.mjs
```diff
@@ -0,0 +1,167 @@
...
+      'X-Kit-Api-Key': apiKey,
+      Accept: 'application/json',
+      ...(opts.body !== undefined
+        ? { 'Content-Type': 'application/json' }
+        : {}),
+    },
+    ...(opts.body !== undefined ? { body: JSON.stringify(opts.body) } : {}),
+    signal: AbortSignal.timeout(30_000),
+  });
+  if (!res.ok) {
```
sarah-artificerinnovations, line 71: 🟡 **No 429 retry — rate-limit silently drops the webhook under `continue-on-error: true`**

`ensureKitWebhook` makes at least two sequential Kit API calls (GET list + possible POST create), and `ensureKitWebhooksForTiers` calls it three times — once per environment. If a deployment fires shortly after a prior setup run, six API calls hit Kit in rapid succession. On a 429 response this block throws immediately.

Because the CI step has `continue-on-error: true`, the job succeeds and no alert fires — but the webhook is missing in that environment, and Kit unsubscribe events are silently dropped until someone notices delivery failures.

**Minimum fix:** detect 429 and log clearly so an operator knows to re-run:
```js
if (res.status === 429) {
  const retryAfter = res.headers.get('Retry-After') ?? '?';
  throw new Error(`Kit API rate-limited (429) — Retry-After: ${retryAfter}s. Re-run the CI step.`);
}
```

**Better fix:** one retry with backoff:
```js
if (res.status === 429) {
  const delay = Number(res.headers.get('Retry-After') || 2) * 1000;
  await new Promise(r => setTimeout(r, delay));
  return kitRequest(apiKey, path, opts); // retry once
}
```

scripts/setup-kit.mjs
```diff
@@ -0,0 +1,311 @@
...
+  } catch {
+    return {};
+  }
+}
+
+function mergeRecords(base, overlay) {
+  return { ...base, ...overlay };
+}
+
+function stringifyDotEnv(obj) {
```
sarah-artificerinnovations: 🟡 **`stringifyDotEnv` produces different output than the version in `setup-full.mjs`**

This local implementation:
- Does not emit the `# Generated by…` header comment
- Does not sort keys alphabetically
- Always double-quotes values

`setup-full.mjs`'s version sorts keys and quotes selectively (only when the value contains spaces or `#`). If a developer runs `setup:full` first and then `setup:kit` standalone, `.env.local` gets rewritten without the header, without sorting, and with all values quoted — a noisy cosmetic diff and surprising behavior for what should be an idempotent operation.

**Fix:** extract to `scripts/lib/setup-dotenv.mjs` (which already exports `parseDotEnv` and `escapeDotEnvDoubleQuotedValue`) and import from there in both scripts. This guarantees both paths produce identical output.

scripts/__tests__/setup-kit-script.test.mjs
```diff
@@ -0,0 +1,29 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+
+import { pickKitGithubPayload } from '../setup-kit.mjs';
+import { phaseKit } from '../setup-kit.mjs';
```
sarah-artificerinnovations: 🔵 Duplicate import from the same specifier. Merge into one:
```js
import { pickKitGithubPayload, phaseKit } from '../setup-kit.mjs';
```

scripts/setup-kit.mjs
```diff
@@ -0,0 +1,311 @@
...
+    CLOUD_ENV_PATH,
+    stringifyDotEnv(mergeRecords(cloud, kitOnly)),
+    'utf8'
+  );
+  logInfo(
+    `Merged Kit key(s) into .env.cloud.generated.local (values not shown).`
+  );
+}
+
+function which(cmd) {
```
sarah-artificerinnovations, line 145: 🔵 **`which()` is duplicated.** `setup-full.mjs` has an identical implementation. Extract to `scripts/lib/setup-shell.mjs` (or whichever shared CLI-utilities module already exists) and import from there in both scripts.

scripts/lib/ensure-kit-webhook.mjs
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+ * @param {string} apiKey
+ * @param {string} targetUrl
+ * @returns {Promise<{ id: number; target_url: string } | null>}
+ */
+export async function findKitWebhookByUrl(apiKey, targetUrl) {
+  const normalized = normalizeKitWebhookUrl(targetUrl);
+  if (!normalized) return null;
+  const webhooks = await listKitWebhooks(apiKey);
```
sarah-artificerinnovations: 🔵 **`listKitWebhooks` is fetched once per tier.**

`ensureKitWebhook` calls `findKitWebhookByUrl` → `listKitWebhooks` for each of the three tiers (preview, staging, production). If the Kit account has many webhooks across multiple pages, this amplifies API calls by 3×. Consider fetching the full list once upstream:

```js
// In the caller (setup-kit.mjs or ensure-kit-webhook-endpoint.mjs):
const allWebhooks = await listKitWebhooks(apiKey);
for (const tier of tiers) {
  const existing = findWebhookInList(allWebhooks, tier.webhookUrl);
  if (!existing) await createKitUnsubscribeWebhook(apiKey, tier.webhookUrl);
}
```

sarah-artificerinnovations (commented):
Inline annotations for the items called out in the review comment — see individual line notes for details and suggested fixes.

scripts/setup-kit.mjs
```diff
@@ -0,0 +1,311 @@
...
+  });
+
+  if (flags.standalone && !flags.dryRun && !flags.skipGithub) {
+    await mergeKitEnvFiles(acc);
+    await syncKitGithubSecrets({
+      rl,
+      acc,
+      dryRun: flags.dryRun ?? false,
+      repoOverride: flags.githubRepo ?? '',
+    });
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 1616213: env file merge now runs whenever `standalone && !dryRun`; GitHub sync is gated separately on `!skipGithub`.

scripts/setup-kit.mjs
```diff
@@ -0,0 +1,311 @@
...
+
+  acc[SETUP_KIT_SKIPPED_ENV] = 'false';
+
+  const question = deps.question ?? (q => rl.question(q));
+  const readSecret =
+    deps.readSecret ??
+    (async prompt => {
+      const masked = await readMaskedLineIfTty(rl, input, output, prompt);
+      return masked ?? rl.question(prompt);
+    });
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 1616213: `readSecret` checks `flags.plainSecretPrompts` and uses plain `rl.question()` when set.

scripts/lib/setup-kit.mjs
```diff
@@ -0,0 +1,253 @@
...
+    logInfo = () => {},
+    logWarn = () => {},
+    question = async () => '',
+    readSecret = async () => '',
+  } = opts;
+
+  if (hasKitCredentials(acc)) {
+    logInfo(
+      'Kit credentials already present in session env (values not shown).'
+    );
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 1616213: dry-run no longer generates or stores `KIT_CRON_SECRET`, and webhook provisioning is log-only in dry-run.

scripts/setup-full.mjs
```diff
@@ -2364,6 +2374,13 @@ async function collectMissingGithubCiEnvIntoAcc(flags, rl, promptInput, acc) {
     );
   }
 
+  const kitMissing = details.filter(d => isKitGithubSecretDef(d.def));
+  if (kitMissing.length) {
+    logInfo(
+      `${kitMissing.length} Kit key(s) still missing — run \`npm run setup:kit\` or \`npm run setup:full -- --from=kit\` for guided collection, or enter below.`
+    );
+  }
+
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 1616213: `setup-full` now uses exported `listMissingKitEnvKeys(acc)` for the optional Kit nudge instead of the required-only GitHub CI list.

scripts/__tests__/setup-kit.test.mjs
```diff
@@ -0,0 +1,103 @@
+import assert from 'node:assert/strict';
+import crypto from 'node:crypto';
+import test from 'node:test';
+
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 1616213: added `setup-kit.test.mjs`, `setup-kit-script.test.mjs`, and `ensure-kit-webhook.test.mjs` to `test:unit:scripts`.

.github/workflows/pr-preview-environment.yml
ZappoMan: Fixed in 1616213: preview cron step now resolves `KIT_CRON_SECRET` with `PREVIEW_KIT_CRON_SECRET` fallback, matching staging/production.

scripts/setup-kit.mjs
ZappoMan: Fixed in 1616213: only log success when `ghSecretSetSync` returns status 0.

scripts/lib/setup-kit.mjs
ZappoMan: Fixed in 1616213: partial abort (missing webhook secret) now sets `acc[SETUP_KIT_SKIPPED_ENV] = 'true'` and restores the pre-prompt env snapshot.

scripts/__tests__/setup-kit-script.test.mjs
ZappoMan: Fixed in 1616213: test now imports and asserts `SETUP_KIT_SKIPPED_ENV`.

scripts/lib/setup-kit.mjs
ZappoMan: Same fix as above — addressed in 1616213.

scripts/__tests__/setup-kit-script.test.mjs
ZappoMan: Same fix as above — addressed in 1616213.

scripts/lib/ensure-kit-webhook.mjs
ZappoMan: Fixed in 1616213: `normalizeKitWebhookUrl` now collapses duplicate path slashes via `pathname.replace(/\/+/g, '/')`.

scripts/ensure-kit-webhook-endpoint.mjs
ZappoMan: Fixed in 1616213: CI workflows pass `KIT_SUPABASE_URL`; `ensure-kit-webhook-endpoint.mjs` derives the webhook path with `kitWebhookUrlFromSupabaseUrl`.

.github/workflows/deploy-production.yml
ZappoMan: Fixed in 1616213: switched to `KIT_SUPABASE_URL` (see `deploy-production.yml`).

.github/workflows/deploy-staging.yml
ZappoMan: Fixed in 1616213: switched to `KIT_SUPABASE_URL` (see `deploy-staging.yml`).

.github/workflows/pr-preview-environment.yml
ZappoMan: Fixed in 1616213: switched to `KIT_SUPABASE_URL` (see `pr-preview-environment.yml`).

scripts/__tests__/setup-kit-script.test.mjs
ZappoMan: Fixed in 1616213: test imports `SETUP_KIT_SKIPPED_ENV` and uses it in the assertion.

scripts/lib/setup-kit.mjs
ZappoMan: Fixed in 1616213: `kitEnvSnapshot` / `restoreKitEnvSnapshot` preserve pre-existing `KIT_*` values on partial abort instead of leaving partial writes.

scripts/lib/ensure-kit-webhook.mjs
ZappoMan: Fixed in 1616213: `normalizeKitWebhookUrl` collapses mid-path `//` (plus CI now passes base `KIT_SUPABASE_URL` instead of concatenating in YAML).

.github/workflows/deploy-staging.yml
ZappoMan: Fixed in 1616213: workflow uses `KIT_SUPABASE_URL` + derived path; normalization also collapses double slashes as a safety net.

.github/workflows/deploy-production.yml
ZappoMan: Fixed in 1616213: same as staging — `KIT_SUPABASE_URL` + path derivation + slash normalization.

.github/workflows/pr-preview-environment.yml
ZappoMan: Fixed in 1616213: same as staging/production.

scripts/lib/ensure-kit-webhook.mjs
ZappoMan: Fixed in 1616213: `kitRequest` retries up to 2 times on HTTP 429 with backoff.

scripts/setup-kit.mjs
ZappoMan: Fixed in 1616213: `setup-kit.mjs` now uses shared `stringifySetupDotEnv` from `setup-dotenv.mjs` (sorted keys, selective quoting, header comment).

scripts/__tests__/setup-kit-script.test.mjs
ZappoMan: Fixed in 1616213: merged into a single import line.

scripts/setup-kit.mjs
ZappoMan: Acknowledged — left `which()` duplicated for now to keep this PR focused; happy to extract to a shared util in a follow-up if you'd like.

scripts/lib/ensure-kit-webhook.mjs
ZappoMan: Fixed in 1616213: `ensureKitWebhooksForTiers` fetches `listKitWebhooks` once and passes the cached list into each `ensureKitWebhook` call.

scripts/setup-kit.mjs
ZappoMan: Tracking: https://github.com/Artificer-Innovations/BeakerStack/issues/374

mei-artificerinnovations (approved):
All four blocking items from my previous review are resolved. LGTM.

**Blocking items — all fixed:**

1. **`normalizeKitWebhookUrl` mid-URL double slashes** — now uses the `URL` constructor with `pathname.replace(/\/+/g, '/')`, handling double slashes anywhere in the path, not just trailing ones. ✅

2. **`ensure-kit-webhook-endpoint.mjs` URL derivation** — `resolveKitWebhookUrl()` now prefers `KIT_SUPABASE_URL` + `kitWebhookUrlFromSupabaseUrl()` and falls back to `KIT_WEBHOOK_URL` for explicit overrides. The divergence between setup and CI paths is gone. ✅

3. **YAML concatenation double-slash** — all three workflow files now pass `KIT_SUPABASE_URL` (the raw project URL) and let the script derive the `/functions/v1/kit-webhook` path safely. No more string concatenation risk in YAML. ✅

4. **Test hardcoded string** — `SETUP_KIT_SKIPPED_ENV` is now imported from `setup-kit.mjs` and used directly in the assertion. ✅

sarah-artificerinnovations (approved):
## Re-review — commit 1616213e (ZappoMan)

All blocking and design gap items from the first review are resolved. Approving.

---

### ✅ All items resolved

**🔴 Blocker — `collectKitEnvKeys` deletes pre-existing credentials on rollback**

Fixed with a snapshot/restore pattern: `kitEnvSnapshot(acc)` captures the state of all three Kit env keys before prompting, and `restoreKitEnvSnapshot(acc, snap)` puts them back precisely if the user skips the webhook secret. Keys that existed before the call are preserved; keys that didn't exist before are removed on abort. Clean and correct.

**🟡 `KIT_WEBHOOK_URL` fragile concatenation**

Fixed better than the suggested approach. `normalizeKitWebhookUrl` now parses via `new URL()` and normalizes pathname double-slashes with `u.pathname.replace(/\/+/g, '/')` — more robust than a regex over the full string. More importantly, CI no longer constructs the URL via concatenation at all: the workflow passes `KIT_SUPABASE_URL` to the endpoint script, which calls `kitWebhookUrlFromSupabaseUrl()` to derive the full path. The trailing-slash hazard is gone at the source.

**🟡 `kitRequest` no 429 retry**

Fixed with exponential backoff up to two retries (1s, then 2s). Matches the comment's suggested pattern.

**🟡 `stringifyDotEnv` inconsistency**

The local implementation is gone; `setup-kit.mjs` now imports `stringifySetupDotEnv` from `scripts/lib/setup-dotenv.mjs`, the same shared serializer used by `setup-full.mjs`. Consistent output on both paths.

**🔵 Double import in `setup-kit-script.test.mjs`**

Merged: `pickKitGithubPayload` and `phaseKit` are now in one import from `../setup-kit.mjs`. The separate import on line 6 is from `../lib/setup-kit.mjs` (a different file, for `SETUP_KIT_SKIPPED_ENV`) — not a duplicate.

**🔵 `listKitWebhooks` per-tier fetch**

`ensureKitWebhook` now accepts an optional pre-fetched `webhooks` list via `opts.webhooks`. `ensureKitWebhooksForTiers` fetches the full list once before the loop and passes it through, cutting API calls from 3× to 1× on the list path.

---

### One non-blocking note for follow-up

**`which()` is still duplicated** between `scripts/setup-kit.mjs` (line 145) and `scripts/setup-full.mjs` (line 1353). Identical implementations. Extracting to a shared helper in `scripts/lib/` is a small cleanup — no need to hold this PR for it, but worth a quick ticket.

---

LGTM. Ready to merge.

diego-artificerinnovations (approved):
## Approved

Both blocking and non-blocking items from the first review are resolved. Notes on each:

**1. Preview cron missing `PREVIEW_KIT_CRON_SECRET` fallback** → ✅ Fixed in `.github/workflows/pr-preview-environment.yml`. The "Ensure kit-sync cron job (preview)" step now exposes `PREVIEW_KIT_CRON_SECRET` in the env block and resolves it via `RESOLVED_KIT_CRON_SECRET="${KIT_CRON_SECRET:-${PREVIEW_KIT_CRON_SECRET:-}}"` — identical to the staging and production pattern. An adopter with only the legacy secret set will now get the cron registered correctly.

**2. `ghSecretSetSync` false-positive success log** → ✅ Fixed in `scripts/setup-kit.mjs`. Return value is now captured (`const status = ghSecretSetSync(...)`) and `logInfo` is gated on `status === 0`. The "GitHub secret set: X" message no longer fires after a failure.

---

A few nice extras came along in the fix commit that go beyond what was requested:

- **`kitEnvSnapshot()` / `restoreKitEnvSnapshot()`** — clean snapshot/restore replaces the old per-key `delete` approach in `collectKitEnvKeys`. The rollback on webhook-secret blank is now correct: all Kit keys revert to their pre-prompt state and `SETUP_KIT_SKIPPED_ENV` is set. This closes the latent bug where a partial abort could leave stale keys in `acc`.
- **Pre-fetched webhook list in `ensureKitWebhooksForTiers`** — `listKitWebhooks(apiKey)` is called once before the per-tier loop instead of once per tier. Saves 2 API round-trips on a 3-tier run.
- **`KIT_SUPABASE_URL` → derived webhook URL** — `ensure-kit-webhook-endpoint.mjs` now prefers `KIT_SUPABASE_URL` (used in CI) and derives the URL via `kitWebhookUrlFromSupabaseUrl()`, with `KIT_WEBHOOK_URL` still accepted as an explicit override. Cleaner than constructing the URL inline in the workflow.
- **`mergeKitEnvFiles` always runs in standalone mode** — no longer conditional on `!skipGithub`. Env file merge and GitHub secret sync are now correctly independent operations.
- **`import.meta.url` guard** — uses `url.pathToFileURL(process.argv[1] || '').href` comparison, consistent with the rest of the repo.

Non-blocking items from the first review (missing `collectKitEnvKeys` rollback test, duplicate import in `setup-kit-script.test.mjs`) remain open — neither is a merge blocker.

Ship it.

rahul-artificerinnovations (approved):
Commit `1616213e98` fully resolves both blocking issues and all non-blocking feedback from my initial review, plus every item raised by Mei, Sarah, Diego, and Copilot. This is ready to merge.

---

**Blocking items — resolved**

**1. Double-slash URL bypass of idempotency** — fixed on three fronts:
- CI workflows now pass `KIT_SUPABASE_URL` (the raw project URL) instead of constructing the webhook URL via string concatenation, eliminating the trailing-slash entry-error class entirely.
- `ensure-kit-webhook-endpoint.mjs` derives the URL via `kitWebhookUrlFromSupabaseUrl()`, which uses `new URL()` and reconstructs a clean path.
- `normalizeKitWebhookUrl` now URL-parses the input to collapse any internal `//` as a secondary defense.

**2. Partial abort leaves no `SETUP_KIT_SKIPPED_ENV` signal** — fixed via snapshot/restore pattern: `kitEnvSnapshot` captures pre-existing `KIT_*` values before collection begins; `restoreKitEnvSnapshot` replays them on all abort paths. Every abort path (no API key, no cron secret, no webhook secret) also sets `acc[SETUP_KIT_SKIPPED_ENV] = 'true'`.

---

**Non-blocking items — resolved**

- **Test string literal vs. constant**: `SETUP_KIT_SKIPPED_ENV` is now imported in `setup-kit-script.test.mjs`; assertion uses `acc[SETUP_KIT_SKIPPED_ENV]`.
- **Double guidance in `setup-full`**: The env-merge and gh-sync steps are now clearly separated, removing the redundant prompt sequence.
- **`--plain-secret-prompts` not respected in `readSecret`**: fixed.
- **Dry-run mutates cron secret**: guarded.
- **`kitMissing` from wrong list**: `setup-full.mjs` now uses the exported `listMissingKitEnvKeys(acc)` which checks the accumulator directly.
- **Kit tests not in `test:unit:scripts`**: all three test files added to `package.json`.
- **Preview cron missing fallback** (Diego): `PREVIEW_KIT_CRON_SECRET` with `RESOLVED_KIT_CRON_SECRET` pattern added.
- **`ghSecretSetSync` return value discarded** (Diego): return value checked; success only logged on `status === 0`.
- **Data-loss on partial abort** (Sarah): snapshot/restore as above.
- **No 429 retry** (Sarah): retry added with 1s/2s backoff, up to 2 retries.
- **`stringifyDotEnv` inconsistency** (Sarah): both scripts now use shared `stringifySetupDotEnv` from `setup-dotenv.mjs`; local duplicate removed.
- **Duplicate import** (Sarah): merged into one import statement.
- **`listKitWebhooks` called 3× per tier** (Sarah): list fetched once in `ensureKitWebhooksForTiers`, cache passed through to each `ensureKitWebhook` call.

**One acknowledged deferral**: `which()` is still duplicated between `setup-kit.mjs` and `setup-full.mjs`. This was called out in the fix commit and tracked in issue #374 — acceptable as a Phase 1 cleanup item, not a blocker here.

PR #368, opened 2026-05-24
Expand Playwright E2E coverage and category-grouped CI reports

ZappoMan:
## Summary
- Expands web E2E from ~16 smoke specs to full coverage across admin, profile, billing, marketing, navigation, and waitlist categories (34 spec files, 53 tests).
- Adds path-based category reporting for PR comments and a new `categories.html` artifact, grouped by `specs/<category>/` folder with no spec tagging required.
- Introduces shared fixtures for waitlist mode switching, waitlist data seeding, billing usage reset, and Stripe Checkout automation, plus serial Playwright projects to avoid signup-mode and seed-user conflicts.

## Test plan
- [ ] CI/preview E2E runs with Stripe billing specs (plan upgrade/downgrade, cadence, invoices)
- [ ] CI/preview E2E runs waitlist edge-function specs (public signup submit, invite signup)
- [ ] PR comment groups results by category; `categories.html` artifact uploads after Playwright run
- [ ] Local smoke: `npm run test:e2e:web` (47 passed, 6 skipped without `E2E_STRIPE_READY` / `E2E_WAITLIST_READY`)
- [ ] Category subsets: `npm run test:e2e:web:admin`, `:billing`, `:waitlist`
- [ ] Category report: `npm run test:e2e:web:report`


mei-artificerinnovations (changes requested):
## PR #368 review — Playwright E2E coverage + category CI reports

The CI parallelization approach (prepare → chromium/serial parallel → report) is the right architecture for this scope of tests. The category-reporting scripts are clean, well-factored, and properly tested. Four issues before this merges.

---

### Blocking

#### 1. `billing-stripe.spec.ts` — no billing state teardown; re-runs will fail

The Stripe tests run serially and leave the test user in Pro state (test 1 upgrades, test 2 schedules a downgrade, but the downgrade isn't applied immediately). If the spec suite is re-run against the same preview environment — whether for a retry or a re-trigger — test 1 ("upgrades Free to Pro") will fail because the user isn't on Free anymore. There's no `afterAll` that cancels the subscription and returns the user to Free.

Fix: add an `afterAll` that calls the billing cancel/downgrade RPC (or deletes the billing subscription row via service role) to reset the test user to Free before the suite exits.

#### 2. `completeStripeCheckout` — race condition on Stripe iframe card input

```ts
const numberInput = cardFrame
  .locator('input[name="cardnumber"], ...)
  .first();
if (await numberInput.count()) {
  // fill iframe inputs
} else {
  // fall back to direct page inputs
}
```

`count()` is evaluated at the moment of the call, not after waiting for the element. If the Stripe iframe exists but the card input hasn't rendered yet, `count()` returns 0 and the fallback path runs — against a different DOM structure that also doesn't have inputs yet. This is a race condition that will cause intermittent failures on slower CI runners.

Fix: use `waitForSelector` (or `locator.waitFor()`) on the iframe before checking for inputs, or restructure with a `try/catch` around the iframe-path that falls back only on actual `TimeoutError`.

#### 3. `billing-stripe.spec.ts` `beforeAll` — unhandled probe error causes test failures instead of skips

```ts
test.beforeAll(async ({ browser }) => {
  // ...
  try {
    stripeCheckoutReady = await probeStripeCheckoutReady(page);
  } finally {
    await context.close();
  }
});
```

If `probeStripeCheckoutReady` throws (network error, unexpected page state), the error propagates out of `beforeAll`. In Playwright, a `beforeAll` error causes all tests in the describe block to **fail** rather than be skipped. So a transient network error during the probe turns 4 skippable tests into 4 unexpected failures.

Fix: catch errors in `beforeAll` and set `stripeCheckoutReady = false`:

```ts
try {
  stripeCheckoutReady = await probeStripeCheckoutReady(page);
} catch {
  stripeCheckoutReady = false;
} finally {
  await context.close();
}
```

#### 4. `isWaitlistEdgeReady()` — returns `true` in CI unconditionally; no runtime probe

```ts
export function isWaitlistEdgeReady(): boolean {
  if (process.env.E2E_STRIPE_READY === '1') return true;
  if (process.env.E2E_TARGET === 'preview' || process.env.CI === 'true') return true;
  return false;
}
```

In `public-signup.spec.ts`, `test.skip(!isWaitlistEdgeReady(), ...)` fires inside the test body after the form is already rendered. When `CI === 'true'`, `isWaitlistEdgeReady()` is always true, so `test.skip()` never fires. If the `waitlist-capture` edge function isn't deployed to a given preview (e.g., a PR that doesn't deploy edge functions), the test attempts to submit the form, gets no success response, and times out — a genuine failure rather than a graceful skip.

The Stripe pattern (`probeStripeCheckoutReady` + `beforeAll`) is the right model. Waitlist specs should do the same: call the waitlist-capture endpoint with a canary email in `beforeAll`, set a flag based on whether it responds, and skip the whole describe block if it doesn't. 

Alternatively, if `isWaitlistEdgeReady()` is meant to be a guarantee in preview CI (the edge function is always deployed), that assumption needs to be documented and enforced — but the current flag name ("Ready") implies it's a probe, not a promise.

---

### Non-blocking

#### 5. Playwright browser cache doesn't skip `--with-deps` on hit

```yaml
- uses: actions/cache@v4
  id: playwright-cache
  with:
    key: playwright-${{ runner.os }}-${{ hashFiles('package-lock.json') }}-chromium
- run: npx playwright install --with-deps chromium
```

`--with-deps` reinstalls Chromium's system dependencies (apt packages) even when the browser binary cache hits. The standard pattern:

```yaml
- run: npx playwright install --with-deps chromium
  if: steps.playwright-cache.outputs.cache-hit != 'true'
- run: npx playwright install-deps chromium
  if: steps.playwright-cache.outputs.cache-hit == 'true'
```

Minor CI time saving — applies to all three jobs that cache browsers.

#### 6. `PLAYWRIGHT_OUTCOME` expression is fragile

```yaml
PLAYWRIGHT_OUTCOME: ${{ (needs.e2e-chromium.result != 'success' || needs.e2e-serial.result != 'success') && 'failure' || 'success' }}
```

This evaluates correctly today but is hard to read and will silently break if a third test job is added. Consider an explicit step that sets an output variable:

```yaml
- id: outcome
  run: |
    if [[ "${{ needs.e2e-chromium.result }}" == "success" && "${{ needs.e2e-serial.result }}" == "success" ]]; then
      echo "value=success" >> $GITHUB_OUTPUT
    else
      echo "value=failure" >> $GITHUB_OUTPUT
    fi
```

#### 7. `admin/waitlist.spec.ts` — cleanup may miss newly created invite entry

```ts
const entry = await findWaitlistEntryByEmail(email);
if (entry) {
  await deleteWaitlistEntry(entry.id);
}
```

`findWaitlistEntryByEmail` is called once with no retry, immediately after the invite is created. If there's replication lag or the invite is queued asynchronously, `entry` may be null and the test entry leaks. `waitlist-data.fixture.ts` has a `waitForWaitlistEntry` helper with a 12-attempt retry loop for exactly this reason — use it here instead of the bare `findWaitlistEntryByEmail` call.

#### 8. `PLAYWRIGHT_WORKERS: 4` on 2-vCPU GitHub runners

`ubuntu-latest` has 2 vCPUs. Running 4 workers simultaneously will cause CPU contention, longer page load times, and more flaky failures on Chromium tests. The previous default was 2 workers in CI. Recommend starting at 2 for the first run of the new parallel structure and increasing if tests are stable.

#### 9. `mergePlaywrightJsonReports` test doesn't cover error array merging

The test verifies `suites` and `stats` merging but not `errors`. A second test case with non-empty `errors` arrays would complete the coverage of the merge function's behavior.

---

### Summary

| Issue | Severity |
|---|---|
| Billing teardown missing — re-runs will fail | Blocking |
| `completeStripeCheckout` iframe race condition | Blocking |
| `beforeAll` probe error causes test failures not skips | Blocking |
| `isWaitlistEdgeReady()` — no runtime probe; CI always attempts | Blocking |
| Browser cache doesn't skip `--with-deps` | Non-blocking |
| `PLAYWRIGHT_OUTCOME` expression fragile | Non-blocking |
| Waitlist cleanup uses bare `findWaitlistEntryByEmail` | Non-blocking |
| 4 workers on 2-vCPU runners may increase flakiness | Non-blocking |
| `mergePlaywrightJsonReports` test missing `errors` coverage | Non-blocking |

tests/e2e/web/specs/billing/billing-stripe.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    await expect(
+      page.getByRole('heading', { name: 'Invoices', level: 2 })
+    ).toBeVisible();
+    await expect(
+      page
+        .getByRole('link', { name: 'Explore plans' })
+        .or(page.locator('table'))
+    ).toBeVisible();
+  });
+});
```
rahul-artificerinnovations, line 128: **Blocking: missing `afterAll` — seed user left on Pro subscription.**

The `schedules downgrade to Free` test clicks Confirm downgrade, but Stripe schedules the cancellation for end of billing period — the user remains on Pro. When `metered-usage.spec.ts` runs next alphabetically, `"reaches collection limit on free plan"` expects a `"Limit reached"` button that only appears on the Free plan.

Add an `afterAll` that resets the subscription:
```ts
test.afterAll(async () => {
  const seed = JSON.parse(readFileSync(e2eStatePath, 'utf8')) as E2eSeedState;
  await resetBillingPlanForUser(seed.userId); // cancel sub + revert to Free immediately
});
```
Alternatively add a `beforeAll` in `metered-usage.spec.ts` that asserts/resets the plan before any usage tests run.

tests/e2e/web/specs/admin/waitlist.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    await expect(
+      page.getByRole('button', { name: 'Copy invite link' })
+    ).toBeVisible();
+
+    const entry = await findWaitlistEntryByEmail(email);
+    if (entry) {
+      await deleteWaitlistEntry(entry.id);
+    }
+  });
+
```
rahul-artificerinnovations, line 35: **Non-blocking: cleanup skipped on assertion failure.**

If `expect(page.getByText(...)).toBeVisible()` times out, the `findWaitlistEntryByEmail` / `deleteWaitlistEntry` block below never executes and the entry leaks into the preview DB. The second test in this file uses `try/finally` correctly — match it:

```ts
try {
  await expect(page.getByText(`Invite created for ${email}`)).toBeVisible({ timeout: 15_000 });
  await expect(page.getByRole('button', { name: 'Copy invite link' })).toBeVisible();
} finally {
  const entry = await findWaitlistEntryByEmail(email);
  if (entry) await deleteWaitlistEntry(entry.id);
}
```

scripts/ci/format-playwright-e2e-pr-comment.mjs
```diff
@@ -88,8 +99,8 @@ export function formatMinutes(durationMs) {
   return `${(durationMs / 60_000).toFixed(2)} min`;
 }
 
-/** Background for full-width spec file section rows in the Details HTML table. */
-export const FILE_SECTION_ROW_BG = '#eaeef2';
+/** @deprecated Use CATEGORY_SECTION_ROW_BG; kept for existing tests during transition. */
+export const FILE_SECTION_ROW_BG = CATEGORY_SECTION_ROW_BG;
```
rahul-artificerinnovations: **Non-blocking: deprecation comment is already stale.**

This PR migrates every test to import `CATEGORY_SECTION_ROW_BG` directly — the changes to `format-playwright-e2e-pr-comment.test.mjs` in this diff replace all `FILE_SECTION_ROW_BG` references with `CATEGORY_SECTION_ROW_BG`. The transition is complete. Either remove this re-export entirely, or drop the `@deprecated` annotation so it does not imply an ongoing migration that is already done.

tests/e2e/web/fixtures/stripe-checkout.fixture.ts
```diff
@@ -0,0 +1,197 @@
...
+    return false;
+  }
+
+  try {
+    const checkoutResponse = page.waitForResponse(
+      response =>
+        isBillingStripeCheckoutRequest(
+          response.url(),
+          response.request().method()
+        ),
```
rahul-artificerinnovations, line 147: **Non-blocking: non-iframe fallback is dead code.**

Stripe Checkout has embedded card inputs in iframes since 2019. `input[name="cardNumber"]` on the main page only existed in Stripe.js v2 (EOL 2022). If BeakerStack uses current Stripe Checkout, this branch never executes. Remove it, or add a comment identifying the specific Stripe integration path where it applies.

rahul-artificerinnovations (changes requested):
The jobs split (prepare → chromium → serial → report), the category-aware PR comment, and the fixture library are all well-designed. Two blocking issues need fixes before merge.

## Blocking

**1. `billing-stripe.spec.ts` leaves seed user on Pro — `metered-usage.spec.ts` will fail**

Both specs are in the `shared-state` project and run serially in alphabetical order. `billing-stripe` upgrades the seed user to Pro. The second test *schedules* a downgrade to Free, but Stripe does not execute that until end of billing period — the user remains on Pro. When `metered-usage.spec.ts` runs next, `"reaches collection limit on free plan"` clicks "New collection" twice and expects a "Limit reached" button that only appears on the Free plan. On Pro, that button never appears and the assertion times out.

Needs an `afterAll` in `billing-stripe.spec.ts` that restores the seed user to Free — see inline comment.

**2. `profile-edit.spec.ts` removed with no replacement**

The PR description lists profile as one of the six covered categories, and `playwright.config.ts` wires `/specs\/profile\/` into the `shared-state` project. The only profile change in this diff is the deletion of `profile-edit.spec.ts`. Zero profile tests remain after this PR lands. Either the spec was moved (it should appear elsewhere in the diff) or coverage was silently dropped — neither is acceptable without an explicit decision.

## Non-blocking

**3. `e2e-serial` fires even on `e2e-chromium` infra failure** (`e2e-web-pr-approval.yml`, `e2e-serial` job)

```yaml
if: always() && needs.e2e-prepare.outputs.qualified == 'true' && needs.e2e-prepare.result == 'success'
```

No check on `e2e-chromium.result`. When chromium fails because the preview is unreachable, serial tests start, hit the same infrastructure failure, and waste ~3–5 min. Consider adding `&& needs.e2e-chromium.result != 'cancelled'` so serial still runs on chromium test failures (maximum signal) but skips when chromium itself never ran.

**4. First test in `admin/waitlist.spec.ts` leaks data on assertion failure** — see inline comment.

**5. `FILE_SECTION_ROW_BG` deprecation comment is already stale** — see inline comment.

**6. `completeStripeCheckout` non-iframe fallback is dead code** — see inline comment.

tests/e2e/web/playwright.config.ts
```diff
@@ -7,17 +7,40 @@ const isCi = Boolean(process.env.CI);
...
+    : 4;
+
 export default defineConfig({
   testDir: path.join(configDir, 'specs'),
   fullyParallel: true,
   forbidOnly: isCi,
   retries: isCi ? 1 : 0,
-  workers: isCi ? 2 : undefined,
+  workers: isCi ? ciWorkers : 1,
   reporter: [
```
diego-artificerinnovations, line 40: `workers: isCi ? ciWorkers : 1` forces local dev to 1 worker. The previous value was `undefined`, which uses Playwright's default (50% of CPU cores). Forcing 1 locally makes the 34-spec chromium suite run sequentially — a meaningful perf regression for local iteration.

The serial projects already enforce serial execution via `fullyParallel: false` and `--workers=1` in CI. The chromium project doesn't need to be throttled locally.

Fix:
```ts
workers: isCi ? ciWorkers : undefined,
```

tests/e2e/web/specs/billing/billing-stripe.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+      page.getByRole('heading', { name: 'Downgrade to Free?' })
+    ).toBeVisible();
+    await page.getByRole('button', { name: 'Confirm downgrade' }).click();
+
+    await expect(page.getByText(/downgrade|scheduled|cancel/i)).toBeVisible({
+      timeout: 15_000,
+    });
+  });
+
+  test('shows annual switch option for paid plan', async ({
```
diego-artificerinnovations, line 81: This test assumes the user is still on Pro plan when it runs, but test 2 ("schedules downgrade to Free") runs before it in serial mode. If the billing UI suppresses the annual cadence switch for subscriptions with a pending downgrade, this will fail.

The implicit dependency chain is: test 1 upgrades to Pro → test 2 schedules downgrade → test 3 expects Pro-plan annual-switch button. That's fragile with no guard.

Options:
- Add `await expectCurrentPlan(page, 'Pro');` at the start of this test and test 4 to make the assumption explicit and get a clear failure message if it breaks.
- Reorder: move the downgrade test to last so state is: upgrade → annual switch → invoices → downgrade.

tests/e2e/web/fixtures/waitlist-data.fixture.ts
```diff
@@ -0,0 +1,78 @@
...
+  captureWaitlistEntry,
+  findWaitlistEntryByEmail,
+} from '../../../utils/integration-fixtures';
+import {
+  createServiceRoleClient,
+  createWebTestClient,
+} from '../../../utils/test-clients';
+import { signInTestUser } from '../../../utils/test-helpers';
+import { e2eAdminStatePath, type E2eSeedState } from '../env';
+
```
diego-artificerinnovations, line 12: 12 attempts × 250ms = 3 seconds total. This is too tight for CI when the edge function container is cold. The waitlist-capture function processes the request and writes to DB, but the DB write may not be visible (replication lag, cold start) for 5–15+ seconds after the HTTP response returns.

This will be a source of intermittent CI failures, especially on the first test run after a cold deploy.

Suggested increase:
```ts
for (let attempt = 0; attempt < 40; attempt += 1) {
  const entry = await findWaitlistEntryByEmail(email);
  if (entry) {
    return entry;
  }
  await new Promise(resolve => setTimeout(resolve, 500));
}
```
That gives 20 seconds, which is consistent with what other edge-function-backed polls in this repo use.

diego-artificerinnovations (changes requested):
## Review

The CI refactor (prepare/chromium/serial/report split), Playwright browser caching, auth artifact hand-off, and the category-grouped reporting pipeline are all well-designed. The merge script is clean and the category derivation from spec path is correct. Three blocking issues below; a few suggestions after.

---

### Blocking

**1. Workers regression for local dev** — see inline comment on `playwright.config.ts`

**2. `billing-stripe.spec.ts` serial test state assumption** — see inline comment

**3. `waitlist-data.fixture.ts` poll timeout too tight for CI** — see inline comment

---

### Suggestions (non-blocking)

**`public-signup.spec.ts` — skip check after navigation**

`test.skip(!isWaitlistEdgeReady(), ...)` is placed after navigating to `/signup` and asserting heading/email visibility. For an infrastructure skip (not a UI assertion), it should be the first statement in the test body. In CI it's always ready so it won't flake, but the pattern is misleading.

**`format-playwright-e2e-pr-comment.mjs` — `FILE_SECTION_ROW_BG` alias can be deleted**

The `@deprecated` comment says the alias is "kept for existing tests during transition." The transition is complete — the test file in this same PR already imports `CATEGORY_SECTION_ROW_BG` directly. Delete the alias rather than leaving deprecated dead code.

**`waitlist.fixture.ts` — `withSignupMode` is unused**

`withSignupMode` is exported but never imported anywhere in the specs. Every waitlist/signup-mode spec uses the `setSignupModeForTests` + `restoreSignupMode` pattern instead. Either use `withSignupMode` in those specs (it wraps the same boilerplate) or drop the export.

sarah-artificerinnovations:
## Code review — PR #368 (ZappoMan)

Solid PR. The four-job CI split (prepare → chromium → serial → report) is the right structure: one global-setup, parallelism where safe, strict serial ordering where required, teardown always runs. The category grouping and merge pipeline are well-factored — pure-function exports, isMain guards, graceful null returns on missing reports. Good work overall. A few issues need attention before merge.

---

### 🔴 Two issues

**1. `e2e-chromium` is missing auth artifact verification**

`e2e-serial` verifies the downloaded auth artifacts before running:

```yaml
- name: Verify E2E auth artifacts
  run: |
    test -s tests/e2e/web/.auth/state.json
    test -s tests/e2e/web/.auth/admin-state.json
    ...
```

`e2e-chromium` downloads the same artifact but skips verification. If the artifact download silently produces empty files (e.g., an artifact upload race or `if-no-files-found: ignore` in a future tweak), chromium tests fail with a confusing Playwright "no storage state" error rather than a clear "auth file missing or empty" message. Add the same verify step to `e2e-chromium` immediately after the download step.

**2. `isStripeCheckoutReady()` and `isWaitlistEdgeReady()` auto-enable on `CI=true`**

Both fixtures return `true` when `process.env.CI === 'true'`, regardless of whether the external service is actually functional:

```ts
if (process.env.E2E_TARGET === 'preview' || process.env.CI === 'true') {
  return true;
}
```

This means: if the billing-stripe edge function isn't deployed to preview yet, or if the Stripe webhook secret is wrong, or if waitlist-ops is down — the tests don't skip, they fail. The intent is correct for the happy path, but the failure mode is wrong. If the service isn't ready, the test should skip, not fail with an opaque upstream error.

Consider inverting the logic: these flags default to `false` in CI and require an explicit opt-in (`E2E_STRIPE_READY=1`, `E2E_WAITLIST_READY=1`) set only after the services are confirmed deployed. Or keep the `CI=true` default but have `probeStripeCheckoutReady` / an equivalent probe for waitlist gate the `isWaitlistEdgeReady` flag, with explicit skip on probe failure. The probe pattern you've already built for Stripe is the right idea — just make sure the fallback is `test.skip`, not a hard failure.

---

### 🟡 Four design gaps

**3. `probeStripeCheckoutReady` creates a real Stripe session**

The probe is described in-code as "non-destructive" but it does `await upgradeButton.click()`, which triggers the full `billing-stripe` checkout flow and creates a real Stripe session object. That's a side effect on Stripe's side on every E2E run. It also leaves the probe `page` navigated to `checkout.stripe.com`, which is benign only because the `beforeAll` page is separate from per-test pages.

A cleaner probe: add a `GET /health` or `GET /status` endpoint to the `billing-stripe` edge function that returns 200 if Stripe is configured (checks env vars, optionally pings Stripe API), without creating a session. The probe then hits that endpoint. If that's out of scope for this PR, document the side-effect clearly in the fixture comment and track the cleanup.

**4. `completeStripeCheckout` has a dual iframe/direct-input path that may be legacy**

```ts
const cardFrame = page
  .frameLocator(
    'iframe[name^="__privateStripeFrame"], iframe[title*="Secure"], iframe[src*="stripe"]'
  )
  .first();

const numberInput = cardFrame.locator('input[name="cardnumber"], ...').first();
if (await numberInput.count()) {
  // iframe path
} else {
  // direct input path
}
```

Stripe's hosted Checkout (redirect to `checkout.stripe.com`) renders card inputs as direct DOM elements, not iframes — iframes are the old Stripe.js Elements pattern for embedded checkout. If this PR uses only the hosted redirect flow (which `startProUpgradeCheckout` confirms by `waitForURL(/checkout.stripe.com/)`), the iframe branch is likely dead code from an older integration and will silently pass `await numberInput.count() === 0` every run. Remove the iframe branch, or add a comment explaining when it would fire. Dead selector chains in payment flows are a maintenance risk.

**5. `FILE_SECTION_ROW_BG` deprecated alias shipped without migrating callers**

```ts
/** @deprecated Use CATEGORY_SECTION_ROW_BG; kept for existing tests during transition. */
export const FILE_SECTION_ROW_BG = CATEGORY_SECTION_ROW_BG;
```

If the callers are tests in this same PR, update them now — the "transition" is happening in this PR and there's no reason to ship a deprecated export on day one. If the callers are in other PRs or branches, track it. Either way, the deprecation comment should say where the callers live and what the plan is so it doesn't become permanent.

**6. CI `e2e-serial` project ordering is implicit and fragile**

```yaml
npx playwright test --config tests/e2e/web/playwright.config.ts   --project=shared-state   --project=signup-mode   --no-deps   --workers=1
```

`--no-deps` bypasses Playwright's dependency chain, so `shared-state → signup-mode` ordering depends entirely on the order these `--project` flags are listed. If a future contributor adds a third serial project and puts it last (e.g., `--project=another-serial`), but it actually needs to run before `signup-mode`, CI breaks silently. Add a comment above this step explaining that the flag order is load-bearing and maps to the dependency chain defined in `playwright.config.ts`.

---

### 🔵 Minor

**7. Warm-up curl silently swallows failures**

```yaml
curl -sf ... --retry 5 --retry-all-errors "class URL {
  #context = new URLContext();
  #searchParams;
  #searchParamsModified;

  static {
    setURLSearchParamsModified = (obj) => {
      // When URLSearchParams changes, we lazily update URL on the next read/write for performance.
      obj.#searchParamsModified = true;

      // If URL has an existing search, remove it without cascading back to URLSearchParams.
      // Do this to avoid any internal confusion about whether URLSearchParams or URL is up-to-date.
      if (obj.#context.hasSearch) {
        obj.#updateContext(bindingUrl.update(obj.#context.href, updateActions.kSearch, ''));
      }
    };
  }

  constructor(input, base = undefined, parseSymbol = undefined) {
    markTransferMode(this, false, false);

    if (arguments.length === 0) {
      throw new ERR_MISSING_ARGS('url');
    }

    // StringPrototypeToWellFormed is not needed.
    input = `${input}`;

    if (base !== undefined) {
      base = `${base}`;
    }

    let href;
    if (arguments.length < 3) {
      href = bindingUrl.parse(input, base, true);
    } else {
      const raiseException = parseSymbol !== kParseURLSymbol;
      const interpretAsWindowsPath = parseSymbol === kCreateURLFromWindowsPathSymbol;
      const pathToFileURL = interpretAsWindowsPath || (parseSymbol === kCreateURLFromPosixPathSymbol);
      href = pathToFileURL ?
        bindingUrl.pathToFileURL(input, interpretAsWindowsPath, base) :
        bindingUrl.parse(input, base, raiseException);
    }
    if (href) {
      this.#updateContext(href);
    }
  }

  static parse(input, base = undefined) {
    if (arguments.length === 0) {
      throw new ERR_MISSING_ARGS('url');
    }
    const parsedURLObject = new URL(input, base, kParseURLSymbol);
    return parsedURLObject.href ? parsedURLObject : null;
  }

  [inspect.custom](depth, opts) {
    if (typeof depth === 'number' && depth < 0)
      return this;

    const constructor = getConstructorOf(this) || URL;
    const obj = { __proto__: { constructor } };

    obj.href = this.href;
    obj.origin = this.origin;
    obj.protocol = this.protocol;
    obj.username = this.username;
    obj.password = this.password;
    obj.host = this.host;
    obj.hostname = this.hostname;
    obj.port = this.port;
    obj.pathname = this.pathname;
    obj.search = this.search;
    obj.searchParams = this.searchParams;
    obj.hash = this.hash;

    if (opts.showHidden) {
      obj[contextForInspect] = this.#context;
    }

    return `${constructor.name} ${inspect(obj, opts)}`;
  }

  #getSearchFromContext() {
    if (!this.#context.hasSearch) return '';
    let endsAt = this.#context.href.length;
    if (this.#context.hasHash) endsAt = this.#context.hash_start;
    if (endsAt - this.#context.search_start <= 1) return '';
    return StringPrototypeSlice(this.#context.href, this.#context.search_start, endsAt);
  }

  #getSearchFromParams() {
    if (!this.#searchParams?.size) return '';
    return `?${this.#searchParams}`;
  }

  #ensureSearchParamsUpdated() {
    // URL is updated lazily to greatly improve performance when URLSearchParams is updated repeatedly.
    // If URLSearchParams has been modified, reflect that back into URL, without cascading back.
    if (this.#searchParamsModified) {
      this.#searchParamsModified = false;
      this.#updateContext(bindingUrl.update(this.#context.href, updateActions.kSearch, this.#getSearchFromParams()));
    }
  }

  /**
   * Update the internal context state for URL.
   * @param {string} href New href string from `bindingUrl.update`.
   * @param {boolean} [shouldUpdateSearchParams] If the update has potential to update search params (href/search).
   */
  #updateContext(href, shouldUpdateSearchParams = false) {
    const previousSearch = shouldUpdateSearchParams && this.#searchParams &&
      (this.#searchParamsModified ? this.#getSearchFromParams() : this.#getSearchFromContext());

    this.#context.href = href;

    const {
      0: protocol_end,
      1: username_end,
      2: host_start,
      3: host_end,
      4: port,
      5: pathname_start,
      6: search_start,
      7: hash_start,
      8: scheme_type,
    } = bindingUrl.urlComponents;

    this.#context.protocol_end = protocol_end;
    this.#context.username_end = username_end;
    this.#context.host_start = host_start;
    this.#context.host_end = host_end;
    this.#context.port = port;
    this.#context.pathname_start = pathname_start;
    this.#context.search_start = search_start;
    this.#context.hash_start = hash_start;
    this.#context.scheme_type = scheme_type;

    if (this.#searchParams) {
      // If the search string has updated, URL becomes the source of truth, and we update URLSearchParams.
      // Only do this when we're expecting it to have changed, otherwise a change to hash etc.
      // would incorrectly compare the URLSearchParams state to the empty URL search state.
      if (shouldUpdateSearchParams) {
        const currentSearch = this.#getSearchFromContext();
        if (previousSearch !== currentSearch) {
          setURLSearchParams(this.#searchParams, currentSearch);
          this.#searchParamsModified = false;
        }
      }

      // If we have a URLSearchParams, ensure that URL is up-to-date with any modification to it.
      this.#ensureSearchParamsUpdated();
    }
  }

  toString() {
    // Updates to URLSearchParams are lazily propagated to URL, so we need to check we're in sync.
    this.#ensureSearchParamsUpdated();
    return this.#context.href;
  }

  get href() {
    // Updates to URLSearchParams are lazily propagated to URL, so we need to check we're in sync.
    this.#ensureSearchParamsUpdated();
    return this.#context.href;
  }

  set href(value) {
    value = `${value}`;
    const href = bindingUrl.update(this.#context.href, updateActions.kHref, value);
    if (!href) { throw new ERR_INVALID_URL(value); }
    this.#updateContext(href, true);
  }

  // readonly
  get origin() {
    const protocol = StringPrototypeSlice(this.#context.href, 0, this.#context.protocol_end);

    // Check if scheme_type is not `NOT_SPECIAL`
    if (this.#context.scheme_type !== 1) {
      // Check if scheme_type is `FILE`
      if (this.#context.scheme_type === 6) {
        return 'null';
      }
      return `${protocol}//${this.host}`;
    }

    if (protocol === 'blob:') {
      const path = this.pathname;
      if (path.length > 0) {
        try {
          const out = new URL(path);
          // Only return origin of scheme is `http` or `https`
          // Otherwise return a new opaque origin (null).
          if (out.#context.scheme_type === 0 || out.#context.scheme_type === 2) {
            return `${out.protocol}//${out.host}`;
          }
        } catch {
          // Do nothing.
        }
      }
    }

    return 'null';
  }

  get protocol() {
    return StringPrototypeSlice(this.#context.href, 0, this.#context.protocol_end);
  }

  set protocol(value) {
    const href = bindingUrl.update(this.#context.href, updateActions.kProtocol, `${value}`);
    if (href) {
      this.#updateContext(href);
    }
  }

  get username() {
    if (this.#context.protocol_end + 2 < this.#context.username_end) {
      return StringPrototypeSlice(this.#context.href, this.#context.protocol_end + 2, this.#context.username_end);
    }
    return '';
  }

  set username(value) {
    const href = bindingUrl.update(this.#context.href, updateActions.kUsername, `${value}`);
    if (href) {
      this.#updateContext(href);
    }
  }

  get password() {
    if (this.#context.host_start - this.#context.username_end > 0) {
      return StringPrototypeSlice(this.#context.href, this.#context.username_end + 1, this.#context.host_start);
    }
    return '';
  }

  set password(value) {
    const href = bindingUrl.update(this.#context.href, updateActions.kPassword, `${value}`);
    if (href) {
      this.#updateContext(href);
    }
  }

  get host() {
    let startsAt = this.#context.host_start;
    if (this.#context.href[startsAt] === '@') {
      startsAt++;
    }
    // If we have an empty host, then the space between components.host_end and
    // components.pathname_start may be occupied by /.
    if (startsAt === this.#context.host_end) {
      return '';
    }
    return StringPrototypeSlice(this.#context.href, startsAt, this.#context.pathname_start);
  }

  set host(value) {
    const href = bindingUrl.update(this.#context.href, updateActions.kHost, `${value}`);
    if (href) {
      this.#updateContext(href);
    }
  }

  get hostname() {
    let startsAt = this.#context.host_start;
    // host_start might be "@" if the URL has credentials
    if (this.#context.href[startsAt] === '@') {
      startsAt++;
    }
    return StringPrototypeSlice(this.#context.href, startsAt, this.#context.host_end);
  }

  set hostname(value) {
    const href = bindingUrl.update(this.#context.href, updateActions.kHostname, `${value}`);
    if (href) {
      this.#updateContext(href);
    }
  }

  get port() {
    if (this.#context.hasPort) {
      return `${this.#context.port}`;
    }
    return '';
  }

  set port(value) {
    const href = bindingUrl.update(this.#context.href, updateActions.kPort, `${value}`);
    if (href) {
      this.#updateContext(href);
    }
  }

  get pathname() {
    let endsAt;
    if (this.#context.hasSearch) {
      endsAt = this.#context.search_start;
    } else if (this.#context.hasHash) {
      endsAt = this.#context.hash_start;
    }
    return StringPrototypeSlice(this.#context.href, this.#context.pathname_start, endsAt);
  }

  set pathname(value) {
    const href = bindingUrl.update(this.#context.href, updateActions.kPathname, `${value}`);
    if (href) {
      this.#updateContext(href);
    }
  }

  get search() {
    // Updates to URLSearchParams are lazily propagated to URL, so we need to check we're in sync.
    this.#ensureSearchParamsUpdated();
    return this.#getSearchFromContext();
  }

  set search(value) {
    const href = bindingUrl.update(this.#context.href, updateActions.kSearch, StringPrototypeToWellFormed(`${value}`));
    if (href) {
      this.#updateContext(href, true);
    }
  }

  // readonly
  get searchParams() {
    // Create URLSearchParams on demand to greatly improve the URL performance.
    if (this.#searchParams == null) {
      this.#searchParams = new URLSearchParams(this.#getSearchFromContext());
      setURLSearchParamsContext(this.#searchParams, this);
      this.#searchParamsModified = false;
    }
    return this.#searchParams;
  }

  get hash() {
    if (!this.#context.hasHash || (this.#context.href.length - this.#context.hash_start <= 1)) {
      return '';
    }
    return StringPrototypeSlice(this.#context.href, this.#context.hash_start);
  }

  set hash(value) {
    const href = bindingUrl.update(this.#context.href, updateActions.kHash, `${value}`);
    if (href) {
      this.#updateContext(href);
    }
  }

  toJSON() {
    // Updates to URLSearchParams are lazily propagated to URL, so we need to check we're in sync.
    this.#ensureSearchParamsUpdated();
    return this.#context.href;
  }

  static canParse(url, base = undefined) {
    if (arguments.length === 0) {
      throw new ERR_MISSING_ARGS('url');
    }

    url = `${url}`;

    if (base !== undefined) {
      return bindingUrl.canParse(url, `${base}`);
    }

    // It is important to differentiate the canParse call statements
    // since they resolve into different v8 fast api overloads.
    return bindingUrl.canParse(url);
  }
}" -o /dev/null || true
```

The `|| true` means a cold preview that never responds won't surface until E2E global setup fails with a cryptic connection error. Consider replacing with `|| echo "::warning::Preview warm-up failed — E2E setup may be slower"` so there's a visible signal without blocking the run.

**8. Category summary table collapses `passedAfterRetry` into the Passed column**

```ts
rows.push(`| ${category} | ${counts.passed + counts.passedAfterRetry} | ${counts.failed} | ${counts.skipped} |`);
```

A test that passes only after retry is a flaky test. Hiding it in the Passed count means the category table gives no signal about test reliability trends. Consider a separate column or a footnote (e.g., "12 passed (2 after retry)") — the detailed table already tracks this per-test, so the summary hiding it feels inconsistent.

**9. `withSignupMode` default assumes `'open'` when DB returns null**

```ts
const previous = (await getWaitlistMode()) ?? 'open';
```

If `getWaitlistMode()` returns null because the mode row doesn't exist in the DB yet, the fixture assumes `'open'`. If the application treats a missing mode row as something other than open (e.g., `'invite_only'`), the restore step will set the wrong mode. Add a comment noting this assumption and verify it matches the application default in `waitlist-ops`.

tests/e2e/web/specs/waitlist/pricing-cta.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  test('pricing CTA navigates to signup with plan intent', async ({ page }) => {
+    await gotoRoute(page, '/#pricing');
+    await page
+      .getByRole('button', { name: 'Join the waitlist for Pro' })
+      .click();
+
+    await expect(page).toHaveURL(/\/signup/);
+    await expect(page.url()).toMatch(/plan=beakerstack_pro|plan=pro/i);
```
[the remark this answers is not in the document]
ZappoMan: Fixed in `53ebff8`: replaced the split `toHaveURL` + `page.url()` checks with a single `await expect(page).toHaveURL(/\/signup.*plan=(beakerstack_pro|pro)/i)`.

tests/e2e/web/fixtures/waitlist-data.fixture.ts
```diff
@@ -0,0 +1,78 @@
...
+    );
+  }
+
+  return { entryId: entry.id, inviteToken: approved.invite_token };
+}
+
+export async function deleteWaitlistEntry(entryId: string): Promise<void> {
+  const admin = createServiceRoleClient();
+  await admin.from('waitlist_invites').delete().eq('entry_id', entryId);
+  await admin.from('waitlist_entries').delete().eq('id', entryId);
```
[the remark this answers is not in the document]
ZappoMan: Fixed in `53ebff8`: `deleteWaitlistEntry` now checks `{ error }` from both `waitlist_invites` and `waitlist_entries` deletes and throws with context if either fails.

tests/e2e/web/fixtures/waitlist.fixture.ts
```diff
@@ -0,0 +1,52 @@
...
+    await setWaitlistMode(mode);
+  }
+}
+
+/** True when waitlist-capture / waitlist-ops edge functions are expected to work (preview CI). */
+export function isWaitlistEdgeReady(): boolean {
+  if (process.env.E2E_WAITLIST_READY === '1') {
+    return true;
+  }
+  if (process.env.E2E_TARGET === 'preview' || process.env.CI === 'true') {
```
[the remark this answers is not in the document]
ZappoMan: Fixed in `53ebff8`: `isWaitlistEdgeReady` now uses `Boolean(process.env.CI)` instead of `CI === 'true'`, matching `playwright.config.ts`.

tests/e2e/web/fixtures/stripe-checkout.fixture.ts
```diff
@@ -0,0 +1,197 @@
...
+  exp: '1234',
+  cvc: '123',
+  zip: '12345',
+};
+
+export function isStripeCheckoutReady(): boolean {
+  if (process.env.E2E_STRIPE_READY === '1') {
+    return true;
+  }
+  if (process.env.E2E_TARGET === 'preview' || process.env.CI === 'true') {
```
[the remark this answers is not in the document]
ZappoMan: Fixed in `53ebff8`: `isStripeCheckoutReady` now uses `Boolean(process.env.CI)` instead of a strict `'true'` string compare.

tests/e2e/web/playwright.config.ts
```diff
@@ -7,17 +7,40 @@ const isCi = Boolean(process.env.CI);
...
+  if (path.isAbsolute(configured)) {
+    return configured;
+  }
+  // CI/npm often pass repo-root paths (tests/e2e/web/results/...).
+  if (configured.startsWith('tests/')) {
+    return path.resolve(process.cwd(), configured);
+  }
+  // Paths relative to this config file (e.g. results/chromium-results.json).
+  return path.join(configDir, configured);
+}
```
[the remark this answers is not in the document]
ZappoMan: Fixed in `53ebff8`: `resolveJsonReportPath` normalizes backslashes to `/` before the `tests/` repo-root check so Windows-style `PLAYWRIGHT_JSON_REPORT` paths resolve correctly.

tests/e2e/web/run-global-setup.ts
```diff
@@ -0,0 +1,11 @@
+import type { FullConfig } from '@playwright/test';
+import globalSetup from './global-setup';
+
+async function main(): Promise<void> {
+  await globalSetup({} as FullConfig);
```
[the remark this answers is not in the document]
ZappoMan: Fixed in `53ebff8`: extracted `runE2eGlobalSetup()` from `global-setup.ts`; `run-global-setup.ts` calls that helper directly (no `FullConfig` cast).

tests/e2e/web/specs/billing/billing-stripe.spec.ts
ZappoMan: Fixed in `53ebff8`: added `resetBillingPlanForUser` in `billing.fixture.ts` and an `afterAll` in `billing-stripe.spec.ts` that upserts the seed user back to Free after Stripe flows run, so `metered-usage` free-plan limits are not affected by a scheduled downgrade still showing Pro.

tests/e2e/web/specs/admin/waitlist.spec.ts
ZappoMan: Fixed in `53ebff8`: wrapped the invite assertions in `try/finally` so `deleteWaitlistEntry` runs even when `toBeVisible` times out.

scripts/ci/format-playwright-e2e-pr-comment.mjs
ZappoMan: Fixed in `53ebff8`: removed the stale `FILE_SECTION_ROW_BG` re-export entirely — all tests already import `CATEGORY_SECTION_ROW_BG` directly.

tests/e2e/web/fixtures/stripe-checkout.fixture.ts
ZappoMan: Fixed in `53ebff8`: removed the non-iframe Stripe.js v2 fallback in `completeStripeCheckout`; we now require the hosted Checkout iframe inputs via `expect(...).toBeVisible()`.

tests/e2e/web/playwright.config.ts
ZappoMan: Fixed in `53ebff8`: restored `workers: isCi ? ciWorkers : undefined` so local runs use Playwright's default parallelism again.

tests/e2e/web/specs/billing/billing-stripe.spec.ts
ZappoMan: Fixed in `53ebff8`: reordered serial Stripe tests to upgrade → annual switch → invoices → downgrade, and added `expectCurrentPlan(page, 'Pro')` at the start of the paid-plan tests so failures are explicit if plan state is wrong.

tests/e2e/web/fixtures/waitlist-data.fixture.ts
ZappoMan: Fixed in `53ebff8`: increased `waitForWaitlistEntry` polling to 40 attempts × 500ms (20s total) for cold-start edge function latency in CI.

ZappoMan:
## Blocking review items — addressed in `05f4572`

### Sarah #1 — Chromium auth artifact verification
Added the same post-download verify step to `e2e-chromium` as `e2e-serial`: checks that `state.json`, `user.json`, `admin-state.json`, and `admin.json` exist and are non-empty before running specs.

### Mei #4 / Sarah #2 — Waitlist runtime probe (Stripe-style)
Replaced static `isWaitlistEdgeReady()` skips with a runtime probe in `waitlist.fixture.ts`:
- POST canary to `waitlist-capture`, validate JSON `ok`, clean up entry
- Invoke `waitlist-ops` validate action for unknown token
- Memoized via `resolveWaitlistEdgeReady()`; `beforeAll` + `beforeEach` skip with explicit reason in `public-signup.spec.ts` and `invite-signup.spec.ts`

### Mei #3 — Stripe `beforeAll` resilience
Wrapped `probeStripeCheckoutReady()` in try/catch so browser/probe failures downgrade to a skip with a logged reason instead of failing the serial job.

---

## Follow-up items (non-blocking)

Tracked in #370:
- Playwright cache / `--with-deps` optimization
- Explicit `PLAYWRIGHT_OUTCOME` step
- Admin waitlist cleanup via `waitForWaitlistEntry`
- Worker count tuning, merge report unit tests, category table `passedAfterRetry`, Stripe probe side-effect docs, workflow comments, warm-up curl warnings, `withSignupMode` default assumption

sarah-artificerinnovations:
## Re-review — commit 05f4572 (ZappoMan)

Both blocking items from the first review are addressed. Auth artifact verification is in the chromium job, and the probe pattern gates waitlist tests on actual service availability instead of the raw `CI=true` flag. Good call mirroring the Stripe probe pattern for waitlist. A few issues with the new `probeWaitlistEdgeReady` implementation need fixing before merge.

---

### ✅ Blocking items resolved

**Issue 1 — auth artifact verification in chromium.** Added verbatim from the serial job. Done.

**Issue 2 — `isWaitlistEdgeReady()` auto-enable.** The probe approach is the right solution: `CI=true` still enables the probe attempt, but if the probe returns `{ ready: false }`, `beforeEach` skips with a clear reason string rather than letting the test fail with an opaque upstream error. The `try/catch` added to the Stripe `beforeAll` closes the same gap there.

---

### 🔴 One issue in the new probe

**Canary cleanup is not in a `finally` block**

`probeWaitlistEdgeReady` inserts a real waitlist entry, then cleans it up inline:

```ts
const entry = await findWaitlistEntryByEmail(canaryEmail);
if (entry) {
  await deleteWaitlistEntry(entry.id);
}

const client = createWebTestClient();
const { data, error } = await client.functions.invoke('waitlist-ops', { ... });
if (error) {
  return { ready: false, reason: ... }; // ← cleanup already ran, ok here
}
```

But if `findWaitlistEntryByEmail` throws (network error, Supabase unavailable), or if `deleteWaitlistEntry` throws, execution falls into the outer `catch` block and the canary entry is never removed. On the next probe (or the next test run with the same canary email pattern), you're querying a table that may already have the entry.

Additionally, `canaryEmail` includes `Date.now()` so it's unique per run — repeated failures don't collide — but leaked entries accumulate in the waitlist table over time. Fix: wrap the capture + cleanup block in `try/finally`:

```ts
const canaryEmail = `e2e-waitlist-probe-${Date.now()}@example.com`;
let entryId: string | null = null;

try {
  // ... captureResponse check ...
  const entry = await findWaitlistEntryByEmail(canaryEmail);
  entryId = entry?.id ?? null;
  // ... waitlist-ops probe ...
  return { ready: true, reason: '...' };
} catch (error) {
  return { ready: false, reason: `...` };
} finally {
  if (entryId) {
    await deleteWaitlistEntry(entryId).catch(() => {});
  }
}
```

---

### 🟡 One design gap in the new probe

**`waitlist-ops` probe couples to a specific API contract**

The probe sends `{ action: 'validate', token: 'e2e-waitlist-probe-invalid' }` to `waitlist-ops` and then asserts:

```ts
if (typeof (data as { valid?: boolean })?.valid !== 'boolean') {
  return { ready: false, reason: 'waitlist-ops probe returned unexpected payload...' };
}
```

This means the probe returns `{ ready: false }` if `waitlist-ops` returns a 4xx status for an invalid token — which is a perfectly reasonable API behavior. If the function is deployed and working but returns `{ error: 'invalid token' }` (without a `valid` field) for a bad token, every probe returns false and every test skips, even though the function is running fine.

The probe's goal is "is the function deployed and responding?" — not "does it return the correct payload for this invalid input." A more robust probe: treat any HTTP response (including 4xx) as "the function is up," and treat only network errors or timeouts as "not ready":

```ts
const response = await fetch(`${base}/waitlist-ops`, {
  method: 'POST',
  headers: { ... },
  body: JSON.stringify({ action: 'validate', token: 'e2e-probe-invalid' }),
  signal: AbortSignal.timeout(10_000),
});
// Any response means the function is deployed. Non-2xx is expected for invalid input.
return { ready: true, reason: `waitlist-ops responded (${response.status})` };
```

Or alternatively: use a dedicated `{ action: 'health' }` call if `waitlist-ops` supports one.

---

### 🔵 Two minor items

**`@example.com` canary email domain.** Some email validation implementations reject RFC 2606 reserved domains (`example.com`). If `waitlist-capture` validates email domains, the probe will consistently fail. Use a subdomain like `e2e-probe.internal` or check whether `waitlist-capture` has a test-mode bypass.

**`getWaitlistFunctionsBaseUrl()` calls `applyE2eSupabaseEnv()` as a side effect.** Environment setup inside a URL-getter is unexpected. Move the `applyE2eSupabaseEnv()` call to the top of `probeWaitlistEdgeReady()` before the URL lookup, so the side effect is visible at the call site.

diego-artificerinnovations (changes requested):
## Re-review after 05f4572

### ✅ Addressed

- **Skip placement in `public-signup.spec.ts` and `invite-signup.spec.ts`** — skip logic moved to `beforeEach` via `testInfo.skip`, away from the mid-test position. Clean.
- **Stripe probe try/catch** — `beforeAll` probe errors now set `stripeCheckoutReady = false` rather than propagating. Correct.
- **Waitlist edge probe** — `probeWaitlistEdgeReady()` is a clean Stripe-style probe that validates both `waitlist-capture` HTTP response and `waitlist-ops` function availability before letting tests run. The memoized `resolveWaitlistEdgeReady()` avoids double-probing across spec files sharing the same worker process.

---

### ❌ Still blocking

**1. `playwright.config.ts` workers regression — not addressed**

`workers: isCi ? ciWorkers : 1` was not changed. Local dev runs the full chromium project (34 specs) with a single worker. Fix remains: `isCi ? ciWorkers : undefined`.

**2. `billing-stripe.spec.ts` serial state ordering — not addressed**

Test order is still: upgrade to Pro → downgrade → annual switch check → invoices. The annual switch test (now line ~72) runs after the downgrade test with no explicit guard that the account is still on Pro. If the billing UI suppresses cadence switching for pending-downgrade subscriptions, this test will flake.

Fix: add `await expectCurrentPlan(page, 'Pro');` at the top of the annual switch test, or move the downgrade test to last.

**3. `waitlist-data.fixture.ts` poll timeout — root cause unaddressed**

The `waitForWaitlistEntry` loop (12 × 250ms = 3s) in `waitlist-data.fixture.ts` was not changed. The new probe helps skip tests when the edge function is *down*, but does not protect against the "function up, DB write slow" scenario: the probe's `findWaitlistEntryByEmail` call is a single non-retrying lookup after `waitlist-capture` responds — if the DB write hasn't propagated yet, the probe passes and tests run, but `waitForWaitlistEntry` in the actual test still times out after 3s.

Fix: increase the poll loop to at least 40 × 500ms in `waitlist-data.fixture.ts`.

---

**Minor: probe may leave canary entries in DB**

In `probeWaitlistEdgeReady`, after a successful `waitlist-capture` POST, `findWaitlistEntryByEmail` is called once with no retry. If the write hasn't propagated yet, `entry` is null and cleanup is skipped — leaving a `e2e-waitlist-probe-*@example.com` entry in the DB. Across many CI runs this accumulates. Consider using the same retry loop (or a short timeout) before the cleanup step, or deleting by email in a follow-up query rather than by `entry.id`.

mei-artificerinnovations (approved):
## Re-review — commit 05f4572

All four blocking items are resolved. Approving.

---

**1. Billing teardown — no longer missing** ✅

`test.afterAll` calls `resetBillingPlanForUser(seed.userId)`, which does a service-role upsert to reset `billing_subscriptions` to Free. The guard (`if (!stripeCheckoutReady) return`) is correct — if the probe failed, nothing changed, so no cleanup needed. Re-runs are now safe.

One note: the teardown nulls out `stripe_customer_id` / `stripe_subscription_id` but doesn't cancel the Stripe test-mode subscription. For test mode this is acceptable (no real charges), but the orphaned subscriptions will accumulate in the Stripe test dashboard over time. Not blocking — just worth knowing if you ever debug unexpected Stripe state.

**2. Stripe iframe race condition — fixed** ✅

`await expect(numberInput).toBeVisible({ timeout: 30_000 })` replaces the old `if (await numberInput.count())` instantaneous check. No fallback path, no race. Correct.

**3. `beforeAll` probe error → test failures — fixed** ✅

`try/catch` wraps `probeStripeCheckoutReady` and sets `stripeCheckoutReady = false` on any throw. Test failures from transient network errors during the probe are now skips.

**4. `isWaitlistEdgeReady()` unconditional CI probe — fixed** ✅

`probeWaitlistEdgeReady()` does a real runtime check: canary POST to `waitlist-capture`, DB entry lookup + cleanup, and `waitlist-ops validate` ping. `resolveWaitlistEdgeReady()` memoizes the result per worker. Both `public-signup.spec.ts` and `invite-signup.spec.ts` now follow the Stripe model exactly — `resolveWaitlistEdgeReady()` in `beforeAll`, `test.beforeEach(testInfo.skip(...))` for skip propagation. Skips are now descriptive and based on actual runtime state.

---

**Small observation (non-blocking):** `probeWaitlistEdgeReady()` cleans up the canary entry with a bare `findWaitlistEntryByEmail` — same gap as non-blocking item #7 in my original review. If there's replication lag the canary entry leaks. `waitlist-data.fixture.ts` has `waitForWaitlistEntry` with a 12-attempt retry for exactly this reason. Worth using here, but not blocking.

**Bonus:** The auth artifact verification step in `e2e-chromium` (`test -s .auth/state.json` etc.) is a good hardening addition. Fails fast instead of producing misleading auth errors downstream.

Non-blocking items #5–9 from the original review remain open. None block merge.

rahul-artificerinnovations (approved):
All blocking and non-blocking items from the first review are resolved. Approved.

---

**Blocking items — resolved:**

1. ✅ `billing-stripe.spec.ts` — `afterAll` added, calls `resetBillingPlanForUser(seed.userId)` which upserts the `billing_subscriptions` row to Free plan. Cleans up Pro subscription before `metered-usage.spec.ts` runs.

2. ✅ `profile-edit.spec.ts` replaced with `profile.spec.ts` — two properly structured serial tests covering display name update and bio/website/location persistence across a reload.

**Non-blocking items — resolved:**

3. ✅ `admin/waitlist.spec.ts` first test wrapped in `try/finally` — cleanup now runs on assertion failure.

4. ✅ `FILE_SECTION_ROW_BG` re-export removed entirely.

5. ✅ Dead non-iframe fallback in `completeStripeCheckout` removed; replaced with `await expect(numberInput).toBeVisible({ timeout: 30_000 })` which is a better pattern.

**New additions look good:**

- `probeWaitlistEdgeReady()` + `resolveWaitlistEdgeReady()` mirrors the Stripe probe pattern cleanly; the canary-email cleanup inside the probe is correct (runs in the try block, failure surfaces as `ready: false`).
- Auth artifact verification step added to `e2e-chromium` job.
- Pacific timestamp footer on PR comment is a nice touch.

**One minor note on `resetBillingPlanForUser`:** the upsert sets `stripe_customer_id: null`, which orphans the Stripe test-mode customer record. This is fine for E2E test cleanup — it just means the next upgrade test creates a fresh Stripe customer. Worth a short note in the function doc so future readers don't wonder why customer ID is cleared. Not a blocker.

ZappoMan:
## Review follow-up — `7f31d61`

### @diego-artificerinnovations — your three blocking items (from `53ebff8`)

All three were already on HEAD before this commit; flagging explicitly for re-review:

| Item | Status on HEAD |
| --- | --- |
| **Workers regression** | `workers: isCi ? ciWorkers : undefined` in `playwright.config.ts` (local uses Playwright default parallelism again) |
| **billing-stripe serial ordering** | Order is upgrade → annual switch → invoices → **downgrade last**; paid-plan tests start with `expectCurrentPlan(page, 'Pro')` |
| **waitlist poll timeout** | `waitForWaitlistEntry` is 40 × 500ms (20s) in `waitlist-data.fixture.ts` |

### @sarah-artificerinnovations — re-review probe issues (`7f31d61`)

**Canary cleanup (`try/finally`)** — `probeWaitlistEdgeReady()` now tracks `canaryEntryId`, cleans up in `finally` via `deleteWaitlistEntry` or `deleteWaitlistEntryByEmail` (with retry), and logs best-effort cleanup failures instead of leaking rows.

**waitlist-ops availability check** — probe uses raw `fetch` to `waitlist-ops`; any HTTP response (including 4xx for invalid token) counts as deployed. Only network/timeout errors mark the probe not ready.

**Other minor items**
- `applyE2eSupabaseEnv()` moved to the top of `probeWaitlistEdgeReady()` (no longer hidden inside the URL getter)
- Canary email uses `generateE2ETestEmail()` (same `@example.com` pattern as the rest of the E2E suite)
- Documented load-bearing `--project=shared-state` → `--project=signup-mode` order in the `e2e-serial` workflow step

Please re-review when convenient — happy to adjust anything still blocking merge.

.github/workflows/e2e-web-pr-approval.yml
```diff
@@ -283,44 +299,307 @@ jobs:
...
+          CI: true
+          E2E_TARGET: preview
+          WEB_URL: ${{ needs.e2e-prepare.outputs.web_origin }}
+          WEB_BASE_PATH: ${{ needs.e2e-prepare.outputs.web_base_path }}
+          PREVIEW_SUPABASE_URL: ${{ secrets.PREVIEW_SUPABASE_URL }}
+          PREVIEW_SUPABASE_ANON_KEY: ${{ secrets.PREVIEW_SUPABASE_ANON_KEY }}
+          PR_TESTING_SUPABASE_SERVICE_ROLE_KEY: ${{ secrets.PR_TESTING_SUPABASE_SERVICE_ROLE_KEY }}
+          SUPABASE_ACCESS_TOKEN: ${{ secrets.SUPABASE_ACCESS_TOKEN }}
+          SUPABASE_PREVIEW_PROJECT_REF: ${{ secrets.SUPABASE_PREVIEW_PROJECT_REF }}
+        run: npm run test:e2e:web:teardown
```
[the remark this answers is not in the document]
ZappoMan: Fixed in `c1e6dd9` — `e2e-report` now downloads the `e2e-auth-state` artifact into `tests/e2e/web/.auth/` and verifies `state.json` / `admin-state.json` exist before `npm run test:e2e:web:teardown`, so seeded-user deletion and preview auth/waitlist snapshot restore run reliably.

tests/e2e/web/playwright.config.ts
```diff
@@ -31,6 +55,35 @@ export default defineConfig({
...
+      fullyParallel: false,
+      dependencies: ['chromium'],
+      use: { ...devices['Desktop Chrome'] },
+    },
+    {
+      name: 'signup-mode',
+      testMatch: [/specs\/waitlist\//, /admin\/waitlist-settings\.spec\.ts/],
+      fullyParallel: false,
+      dependencies: ['shared-state'],
       use: { ...devices['Desktop Chrome'] },
```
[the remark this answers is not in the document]
ZappoMan: Fixed in `c1e6dd9` — added `workers: 1` to the `shared-state` and `signup-mode` projects so serial specs stay single-worker even when local/CI top-level workers are higher.

docs/TESTING.md
```diff
@@ -272,7 +272,44 @@ export E2E_TARGET=preview
...
+**Waitlist edge-function E2E** (`specs/waitlist/public-signup.spec.ts` submit step, `invite-signup.spec.ts`):
+
+- Requires deployed `waitlist-capture` and `waitlist-ops` edge functions.
+- Runs automatically in CI/preview (`E2E_TARGET=preview`).
+- Local runs need `supabase functions serve` (or equivalent) and `export E2E_WAITLIST_READY=1`.
+- Without edge functions, waitlist form/mode UI specs still run; submit and invite-signup flows are skipped locally.
+
+**Shared-state specs** (`specs/profile/`, `specs/billing/metered-usage.spec.ts`, `specs/billing/billing-stripe.spec.ts`, `specs/admin/users.spec.ts`, `specs/admin/waitlist.spec.ts`) run serially in a second Playwright project to avoid conflicting edits on the seeded user.
+
+**CI parallelism:** Preview E2E runs in three jobs — prepare (global setup once), parallel `chromium` (`PLAYWRIGHT_WORKERS=4`), then a serial job for shared-state/waitlist specs. Playwright Chromium binaries are cached under `~/.cache/ms-playwright`. Local runs still use `workers: 1` unless you set `PLAYWRIGHT_WORKERS`.
```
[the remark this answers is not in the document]
ZappoMan: Fixed in `c1e6dd9` — updated the CI parallelism note: local `chromium` uses Playwright defaults; only `shared-state` / `signup-mode` pin `workers: 1` (CI chromium still uses `PLAYWRIGHT_WORKERS=4`).

docs/TESTING.md
```diff
@@ -272,7 +272,44 @@ export E2E_TARGET=preview
...
 
 **Structure:**
```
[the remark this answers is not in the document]
ZappoMan: Fixed in `c1e6dd9` — structure example now references `specs/profile/profile.spec.ts` (replacing the removed `profile-edit.spec.ts`).

tests/e2e/web/fixtures/billing.fixture.ts
```diff
@@ -0,0 +1,104 @@
...
+      stripe_price_id: null,
+      cancel_at_period_end: false,
+      canceled_at: null,
+      trial_start: null,
+      trial_end: null,
+      current_period_start: periodStart,
+      current_period_end: periodEnd,
+    },
+    { onConflict: 'user_id,product_id' }
+  );
```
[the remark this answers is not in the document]
ZappoMan: Fixed in `0493c48` — `resetBillingPlanForUser` now sets `pending_target_plan_id: null` alongside the existing scheduling clears (`cancel_at_period_end: false`, `canceled_at: null`) so a scheduled downgrade from the Stripe E2E spec does not leave the subscription in `downgrade_pending` for later tests.

sarah-artificerinnovations (approved):
## Final re-review — commits 7f31d613, 8f9a48cb, c1e6dd9f, 0493c48b (ZappoMan)

All blocking items from the first review and all three items from the re-review are resolved. Approving.

---

### ✅ All items resolved

**From the original review:**
- Auth artifact verification added to `e2e-chromium` job — done.
- `isWaitlistEdgeReady()` auto-enable handled via probe-then-skip pattern — done.

**From the re-review:**
- Canary cleanup moved into a `finally` block via `cleanupWaitlistProbeCanary` — done. The fallback to `deleteWaitlistEntryByEmail` when `canaryEntryId` is null is the right safety net.
- `waitlist-ops` probe decoupled from API response shape — now any HTTP response (including 4xx) is treated as "function is reachable." Only network errors or timeouts return `ready: false`. Correct.
- Canary email switched from `@example.com` to `generateE2ETestEmail()` — done.
- `applyE2eSupabaseEnv()` moved to top of `probeWaitlistEdgeReady` — done.
- `--no-deps` ordering comment added to the serial workflow step — done.

**Additional improvements not in my review:**
- Auth artifacts downloaded and verified in `e2e-report` before teardown runs — correct, teardown needs credentials.
- `workers: 1` explicit in serial project config — reduces reliance on the CLI `--workers=1` flag.
- `pending_target_plan_id: null` added to `resetBillingPlanForUser` — correct, scheduled downgrade state was leaking across tests.
- Docs updated to match the 4-job structure and correct test ordering.

---

### Two non-blocking notes for follow-up

**Pacific timestamp extraction is slightly fragile.** `now.toLocaleString(...).split(' ').pop()` to extract the timezone abbreviation relies on the locale string format not changing. `Intl.DateTimeFormat(...).formatToParts()` with `type === 'timeZoneName'` is more robust, but this is cosmetic.

**`FILE_SECTION_ROW_BG` deprecated alias.** Still exported. Track the callers and remove it once they're migrated — don't let "transition" mean "forever."

---

LGTM. Ready to merge.

diego-artificerinnovations (approved):
## Approved

All three blocking items are resolved. Notes on each:

**1. `playwright.config.ts` workers** → ✅ `workers: isCi ? ciWorkers : undefined` at line 39. Project-level `workers: 1` on `shared-state` and `signup-mode` is the right place to pin concurrency for those projects.

**2. `billing-stripe.spec.ts` test ordering** → ✅ Reordered to upgrade → annual switch → invoices → downgrade. `expectCurrentPlan(page, 'Pro')` guards at the top of annual switch and invoices tests make the assumption explicit and give a clear failure message if state drifts.

**3. `waitlist-data.fixture.ts` poll timeout** → ✅ `findWaitlistEntryWithRetry(email, 40, 500)` gives 20 seconds, consistent with other edge-function-backed polls in the repo. The `deleteWaitlistEntryByEmail` helper used in probe cleanup is also correct — the 8×250ms retry before deletion handles the "write hasn't propagated yet" case cleanly.

---

Two non-blocking items from my earlier reviews remain open for follow-up:

- `withSignupMode` in `waitlist.fixture.ts` is still exported but unused — safe to clean up whenever.
- `FILE_SECTION_ROW_BG` deprecated alias in `format-playwright-e2e-pr-comment.mjs` can be deleted now that `CATEGORY_SECTION_ROW_BG` is used directly in the tests.

Neither is a merge blocker. Ship it.

PR #372, opened 2026-05-24
feat: adopter-owned app zones for fork-safe customization (#369)

ZappoMan:
## Summary
- Introduces an `adopter/` zone for fork-owned branding, legal, billing, landing, dashboard UI, assets, and app-schema DB migrations, separate from template code in `apps/`, `packages/`, and `supabase/migrations/`.
- Adds `configureAdopter()` / `getAdopterConfig()` in `@beakerstack/shared` (2.0.0), extension slots for web routes and mobile screens, ESLint import boundaries, and `.gitattributes` `merge=ours` on `adopter/**` to reduce upgrade conflicts.
- Adds adopter DB tooling (`db:init-adopter`, `db:apply-adopter`), deploy workflow hooks, reusable adopter CI workflow, docs (`docs/CUSTOMIZING.md`, design notes), and test/coverage fixes including auto-migrating before pgTAP runs.

## Test plan
- [ ] `npm run type-check` passes
- [ ] `npm run test:coverage:mobile`, `npm run test:coverage:web`, and `cd packages/shared-tests && npm run test:coverage` pass
- [ ] `npm run test:db` passes (applies pending local migrations, then pgTAP)
- [ ] `node --test scripts/__tests__/db-apply-adopter.test.mjs` passes
- [ ] Web app boots with adopter branding/landing; post-login redirects to adopter dashboard
- [ ] Mobile app boots; `postLoginPathMobile` validation rejects unregistered screen names
- [ ] `npm run db:init-adopter` and `npm run db:apply-adopter -- --dry-run` behave as documented


.eslintrc.js
```diff
@@ -43,6 +55,21 @@ module.exports = {
...
   ],
   overrides: [
+    {
+      files: [
+        'apps/web/src/main.tsx',
+        'apps/web/scripts/prerender-home.ts',
+        'apps/web/src/App.tsx',
+        'apps/mobile/src/navigation/AppNavigator.tsx',
+        'apps/mobile/App.tsx',
+        'apps/web/src/**/*.{ts,tsx}',
```
sarah-artificerinnovations: 🔴 **ESLint boundary is unenforced — this glob cancels the specific-file allowlist above it**

The override lists `apps/web/src/main.tsx`, `apps/web/src/App.tsx`, `apps/mobile/src/navigation/AppNavigator.tsx`, and `apps/mobile/App.tsx` as the intended seam files — but then lines 65–66 add `apps/web/src/**/*.{ts,tsx}` and `apps/mobile/src/**/*.{ts,tsx}`, which are supersets of those specific paths. ESLint applies the first matching override, so the glob on line 65 makes all four specific file entries irrelevant: **any file in `apps/web/src/` or `apps/mobile/src/` can freely import `@adopter/*` without a lint error**.

The boundary this PR was designed to enforce is now effectively absent for the entire app source tree.

**Fix:** remove lines 65–66 and keep only the specific seam files:
```js
files: [
  'apps/web/src/main.tsx',
  'apps/web/scripts/prerender-home.ts',
  'apps/web/src/App.tsx',
  'apps/mobile/src/navigation/AppNavigator.tsx',
  'apps/mobile/App.tsx',
],
```

If tests need to import adopter config directly (e.g. for fixture setup), they're already covered by the test-file override on line 81+.

apps/web/src/App.tsx
```diff
@@ -150,7 +162,13 @@ function App() {
                       </Suspense>
                     }
                   >
-                    <Route path='/dashboard' element={<DashboardPage />} />
+                    {adopterRouteExtensions.map(extension => (
```
sarah-artificerinnovations: 🔴 **Public adopter routes are unreachable without auth — outer `<ProtectedRoute>` ignores the `auth` field**

All adopter routes land inside the `<Route element={<ProtectedRoute><Outlet /></ProtectedRoute>}>` wrapper (file line 152–156). `AdopterRoute` (line 75–85) correctly checks `extension.auth` and skips the inner `<ProtectedRoute>` for `'public'` routes — but the outer guard already redirected unauthenticated users to login before `AdopterRoute` even renders. The `auth: 'public'` extension type is therefore a broken API promise.

Currently there is only one extension (`/dashboard`, `'protected'`), so the bug is latent. But any adopter who adds a public route (landing page, marketing page, doc page) will see it silently redirect to login.

**Fix:** move adopter route rendering to a position where it can branch on `auth` before any outer guard fires:

```tsx
// Outside the ProtectedRoute wrapper, alongside the public RootLayout routes:
{adopterRouteExtensions
  .filter(ext => resolveAdopterRouteAuth(ext.auth) === 'public')
  .map(ext => (
    <Route key={ext.path} path={ext.path} element={<>{ext.element}</>} />
  ))}

// Inside the ProtectedRoute wrapper:
{adopterRouteExtensions
  .filter(ext => resolveAdopterRouteAuth(ext.auth) === 'protected')
  .map(ext => (
    <Route key={ext.path} path={ext.path} element={<ProtectedRoute>{ext.element}</ProtectedRoute>} />
  ))}
```

Or keep `AdopterRoute` as the auth decision-maker and place all adopter routes outside the outer `<ProtectedRoute>` block, so it's the sole gate.

packages/shared/src/navigation/adopterExtensions.ts
```diff
@@ -0,0 +1,21 @@
+import type { ComponentType, ReactNode } from 'react';
+
+export type AdopterRouteAuth = 'public' | 'protected';
+
+export type AdopterRouteExtension = {
+  path: string;
+  element: ReactNode;
```
sarah-artificerinnovations, line 12: 🟡 **`element: ReactNode` vs `component: ComponentType` asymmetry**

Web extensions take a pre-rendered element (an instance); mobile extensions take a component constructor. The difference is more than cosmetic:

- A `ReactNode` is created at the time the `adopterRouteExtensions` array is evaluated — i.e., at module import time. The current usage wraps in `<Suspense>` explicitly, which is fine. But the type doesn't prevent an adopter from writing `element: <DashboardPage />` without `<Suspense>`, which would break code splitting.
- A `ComponentType` is a constructor: React Router and React Native both call it lazily at render time and handle code splitting correctly.

The inconsistency will also confuse adopters building extensions. Consider aligning on `ComponentType` for web too (let the platform wire the `<Suspense>`), or at minimum document why the types differ:

```ts
export type AdopterRouteExtension = {
  path: string;
  component: ComponentType;  // aligned with mobile
  auth?: AdopterRouteAuth;
};
```

packages/shared/src/config/adopterRuntime.ts
```diff
@@ -0,0 +1,43 @@
...
+    'configureAdopter() must be called before reading adopter config'
+  );
+}
+
+export function resetAdopterConfigForTests(): void {
+  configured = null;
+  configureWarningShown = false;
+}
+
+export function ensureAdopterConfigured(config: AdopterConfig): void {
```
sarah-artificerinnovations, line 32: 🟡 **`ensureAdopterConfigured()` has opposite semantics to `configureAdopter()` with no documentation**

`configureAdopter()` warns (or throws in tests) on a second call — it treats double-configure as a programming error. `ensureAdopterConfigured()` silently no-ops if already configured — it treats double-configure as expected. A caller switching from one to the other changes behavior entirely without any type-level indication.

Without documentation, adopters will reach for whichever they see first. If the intended contract is "call `configureAdopter` exactly once at app startup, `ensureAdopterConfigured` only for lazy-init in testing helpers," that should be stated:

```ts
/**
 * Idempotent variant of configureAdopter — safe to call multiple times.
 * Use ONLY for test helpers and lazy-init scenarios where the caller
 * cannot guarantee the config has been set. Prefer configureAdopter()
 * at app startup.
 */
export function ensureAdopterConfigured(config: AdopterConfig): void {
```

Also note: `configureWarningShown` is set to `true` before the `throw` on line 9, so in test environments a caught double-call leaves `configureWarningShown = true`. Subsequent double-calls in the same module scope will then silently replace the config without throwing. This may be intentional, but it's surprising — document it or restructure so the flag is only set when the warning actually fires (i.e., move it after the `throw` branch).

apps/mobile/src/navigation/AppNavigator.tsx
```diff
@@ -23,18 +26,29 @@ export type { RootStackParamList } from './types';
...
+      `postLoginPathMobile "${config.postLoginPathMobile}" is not registered in adopter/mobile/screenExtensions.tsx`
+    );
+  }
+}
+
 export const AppNavigator = () => {
   const { showNativeHeader } = useFeatureFlags();
 
-  // Expose navigation to global scope for debugging (dev only)
+  validatePostLoginPathMobile();
```
sarah-artificerinnovations: 🟡 **`validatePostLoginPathMobile()` runs on every render**

This is called directly in the function body of `AppNavigator` — it fires on every render cycle, including re-renders triggered by unrelated state changes (feature flags, navigation events, auth state). Each call iterates `adopterStackScreens` and reads from `getAdopterConfig()`.

Since this is a programming-error guard (a misconfigured `postLoginPathMobile` should crash immediately, not on every re-render), it belongs at module evaluation time or in a one-time startup effect:

```ts
// Module level — runs once when the module is first imported:
const _registered = new Set(adopterStackScreens.map(s => s.name));
if (!_registered.has(getAdopterConfig().postLoginPathMobile)) {
  throw new Error(
    `postLoginPathMobile "${getAdopterConfig().postLoginPathMobile}" is not registered`
  );
}
```

Or move it into `App.tsx`'s startup sequence (outside the render path) so the error surfaces as a startup crash with a clear stack trace, not as a render-phase error caught by `AppErrorBoundary`.

packages/shared/src/config/adopterConfigSchema.ts
```diff
@@ -0,0 +1,33 @@
...
+  legalEntityName: z.string(),
+  contactEmail: z.string(),
+  mailingAddress: z.string(),
+});
+
+export const adopterConfigSchema = z
+  .object({
+    branding: brandingSchema,
+    legal: legalSchema,
+    postLoginPath: z.string(),
```
sarah-artificerinnovations: 🟡 **`postLoginPath` and `postLoginPathMobile` accept any string — format errors are caught late or not at all**

- `postLoginPath` should start with `/`. A value of `'dashboard'` (missing leading slash) silently fails to navigate anywhere on web.
- `postLoginPathMobile` is validated at runtime in `validatePostLoginPathMobile()` — but only when `AppNavigator` first renders, not at the schema parse boundary. A typo in `adopter/config/index.ts` won't be caught until the app starts.

Suggested additions:

```ts
postLoginPath: z.string().startsWith('/'),
postLoginPathMobile: z.string().min(1),
```

Also on `legalSchema`:
- `contactEmail: z.string().email()` — catches malformed email at config-parse time.
- `brandUrl: z.string().url()` — catches missing protocol, typos.

These are build-time constants, not user input, so stricter validation here gives a clear error message to the adopter during `configureAdopter()` rather than a broken UI later.

packages/shared/src/config/adopterConfigSchema.ts
```diff
@@ -0,0 +1,33 @@
...
+    postLoginPathMobile: z.string(),
+  })
+  .strip();
```
sarah-artificerinnovations: 🟡 **`.strip()` silently drops unknown keys**

If an adopter misspells a key — e.g. `postLoginpPath: '/dashboard'` instead of `postLoginPath` — Zod silently strips it and falls through to the `getAdopterConfig()` call, which returns the stale `configured` value (or throws "not configured" on first call). The typo is invisible until the app misbehaves.

Consider `.strict()` instead, which throws a `ZodError` listing unknown keys. If forward-compatibility requires accepting new keys without breaking old deployments, `.strip()` is the right call — but that trade-off should be documented.

scripts/upgrade-check.mjs
```diff
@@ -0,0 +1,52 @@
...
+import { fileURLToPath } from 'node:url';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const repoRoot = path.resolve(__dirname, '..');
+
+const checks = [
+  {
+    name: '.gitattributes adopter merge=ours',
+    ok: () =>
+      readFileSync(path.join(repoRoot, '.gitattributes'), 'utf8').includes(
```
sarah-artificerinnovations: 🟡 **`readFileSync` throws on a missing `.gitattributes` instead of reporting a clean failure**

If a repo doesn't yet have `.gitattributes` (pre-adopter setup, or a fresh clone that hasn't run the adopter init), this line throws `ENOENT` and the script exits with an unhandled exception rather than the clean `upgrade:check failed — …` message printed by the `failed` path.

```js
ok: () => {
  try {
    return readFileSync(path.join(repoRoot, '.gitattributes'), 'utf8')
      .includes('adopter/** merge=ours');
  } catch {
    return false;
  }
},
```

.github/workflows/adopter-tests-reusable.yml
```diff
@@ -0,0 +1,29 @@
...
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ${{ inputs.node-version }}
+          cache: npm
+
+      - run: npm ci
+
+      - name: Adopter zone checks
+        run: npm run test:adopter
```
sarah-artificerinnovations, line 27: 🔵 **`upgrade:check` is not wired into CI**

The reusable workflow runs `test:adopter` and `lint:adopter` but not `upgrade:check`. The upgrade check verifies adopter zone structural integrity (gitattributes, config files, runtime, lint script) — it should be part of the CI that adopters run after merging template updates. Consider adding:

```yaml
- name: Upgrade structure check
  run: npm run upgrade:check
```

scripts/db-apply-adopter.mjs
```diff
@@ -0,0 +1,152 @@
...
+    console.log('No pending adopter migrations');
+  }
+}
+
+function main() {
+  const linked = process.argv.includes('--linked');
+  applyAdopterMigrations({ linked });
+}
+
+if (
```
sarah-artificerinnovations: 🔵 **Fragile `isMain` guard — `.endsWith()` misfires on symlinked paths and URL-encoded characters**

`import.meta.url` is a `file://` URL; `process.argv[1]` is a filesystem path. The `.endsWith()` check passes when the path is a plain suffix of the URL, but fails when paths contain spaces (URL-encoded as `%20`), or when `process.argv[1]` is a symlink resolving to a different absolute path.

Standard pattern:
```js
import { fileURLToPath } from 'node:url';
if (process.argv[1] && fileURLToPath(import.meta.url) === path.resolve(process.argv[1])) {
  main();
}
```

Note: `db-init-adopter.mjs` calls `main()` unconditionally without any guard — if the intent is for it to also be importable as a library, it needs the same fix.

adopter/config/billing.ts
```diff
@@ -126,10 +126,17 @@ export const beakerstackBillingConfig = defineBillingConfig({
...
 
-export const BEAKERSTACK_METER_AI_SUMMARIZE = 'ai_summarize' as const;
+export const METER_AI_SUMMARIZE = 'ai_summarize' as const;
 
-export type BeakerstackFeatureKey = InferFeatureKeys<
-  typeof beakerstackBillingConfig
->;
+export type AdopterFeatureKey = InferFeatureKeys<typeof billingConfig>;
+
+/** @deprecated Use billingConfig */
```
sarah-artificerinnovations: 🔵 **Deprecated aliases with no removal tracking**

Three deprecated re-exports are introduced here (`beakerstackBillingConfig`, `BEAKERSTACK_METER_AI_SUMMARIZE`, `BeakerstackFeatureKey`). Without a tracking issue or comment stating when/what the removal condition is, "deprecated" tends to mean "forever." Consider a `TODO(#NNN)` comment pointing to a follow-up issue, or a `@since` tag with the version it was deprecated in, so it can be cleaned up in the next major template version.

sarah-artificerinnovations (dismissed):
## Code review — PR #372 (adopter-owned app zones, Phase 0)

The overall architecture is right: a first-class `adopter/` zone with Zod-validated config, typed extension slots, `merge=ours` on upgrade, and ESLint boundaries to enforce the seam. The DB tooling (`db:init-adopter`, `db:apply-adopter`) and the reusable CI workflow are clean additions. Two issues need fixing before merge.

---

### 🔴 Two blockers

**1. ESLint boundary is unenforced** (`.eslintrc.js` lines 65–66)

**2. Public adopter routes are unreachable** (`apps/web/src/App.tsx` line 165)

See inline comments for details and fixes on both.

---

### 🟡 Four design gaps

1. `AdopterRouteExtension.element` is `ReactNode`; `AdopterScreenExtension.component` is `ComponentType` — the asymmetry makes web extensions eager-by-default (the element is created at import time) and makes the type unsafe (no `<Suspense>` required by the type).

2. `ensureAdopterConfigured()` is idempotent-on-second-call while `configureAdopter()` warns/throws on second call — opposite semantics in the same file, no documentation on which to prefer.

3. `validatePostLoginPathMobile()` runs on every render of `AppNavigator`, not once at startup.

4. Schema string fields (`postLoginPath`, `contactEmail`, `brandUrl`) have no format constraints; a typo in adopter config is silently accepted by `.strip()`.

---

### 🔵 Three minor

- `upgrade-check.mjs` throws on a fresh repo instead of reporting a clean failure; `upgrade:check` is not wired into the adopter CI workflow.
- `db-apply-adopter.mjs` uses a fragile `isMain` guard that can misfire on symlinked paths.
- Deprecated billing aliases (`beakerstackBillingConfig`, etc.) carry no removal tracking.

.eslintrc.js
```diff
@@ -43,6 +55,21 @@ module.exports = {
     '*.config.ts',
   ],
   overrides: [
+    {
+      files: [
+        'apps/web/src/main.tsx',
+        'apps/web/scripts/prerender-home.ts',
+        'apps/web/src/App.tsx',
+        'apps/mobile/src/navigation/AppNavigator.tsx',
```
diego-artificerinnovations, line 66: `apps/web/src/**/*.{ts,tsx}` (and the matching mobile glob) makes the `no-restricted-imports: @adopter/*` rule a no-op for nearly every file in both apps. Only the seam entry points should be exempted:

```js
{
  files: [
    'apps/web/src/App.tsx',
    'apps/web/src/main.tsx',
    'apps/web/scripts/prerender-home.ts',
    'apps/mobile/src/navigation/AppNavigator.tsx',
    'apps/mobile/App.tsx',
    'adopter/**/*.{ts,tsx}',
  ],
  rules: { 'no-restricted-imports': 'off' },
},
```

With the current override, `apps/web/src/billing/BillingOverviewPage.tsx` can freely import `@adopter/config/billing` without a lint error — which is exactly the accidental coupling the rule is meant to prevent.

apps/mobile/src/navigation/AppNavigator.tsx
diego-artificerinnovations: `validatePostLoginPathMobile()` runs on every re-render of `AppNavigator` — it creates a `new Set(...)` and calls `getAdopterConfig()` each time. This is a side-effect in the render phase.

Since this is a startup invariant (wrong config = crash), validate once at module load instead:

```ts
// outside the component, after the function definition
validatePostLoginPathMobile();

export const AppNavigator = () => {
  const { showNativeHeader } = useFeatureFlags();
  // ...
};
```

Fails fast on load rather than on first render, and never repeats.

scripts/db-apply-adopter.mjs
```diff
@@ -0,0 +1,152 @@
...
+    } catch (error) {
+      console.error(`Failed applying ${filename}:`, error.message);
+      process.exitCode = 1;
+      break;
+    }
+  }
+
+  if (pending.length === 0) {
+    console.log('No pending adopter migrations');
+  }
```
diego-artificerinnovations, line 122: The test plan calls out `npm run db:apply-adopter -- --dry-run` but `main()` only reads `--linked`. If the documentation describes dry-run output (listing pending migrations without applying them), the flag is silently ignored.

Minimal fix:
```js
function main() {
  const linked = process.argv.includes('--linked');
  const dryRun = process.argv.includes('--dry-run');
  applyAdopterMigrations({ linked, dryRun });
}
```
And in `applyAdopterMigrations`, gate the `applyMigration` call on `!options.dryRun` and log what would run.

packages/shared/src/config/adopterRuntime.ts
```diff
@@ -0,0 +1,43 @@
...
+let configureWarningShown = false;
+
+export function configureAdopter(config: AdopterConfig): void {
+  if (configured !== null) {
+    if (!configureWarningShown) {
+      configureWarningShown = true;
+      const message =
+        'configureAdopter() called more than once; subsequent calls replace the config';
+      if (
+        typeof process !== 'undefined' &&
```
diego-artificerinnovations: `process.env?.['NODE_ENV'] === 'test'` is the env-sniffing the plan intended to avoid. In practice the risk is low — this only changes throw-vs-warn, not config values. But it can be removed cleanly:

```ts
if (configured !== null) {
  throw new Error('configureAdopter() called more than once');
}
```

Always throw. Tests already call `resetAdopterConfigForTests()` in `beforeEach`/`afterEach` via `setup.adopter.ts`, so they're never in the "configured !== null" path when `configureAdopter()` is called. Non-test environments would throw on accidental double-init, which is the correct behavior.

diego-artificerinnovations (dismissed):
## Review

The adopter zone design holds up well against the plan from issue #369. The four seams are clean (`App.tsx` / `AppNavigator.tsx` consume the extension arrays; billing and waitlist configs live in `adopter/config/`), `configureAdopter()` / `getAdopterConfig()` follow the singleton pattern, `app.schema_migrations` bootstrap is idempotent, the reusable CI workflow is a thin wrapper, and the deprecated re-exports in `billing.ts` / `waitlist.ts` give a migration path without breaking existing imports. One blocking issue on the ESLint boundary enforcement; four non-blocking items below.

---

### Blocking

**ESLint boundary override exempts all of `apps/` — see inline comment on `.eslintrc.js`**

The `no-restricted-imports` rule fires the message "Import `@adopter/*` only from extension seam files" — but the override that silences it includes `apps/web/src/**/*.{ts,tsx}` and `apps/mobile/src/**/*.{ts,tsx}`. That makes the boundary enforcement a no-op for the majority of files in both apps: any component under `apps/web/src/billing/`, `apps/web/src/components/`, etc. can import `@adopter/*` without a lint error.

The only code that actually needs the exemption is the seam entry points (`App.tsx`, `AppNavigator.tsx`). Granting it to the entire `apps/` tree defeats the rule.

---

### Non-blocking

**`validatePostLoginPathMobile()` runs on every render — see inline comment on `AppNavigator.tsx`**

Called unconditionally in the component body (render phase), it allocates a new `Set` and calls `getAdopterConfig()` on every re-render. The invariant should be checked once — at module load time, not on every render.

**`--dry-run` not parsed in `db-apply-adopter.mjs` — see inline comment**

The test plan calls out `npm run db:apply-adopter -- --dry-run`; `main()` only reads `--linked`. If the docs describe dry-run output, the flag is silently ignored.

**`configureAdopter()` env-sniffing** — see inline comment on `adopterRuntime.ts`

The `NODE_ENV === "test"` check determines throw-vs-warn on a second `configureAdopter()` call. The design intent from #369 was no env-sniffing. The risk is low (this only controls error behavior, not config values), but it can be removed: always throw on the second call in all environments, and rely on `resetAdopterConfigForTests()` between tests. That's already what `setup.adopter.ts` does.

**`import.meta.url.endsWith()` check in `db-apply-adopter.mjs`**

The "am I the main module?" guard uses `endsWith(process.argv[1])` which can false-positive if two scripts share a suffix. The rest of the repo uses `import.meta.url === url.pathToFileURL(process.argv[1] || '').href`.

packages/shared/src/config/adopterConfigSchema.ts
rahul-artificerinnovations: **Non-blocking: bare `z.string()` permits empty strings and absolute URLs.**

`postLoginPath: z.string()` passes Zod validation with an empty string (silent redirect to current page) or an absolute URL like `https://evil.com` (open redirect — developer-error class, not user-exploitable, but easy to prevent). Same applies to `postLoginPathMobile` on line 27.

Suggested fix:
```ts
postLoginPath: z.string().min(1).refine(s => s.startsWith('/'), {
  message: 'postLoginPath must be a relative path starting with /'
}),
postLoginPathMobile: z.string().min(1),
```
The relative-path constraint on `postLoginPath` eliminates the open-redirect class of misconfiguration at configuration time rather than at runtime.

.github/workflows/adopter-tests-reusable.yml
```diff
@@ -0,0 +1,29 @@
...
+
+      - name: Lint adopter tree
+        run: npm run lint:adopter
```
rahul-artificerinnovations, line 33: **Non-blocking: reusable workflow has no visible caller.**

This workflow (`on: workflow_call`) is defined but I don't see it called from `deploy-production.yml`, `deploy-staging.yml`, or a PR CI workflow in this diff. If there is no calling workflow, `npm run test:adopter` and `npm run lint:adopter` don't run in CI and adopter-zone regressions won't be caught automatically.

If a main CI workflow (e.g., `ci.yml` or `e2e-web-pr-approval.yml`) calls this, please add a comment here or in the docs pointing to the caller. If it hasn't been wired yet, that's the missing piece before adopter CI coverage is actually active.

scripts/db-apply-adopter.mjs
```diff
@@ -0,0 +1,152 @@
...
+    .map(line => line.trim())
+    .filter(Boolean);
+}
+
+export function pendingMigrations(allFiles, applied) {
+  const appliedSet = new Set(applied);
+  return allFiles.filter(file => !appliedSet.has(file));
+}
+
+function resolveDatabaseUrl(linked) {
```
rahul-artificerinnovations: **Non-blocking: `resolveDatabaseUrl` is duplicated verbatim in `db-init-adopter.mjs`.**

Both scripts share the same three-path resolution logic (`DATABASE_URL` → `supabase db remote connection-string` → local default). If the `--linked` error handling or local default URL ever changes in one file, it's easy for the two to diverge silently.

Consider extracting to a shared utility, e.g. `scripts/lib/resolve-db-url.mjs`, so both scripts import it:
```js
export function resolveAdopterDatabaseUrl({ linked } = {}) { ... }
```
`db-init-adopter.mjs` could then simply call `applyAdopterMigrations({ linked })` and be reduced to a one-line wrapper, since `applyAdopterMigrations` already calls `ensureInit` internally.

rahul-artificerinnovations (approved):
Phase 0 is solid. The adopter runtime singleton (`configureAdopter` / `getAdopterConfig` / `resetAdopterConfigForTests`), extension-slot types, Zod schema with `.strip()`, `.gitattributes` `merge=ours`, and the adopter DB migration runner all land correctly. The ESLint import boundary — restricting `packages/` from importing `@adopter/*` while allowing `apps/` — is the right scope for a monorepo where app-level code is intentionally fork-owned. No blocking items.

Five non-blocking notes below.

---

**1. `postLoginPath` and `postLoginPathMobile` Zod schemas accept empty strings** — see inline comment on `adopterConfigSchema.ts`.

**2. `validatePostLoginPathMobile()` fires on every render of `AppNavigator`**

```tsx
export const AppNavigator = () => {
  const { showNativeHeader } = useFeatureFlags();
  validatePostLoginPathMobile(); // ← runs on every re-render
```

This allocates a new `Set` from `adopterStackScreens` and calls `getAdopterConfig()` on every render cycle. In React Strict Mode (dev), it runs twice per mount. The validation itself is correct and useful — the concern is placement: if `configureAdopter()` hasn't been called before `AppNavigator` renders (e.g., a future test that forgets setup), the throw lands inside a React render, which React catches and hands to the nearest error boundary rather than failing clearly as a setup error.

Since the config is a stable singleton that won't change between renders, validate once at app boot instead:
```tsx
// apps/mobile/App.tsx, before mounting <AppNavigator>
configureAdopter(adopterConfig);
validatePostLoginPathMobile(); // once, before React mounts
```
Or guard with `useMemo` / `useRef` inside the component so the check runs once per mount, not per render.

**3. `adopter-tests-reusable.yml` is defined but has no known caller** — see inline comment on the workflow file.

**4. `resolveDatabaseUrl` logic is duplicated between `db-apply-adopter.mjs` and `db-init-adopter.mjs`** — see inline comment on `db-apply-adopter.mjs`.

**5. `screen.name as keyof RootStackParamList` is an unsafe cast**

In `AppNavigator.tsx`:
```tsx
<Stack.Screen
  name={screen.name as keyof RootStackParamList}
```

`RootStackParamList` was defined with the original hardcoded screens (`Home`, `Login`, `Dashboard`, etc.). When an adopter adds a new screen whose name is NOT already in `RootStackParamList` — say `'Settings'` — the cast lies to TypeScript and `navigate('Settings')` call sites won't type-check. Consider documenting that `RootStackParamList` must be augmented when new adopter screens are added, or use module augmentation so adopters can extend it:
```ts
// adopter/mobile/screenExtensions.tsx
declare module '@beakerstack/mobile/navigation/types' {
  interface AdopterScreens {
    Dashboard: undefined;
  }
}
```
This is a Phase 0 limitation worth tracking as a Phase 1 AC.

.eslintrc.js
mei-artificerinnovations: These two lines (`apps/web/src/**/*.{ts,tsx}` and `apps/mobile/src/**/*.{ts,tsx}`) turn off the `@adopter/*` boundary for every file in both app trees — only `packages/**` remains guarded. The boundary Phase 0 is establishing does not exist for the apps themselves.

Remove lines 65–66. The four named seam files on lines 60–64 are already listed explicitly; the broad globs are not needed and they defeat the entire architectural constraint.

.eslintrc.js
```diff
@@ -43,6 +55,21 @@ module.exports = {
...
+        'apps/mobile/App.tsx',
+        'apps/web/src/**/*.{ts,tsx}',
+        'apps/mobile/src/**/*.{ts,tsx}',
```
mei-artificerinnovations: Together with line 65, this disables `no-restricted-imports` for `@adopter/*` across the entire `apps/` tree. Any template component in either app can freely import adopter config and ESLint will not flag it. The boundary only exists for `packages/**`.

apps/web/src/App.tsx
mei-artificerinnovations: These adopter routes are nested inside `<ProtectedRoute><Outlet /></ProtectedRoute>` at line 153. The outer `ProtectedRoute` redirects unauthenticated users before `AdopterRoute` ever runs — so `auth: 'public'` extensions are silently auth-gated regardless of the `auth` field.

The `AdopterRoute` component (lines 75–85) correctly handles public vs. protected logic, but the ancestor guard fires first. Split the map: render `auth: 'public'` and `auth: 'unauthenticated'` extensions outside the `ProtectedRoute` wrapper alongside the existing public routes, the same way `LoginPage`/`SignupPage` are registered.

packages/shared/src/config/adopterRuntime.ts
```diff
@@ -0,0 +1,43 @@
...
+        process.env?.['NODE_ENV'] === 'test'
```
mei-artificerinnovations: `NODE_ENV === 'test'` is the same env-sniffing pattern the Phase 0 design doc eliminated (see also the Round 2 re-review on issue #369). Production code should not branch on test-environment detection — a stale `NODE_ENV=test` in a staging runner now throws on double-call instead of warning.

Simpler fix: always throw on double-call. `setup.adopter.ts` calls `resetAdopterConfigForTests()` in `beforeEach` before every test, so a test that double-calls without a reset is a real bug that should be caught. Production `main.tsx` will never double-call `configureAdopter()`, so the warn path is not needed.

apps/web/src/main.tsx
```diff
@@ -4,14 +4,17 @@ import {
...
 } from '@beakerstack/observability/web';
+import { ensureAdopterConfigured } from '@beakerstack/shared/config/adopterRuntime';
+import { adopterConfig } from '@adopter/config';
 import { beakerstackObservabilityConfig } from './config/observability';
 import { setupLogging } from '@beakerstack/logger';
 import { ThemeProvider } from './contexts/ThemeContext';
 import { PublicShell } from './PublicShell';
 import './index.css';
 
+ensureAdopterConfigured(adopterConfig);
```
mei-artificerinnovations: `ensureAdopterConfigured` silently no-ops if the config is already set. The canonical plan specified `configureAdopter()` here — the assertion variant — because `main.tsx` is the single bootstrap entry point and should fail loudly if something else already configured it.

`prerender-home.ts` correctly calls `configureAdopter(adopterConfig)` directly. This should match.

apps/mobile/src/navigation/AppNavigator.tsx
mei-artificerinnovations: `validatePostLoginPathMobile()` throws synchronously in the render body. If `postLoginPathMobile` is not registered in `adopterStackScreens`, React Native crashes the navigator on every render attempt rather than failing once at startup.

Move this call to `apps/mobile/App.tsx` before mounting the navigator — the same location `ensureAdopterConfigured` runs — so a misconfigured adopter gets a clear startup error instead of a mid-render crash.

packages/shared-tests/setup.adopter.ts
```diff
@@ -0,0 +1,14 @@
+import {
+  configureAdopter,
+  resetAdopterConfigForTests,
+} from '@beakerstack/shared/config/adopterRuntime';
+import { adopterConfig } from '../../adopter/config/index';
```
mei-artificerinnovations: Relative path across a package boundary. Use the tsconfig alias:

```ts
import { adopterConfig } from '@adopter/config';
```

The `@adopter/config` path is registered in tsconfig and Vite/Metro config specifically so cross-package imports use stable aliases that survive directory moves.

mei-artificerinnovations (dismissed):
## PR #372 — Adopter-owned app zones Phase 0

Good foundation overall. The runtime singleton design, Zod `.strip()`, extension slot types, and DB isolation approach all land correctly. Three blocking issues before merge, four non-blocking items worth tracking.

---

### Blocking

#### 1. ESLint boundary completely bypassed by app-tree globs (`.eslintrc.js` lines 65–66)

The seam override that turns off `no-restricted-imports` for `@adopter/*` includes two broad globs:

```js
'apps/web/src/**/*.{ts,tsx}',
'apps/mobile/src/**/*.{ts,tsx}',
```

This disables the boundary for **every file** in both app trees — only `packages/**` is left guarded. The boundary that Phase 0 was designed to establish doesn't exist for the apps themselves. Any template component in `apps/web/src/` can freely import `@adopter/*` and ESLint won't say a word.

Fix: remove lines 65–66. The four named seam files (`main.tsx`, `prerender-home.ts`, `App.tsx`, `AppNavigator.tsx`, `App.tsx` mobile) are already listed explicitly on lines 60–64 and 67.

#### 2. `auth: 'public'` adopter routes always require login (`App.tsx` lines 152–165)

All adopter routes are rendered as children of `<ProtectedRoute><Outlet /></ProtectedRoute>` (line 152) **and** inside `<BillingProviderLayout>` (lines 158–164). The outer `ProtectedRoute` redirects unauthenticated users before `AdopterRoute`'s own auth check ever runs. `AdopterRouteExtension.auth: 'public'` is silently ignored for every route.

This means the `auth` field on `AdopterRouteExtension` is currently a no-op for the web router. An adopter shipping a `auth: 'public'` marketing or onboarding page will get a silent redirect to login.

Fix: split the `adopterRouteExtensions.map()` call — render extensions with `auth: 'public'` (and `auth: 'unauthenticated'`) outside the `ProtectedRoute` wrapper, the same way the existing public landing and auth routes are registered. The `AdopterRoute` component can own the protection logic for its own slot.

#### 3. `configureAdopter()` branches on `NODE_ENV === 'test'` (`adopterRuntime.ts` lines 12–16)

```ts
if (typeof process !== 'undefined' && process.env?.['NODE_ENV'] === 'test') {
  throw new Error(message);
}
console.warn(message);
```

This is the same env-sniffing anti-pattern that the canonical plan's Phase 0 design doc explicitly eliminated. `NODE_ENV === 'test'` has the same risks as `VITEST`: a staging runner with stale env vars now throws; tests never exercise the warn path that production will see.

The fix is simpler: always throw on double-call. `setup.adopter.ts` calls `resetAdopterConfigForTests()` in `beforeEach`, so any test that double-calls without a reset is a test bug that should be caught. Production code will never double-call `configureAdopter()` from `main.tsx`. The warn-then-replace behavior for production isn't needed — it just masks a real bug.

---

### Non-blocking

#### 4. `main.tsx` calls `ensureAdopterConfigured` instead of `configureAdopter` (line 15)

The canonical plan specified `configureAdopter()` before `createRoot`. The soft variant (`ensureAdopterConfigured`) silently no-ops if the config is already set, hiding any double-init path. Since `main.tsx` is the single bootstrap entry point, it should assert rather than quietly skip. This is also inconsistent with `prerender-home.ts`, which correctly calls `configureAdopter(adopterConfig)` directly.

#### 5. `validatePostLoginPathMobile()` called in render body (`AppNavigator.tsx` line 50)

This throws synchronously during React Native render. If `postLoginPathMobile` isn't registered, React crashes the navigator on every render attempt rather than failing cleanly at startup. Move the call to `apps/mobile/App.tsx` before mounting the navigator — same location that `ensureAdopterConfigured` runs for web.

#### 6. `setup.adopter.ts` uses relative import across package boundary (line 5)

```ts
import { adopterConfig } from '../../adopter/config/index';
```

The `@adopter/config` tsconfig path alias exists precisely for this. The relative path crosses a package boundary and will silently break if either package moves. Use `import { adopterConfig } from '@adopter/config'`.

#### 7. `app.schema_migrations` diverges from accepted plan

The Round 2 re-review confirmed `adopter.schema_migrations` as the tracking table. The code creates `app.schema_migrations` instead (both in `adopter/db/init.sql` and in the `SELECT` in `db-apply-adopter.mjs`). If the decision changed to `app` schema, that should be noted in `adopter/db/README.md` rather than left as an undocumented deviation from the design doc.

---

### Summary

| Issue | Severity |
|---|---|
| ESLint boundary bypassed — broad app-tree globs in seam override | Blocking |
| `auth: 'public'` adopter routes always require login | Blocking |
| `configureAdopter` branches on `NODE_ENV === 'test'` | Blocking |
| `main.tsx` calls soft `ensureAdopterConfigured` instead of `configureAdopter` | Non-blocking |
| `validatePostLoginPathMobile` called in render body | Non-blocking |
| `setup.adopter.ts` relative import across package boundary | Non-blocking |
| `app.schema_migrations` vs accepted plan's `adopter.schema_migrations` | Non-blocking |

apps/web/src/App.tsx
```diff
@@ -150,7 +162,13 @@ function App() {
...
+                      <Route
+                        key={extension.path}
+                        path={extension.path}
+                        element={<AdopterRoute extension={extension} />}
+                      />
+                    ))}
```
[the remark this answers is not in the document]
ZappoMan: Fixed in `f39068a`. Split adopter routes in `apps/web/src/App.tsx`: `publicAdopterRouteExtensions` render outside the outer `ProtectedRoute` (lines 158–164); `protectedAdopterRouteExtensions` stay inside the ProtectedRoute + BillingProviderLayout branch. `auth: 'public'` extensions are now reachable without authentication.

.eslintrc.js
```diff
@@ -43,6 +55,21 @@ module.exports = {
...
+        'adopter/**/*.{ts,tsx}',
+      ],
+      rules: {
+        'no-restricted-imports': 'off',
+      },
+    },
```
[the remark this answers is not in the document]
ZappoMan: Fixed in `f39068a`. Global `no-restricted-imports` now blocks only `@adopter/web/*` and `@adopter/mobile/*` (not all `@adopter/*`). Override allowlist is limited to explicit seam files, billing integration pages, and `adopter/**` — see `.eslintrc.js`.

docs/design/adopter-zones/db-migrations.md
```diff
@@ -0,0 +1,66 @@
...
+```bash
+npm run db:init-adopter
+```
+
+## `db:apply-adopter` algorithm
+
+1. List `adopter/db/migrations/*.sql` sorted lexicographically
+2. Query `adopter.schema_migrations` for applied filenames
+3. Apply each pending file in a transaction; record filename + checksum on success
+4. On failure: abort, leave DB at last good migration, print failed file + error
```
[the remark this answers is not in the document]
ZappoMan: Fixed in `f39068a`. Updated `docs/design/adopter-zones/db-migrations.md` to reference `app.schema_migrations`, matching `adopter/db/init.sql`.

scripts/upgrade-check.mjs
```diff
@@ -0,0 +1,52 @@
...
+        'adopter/** merge=ours'
+      ),
+  },
```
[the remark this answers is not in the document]
ZappoMan: Fixed in `f39068a`. Added `readGitattributes()` with try/catch in `scripts/upgrade-check.mjs` — a missing/unreadable `.gitattributes` is reported as a failed check instead of crashing the script.

.eslintrc.js
ZappoMan: Fixed in `f39068a`. Removed the broad `apps/web/src/**/*` and `apps/mobile/src/**/*` globs. The override now lists only explicit seam files (`main.tsx`, `App.tsx`, `AppNavigator.tsx`, billing integration pages, `prerender-home.ts`) plus `adopter/**`.

apps/web/src/App.tsx
ZappoMan: Fixed in `f39068a`. Public adopter routes are rendered outside the outer `ProtectedRoute` wrapper; protected routes remain inside it. See `publicAdopterRouteExtensions` / `protectedAdopterRouteExtensions` in `apps/web/src/App.tsx`.

packages/shared/src/navigation/adopterExtensions.ts
ZappoMan: Addressed in `f39068a`. Added JSDoc on `AdopterRouteExtension` and `AdopterScreenExtension` in `packages/shared/src/navigation/adopterExtensions.ts` explaining why web uses `ReactNode` (Suspense wired at the registry) vs mobile `ComponentType`.

packages/shared/src/config/adopterRuntime.ts
ZappoMan: Fixed in `f39068a`. Added JSDoc on `ensureAdopterConfigured()` in `packages/shared/src/config/adopterRuntime.ts`. Also removed `configureWarningShown` and `NODE_ENV` branching — `configureAdopter()` always throws on double-call.

apps/mobile/src/navigation/AppNavigator.tsx
ZappoMan: Fixed in `f39068a`. Exported `validatePostLoginPathMobile()` and call it once at startup in `apps/mobile/App.tsx` (after `configureAdopter()`). Removed the call from `AppNavigator`'s render body.

packages/shared/src/config/adopterConfigSchema.ts
ZappoMan: Fixed in `f39068a`. `postLoginPath` now uses `.min(1)` + `.refine(s => s.startsWith('/'))`; `postLoginPathMobile` uses `.min(1)`. See `packages/shared/src/config/adopterConfigSchema.ts`.

packages/shared/src/config/adopterConfigSchema.ts
ZappoMan: Fixed in `f39068a`. Changed `adopterConfigSchema` from `.strip()` to `.strict()` so misspelled/unknown keys fail validation at configure time.

scripts/upgrade-check.mjs
ZappoMan: Fixed in `f39068a`. Same as the Copilot thread — `readGitattributes()` in `scripts/upgrade-check.mjs` catches read errors and treats them as a failed check.

.github/workflows/adopter-tests-reusable.yml
ZappoMan: Fixed in `f39068a`. Added `npm run upgrade:check` to `.github/workflows/adopter-tests-reusable.yml`. That workflow is now called from `.github/workflows/test.yml` via the `adopter-checks` job.

scripts/db-apply-adopter.mjs
ZappoMan: Fixed in `f39068a`. Replaced fragile `.endsWith()` `isMain` guard with `isScriptMain()` using `path.resolve(fileURLToPath(import.meta.url))` in `scripts/lib/resolve-adopter-database-url.mjs`, used by `db-apply-adopter.mjs`.

adopter/config/billing.ts
ZappoMan: Fixed in `f39068a`. Deprecated billing aliases in `adopter/config/billing.ts` now include a removal note: "remove in the next template major (2027.x)".

.eslintrc.js
ZappoMan: Fixed in `f39068a`. Removed the app-tree globs that disabled the boundary for all of `apps/web/src` and `apps/mobile/src`. Only explicit seam files + billing integration pages + `adopter/**` are allowlisted in `.eslintrc.js`.

apps/mobile/src/navigation/AppNavigator.tsx
ZappoMan: Fixed in `f39068a`. `validatePostLoginPathMobile()` runs once at startup in `apps/mobile/App.tsx`, not on every `AppNavigator` re-render.

scripts/db-apply-adopter.mjs
ZappoMan: Fixed in `f39068a`. `main()` reads `--dry-run`; `applyAdopterMigrations()` logs `Would apply …` and skips `applyMigration` when dry-run is set. See `scripts/db-apply-adopter.mjs`.

packages/shared/src/config/adopterRuntime.ts
ZappoMan: Fixed in `f39068a`. Removed `process.env.NODE_ENV === 'test'` branching. `configureAdopter()` always throws on double-call; tests use `resetAdopterConfigForTests()` in `setup.adopter.ts`.

packages/shared/src/config/adopterConfigSchema.ts
ZappoMan: Fixed in `f39068a`. Added `.min(1)` and a leading-`/` refine on `postLoginPath`, plus `.min(1)` on `postLoginPathMobile`, in `packages/shared/src/config/adopterConfigSchema.ts`.

.github/workflows/adopter-tests-reusable.yml
ZappoMan: Fixed in `f39068a`. Wired the reusable workflow from `.github/workflows/test.yml` (`adopter-checks` job, gated on `run_unit`). Added a comment at the top of `adopter-tests-reusable.yml` pointing to the caller; added `adopter/**` to `test.yml` path triggers.

scripts/db-apply-adopter.mjs
ZappoMan: Fixed in `f39068a`. Extracted shared `resolveAdopterDatabaseUrl()` and `isScriptMain()` to `scripts/lib/resolve-adopter-database-url.mjs`. Both `db-apply-adopter.mjs` and `db-init-adopter.mjs` import it.

.eslintrc.js
ZappoMan: Fixed in `f39068a`. Removed `apps/web/src/**/*` and `apps/mobile/src/**/*` from the ESLint override — the boundary is now enforced across the app trees except for the named seam files.

.eslintrc.js
ZappoMan: Fixed in `f39068a`. Same fix — both broad globs removed from `.eslintrc.js`; only explicit seam files remain allowlisted.

apps/web/src/App.tsx
ZappoMan: Fixed in `f39068a`. Split adopter route rendering in `apps/web/src/App.tsx`: public/unauthenticated extensions register alongside login/signup; protected extensions stay under the ProtectedRoute branch.

packages/shared/src/config/adopterRuntime.ts
ZappoMan: Fixed in `f39068a`. `configureAdopter()` always throws on double-call — no `NODE_ENV` sniffing. See `packages/shared/src/config/adopterRuntime.ts`.

apps/web/src/main.tsx
ZappoMan: Fixed in `f39068a`. `apps/web/src/main.tsx` now calls `configureAdopter(adopterConfig)` to match `prerender-home.ts` and fail loudly on accidental double-init.

apps/mobile/src/navigation/AppNavigator.tsx
ZappoMan: Fixed in `f39068a`. Moved `validatePostLoginPathMobile()` to `apps/mobile/App.tsx` startup (before mounting the navigator) so misconfiguration fails once at boot, not mid-render.

packages/shared-tests/setup.adopter.ts
ZappoMan: Fixed in `f39068a`. `packages/shared-tests/setup.adopter.ts` now imports `adopterConfig` from `@adopter/config` instead of a relative path.

PR #375, opened 2026-05-24
Fix staging adopter DB deploy (--linked)

ZappoMan:
## Summary
- Fix `db:apply-adopter --linked` failing in staging/production deploys because the resolver called a non-existent Supabase CLI command (`supabase db remote connection-string`), which returned help text and caused `psql` to connect to a local socket.
- Build the remote Postgres URI from existing CI secrets (`STAGING_SUPABASE_*`, `PRODUCTION_SUPABASE_*`, etc.) via `postgresConnectionUri`, with URL validation before calling `psql`.
- Add unit tests and update adopter DB docs to describe env-based `--linked` resolution.

## Test plan
- [x] `node --test scripts/__tests__/resolve-adopter-database-url.test.mjs`
- [x] `npm run test:adopter`
- [ ] Re-run Deploy Staging workflow and confirm `Apply adopter database migrations` succeeds


scripts/lib/resolve-adopter-database-url.mjs
```diff
@@ -1,44 +1,113 @@
...
+export function resolveLinkedCredentials(env = process.env) {
+  const pairs = [
+    ['SUPABASE_PROJECT_REF', 'SUPABASE_DB_PASSWORD'],
+    ['STAGING_SUPABASE_PROJECT_REF', 'STAGING_SUPABASE_DB_PASSWORD'],
+    ['PRODUCTION_SUPABASE_PROJECT_REF', 'PRODUCTION_SUPABASE_DB_PASSWORD'],
+  ];
+
+  for (const [refKey, passwordKey] of pairs) {
+    const projectRef = env[refKey]?.trim();
+    const dbPassword = env[passwordKey];
```
mei-artificerinnovations: `projectRef` is trimmed (line 43) but `dbPassword` is not. A trailing newline in a copy-pasted CI secret would get URL-encoded via `encodeURIComponent` and silently cause auth failures. `env[passwordKey]?.trim()` here would be consistent.

scripts/lib/resolve-adopter-database-url.mjs
```diff
@@ -1,44 +1,113 @@
...
+  if (!existsSync(refPath)) {
+    return undefined;
+  }
+
+  const projectRef = readFileSync(refPath, 'utf8').trim();
+  if (!projectRef) {
+    return undefined;
+  }
+
+  const dbPassword = env.SUPABASE_DB_PASSWORD;
```
mei-artificerinnovations: The file-based fallback only reads `SUPABASE_DB_PASSWORD` (not `STAGING_SUPABASE_DB_PASSWORD` or `PRODUCTION_SUPABASE_DB_PASSWORD`). This is probably correct — `supabase link` writes the project-ref file and uses generic vars — but a short comment here would prevent a future contributor from being confused when staging/prod passwords are set but this path still returns `undefined`.

mei-artificerinnovations (approved):
## PR #375 — Fix staging adopter DB deploy (--linked)

Clean, targeted fix. The root cause (non-existent `supabase db remote connection-string` command returning help text, causing `psql` to try a local socket) is correctly identified and resolved. The env-based credential resolution is a more reliable approach that doesn't depend on the Supabase CLI being linked at deploy time.

Specific things that look good:

- `assertPostgresUrl` validation before passing to `psql` — catches help-text output and other non-URL values before they silently connect somewhere unexpected.
- `env` injected as a parameter rather than reading from `process.env` directly — makes unit testing honest and avoids global side effects in tests.
- `postgresConnectionUri` uses `encodeURIComponent`, so passwords with `@`, `&`, or other URI-special characters (confirmed by the `p@ss&word` test case) are handled correctly.
- Priority order (generic → staging → production → file fallback) is documented in the README.
- Test coverage is comprehensive: all three credential tiers, `DATABASE_URL` override, invalid URL rejection, file-based fallback, and local defaults.

Two minor observations (neither blocking):

1. **`dbPassword` is not trimmed** (line 44) while `projectRef` is (line 43). A trailing newline in a CI secret value would get URL-encoded and cause auth failures. `env[passwordKey]?.trim()` would be consistent with how `projectRef` is handled.

2. **`readLinkedProjectRef` only reads `SUPABASE_DB_PASSWORD`** (line 69), not the staged/prod variants. This is probably intentional — the file path is written by `supabase link` which uses the generic vars — but worth a short comment so contributors don't try to debug why it doesn't pick up `STAGING_SUPABASE_DB_PASSWORD` in the fallback path.

LGTM.

scripts/lib/resolve-adopter-database-url.mjs
diego-artificerinnovations: `projectRef` is trimmed on line 43 but `dbPassword` is not. If a secret value has trailing whitespace, psql will reject auth with a generic "authentication failed" error — nothing points at whitespace.

```js
const dbPassword = env[passwordKey]?.trim();
```

scripts/lib/resolve-adopter-database-url.mjs
diego-artificerinnovations: `readLinkedProjectRef` reads only `SUPABASE_DB_PASSWORD` for the file-based fallback, not the staging/production variants. This is probably intentional (local `supabase link` → generic password), but it creates an asymmetry: `resolveLinkedCredentials` will find staging/production credentials in CI, but if the project-ref file is the only path available and `STAGING_SUPABASE_DB_PASSWORD` is set instead of `SUPABASE_DB_PASSWORD`, this returns `undefined` and the caller throws.

If the intent is "file fallback is local-dev only, always uses generic password", a short comment here or a note in the README's priority list (step 5) would make that clear.

diego-artificerinnovations (approved):
## Approved

The fix is correct. Removing `spawnSync('supabase', ['db', 'remote', 'connection-string'])` was the right call — that command doesn't exist, returns help text, and psql silently interprets the resulting string as a socket path. The new approach builds the URL directly from CI secrets, `assertPostgresUrl` guards against invalid values, and the test suite covers the priority chain thoroughly (DATABASE_URL → SUPABASE_* → STAGING_SUPABASE_* → PRODUCTION_SUPABASE_* → file fallback). No blocking issues.

---

### Non-blocking

**`dbPassword` not trimmed in `resolveLinkedCredentials` — see inline comment on `resolve-adopter-database-url.mjs`**

`projectRef` is trimmed on line 43 but `dbPassword` is not on line 44. If a secret has trailing whitespace (possible with `.env` file sourcing, some CI secret configurations, or copy-paste), psql auth fails with a cryptic "authentication failed" message that doesn't point at whitespace. Should be `env[passwordKey]?.trim()`.

**`readLinkedProjectRef` only checks `SUPABASE_DB_PASSWORD` — see inline comment**

The local-link file fallback reads `env.SUPABASE_DB_PASSWORD` directly (line 69) rather than going through `resolveLinkedCredentials`. This means someone running locally after `supabase link` who only has `STAGING_SUPABASE_DB_PASSWORD` set will hit "Failed to resolve" even with the project-ref file present. Probably intentional (local dev → generic password), but the README's priority-order list doesn't call out this constraint for step 5.

scripts/lib/resolve-adopter-database-url.mjs
rahul-artificerinnovations: **Non-blocking: missing-password case silently falls through to the generic error.**

When `supabase/.temp/project-ref` exists but `SUPABASE_DB_PASSWORD` is not set, the function returns `undefined` and the caller falls through to:

```
Failed to resolve remote database URL for --linked. Set DATABASE_URL or project ref + DB password env vars...
```

This is accurate but doesn't tell the developer *why* the file-based path failed — they have the project-ref file from `supabase link`, so they'll be confused why it didn't work. A small diagnostic helps:

```js
const dbPassword = env.SUPABASE_DB_PASSWORD;
if (!dbPassword) {
  // project-ref exists but no password — inform instead of silently returning undefined
  return undefined; // or: throw with a targeted message if you'd prefer to fail fast
}
```

Alternatively, bubble a targeted error from `readLinkedProjectRef` itself when the file is found but the password is absent. Non-blocking — the generic error is still correct and actionable.

rahul-artificerinnovations (approved):
Clean fix for a real CI breakage. The root cause — `supabase db remote connection-string` being a non-existent CLI command that returns help text, causing psql to fall back to a local socket — is clearly identified and the replacement is the right approach: build the URI directly from the CI secrets that were already documented as prerequisites for `--linked`.

Specific things I like:
- **`postgresConnectionUri` handles URL-encoding**: the test proves `p@ss&word` → `p%40ss%26word`, so passwords with special characters are safe.
- **`assertPostgresUrl` on every return path**: validates before handing a URL to psql on all code paths (explicit `DATABASE_URL`, env-var-built URI, file-based URI, and the local `SUPABASE_DB_URL` path). This is an improvement over the old code which returned `DATABASE_URL` raw without validation.
- **Test coverage**: 12 test cases covering all priority tiers, the `DATABASE_URL` override, invalid-URL rejection, missing-credentials error, and both local fallback paths.
- **No callers need to change**: `db-apply-adopter.mjs` and `db-init-adopter.mjs` call `resolveAdopterDatabaseUrl()` without arguments; the new `env` parameter defaults to `process.env` so existing callers work unchanged.

Three non-blocking notes below.

---

**1. `readLinkedProjectRef` silently swallows the missing-password case** — see inline comment.

**2. `assertPostgresUrl` error message omits the bad value**

```js
throw new Error('Resolved database URL is not a Postgres connection string');
```

When this fires in production (e.g., `DATABASE_URL` contains `'Manage remote databases'`), the error gives no hint of what the actual value was. Including the value — even truncated — helps debugging:
```js
throw new Error(
  `Resolved database URL is not a Postgres connection string: ${String(url).slice(0, 80)}`
);
```

**3. No workflow diff, but env vars are in scope** — the PR description says the deploy workflows already inject `STAGING_SUPABASE_PROJECT_REF` / `STAGING_SUPABASE_DB_PASSWORD` (and production equivalents) on the adopter migration step. The old README confirmed this, and the env var names match. Assuming that's still the case after PR #372 landing, no workflow changes are needed. Worth a quick check that the `Apply adopter database migrations` step in `deploy-staging.yml` has those vars in `env:` before merging.

scripts/lib/resolve-adopter-database-url.mjs
```diff
@@ -1,44 +1,113 @@
...
+  const pairs = [
+    ['SUPABASE_PROJECT_REF', 'SUPABASE_DB_PASSWORD'],
+    ['STAGING_SUPABASE_PROJECT_REF', 'STAGING_SUPABASE_DB_PASSWORD'],
+    ['PRODUCTION_SUPABASE_PROJECT_REF', 'PRODUCTION_SUPABASE_DB_PASSWORD'],
+  ];
+
+  for (const [refKey, passwordKey] of pairs) {
+    const projectRef = env[refKey]?.trim();
+    const dbPassword = env[passwordKey];
+    if (projectRef && dbPassword) {
```
sarah-artificerinnovations, line 47: 🟡 **`dbPassword` is not trimmed — inconsistent with `projectRef?.trim()` on the line above**

If a CI secret is stored with surrounding whitespace (copy-paste artifact, trailing newline), `projectRef` is normalized but `dbPassword` is not. The raw password lands verbatim in the Postgres URI via `encodeURIComponent(dbPassword)`. A trailing newline (`%0A`) at the end of the encoded password causes a connection refusal from Postgres, which surfaces as a cryptic auth failure rather than a "bad secret" error.

```js
const dbPassword = env[passwordKey]?.trim();
```

The test suite covers `p@ss&word` (special chars) but not whitespace padding — add a case:

```js
test('resolveLinkedCredentials trims dbPassword whitespace', () => {
  const result = resolveLinkedCredentials({
    SUPABASE_PROJECT_REF: 'abc123',
    SUPABASE_DB_PASSWORD: '  secret\n',
  });
  assert.equal(result.dbPassword, 'secret');
});
```

scripts/lib/resolve-adopter-database-url.mjs
```diff
@@ -1,44 +1,113 @@
...
+  return {};
+}
+
+/**
+ * @param {string} repoRoot
+ * @param {NodeJS.ProcessEnv} env
+ * @returns {string | undefined}
+ */
+export function readLinkedProjectRef(repoRoot, env = process.env) {
+  const refPath = path.join(repoRoot, 'supabase', '.temp', 'project-ref');
```
sarah-artificerinnovations, line 64: 🔵 **`readLinkedProjectRef` is a misleading name — the function returns a full Postgres connection URI, not a project ref string**

The name implies it returns a ref (like `abc123def`) but it actually returns `postgresql://postgres:…@db.abc123def.supabase.co:5432/postgres`. A caller reading only the call site in `resolveAdopterDatabaseUrl` would reasonably expect a short string and be surprised by the URI.

Consider `resolveLinkedUrlFromFile` or `readLinkedConnectionUri` to match the return type:

```js
export function readLinkedConnectionUri(repoRoot, env = process.env) {
```

scripts/lib/resolve-adopter-database-url.mjs
```diff
@@ -1,44 +1,113 @@
...
+/**
+ * @param {NodeJS.ProcessEnv} env
+ * @returns {{ projectRef?: string; dbPassword?: string }}
+ */
+export function resolveLinkedCredentials(env = process.env) {
+  const pairs = [
+    ['SUPABASE_PROJECT_REF', 'SUPABASE_DB_PASSWORD'],
+    ['STAGING_SUPABASE_PROJECT_REF', 'STAGING_SUPABASE_DB_PASSWORD'],
+    ['PRODUCTION_SUPABASE_PROJECT_REF', 'PRODUCTION_SUPABASE_DB_PASSWORD'],
+  ];
```
sarah-artificerinnovations, line 42: 🔵 **`PREVIEW_SUPABASE_PROJECT_REF` / `PREVIEW_SUPABASE_DB_PASSWORD` are not in the lookup pairs**

The list covers bare `SUPABASE_*`, `STAGING_SUPABASE_*`, and `PRODUCTION_SUPABASE_*` — but if the PR preview workflow (`pr-preview-environment.yml`) runs `db:apply-adopter -- --linked`, it would inject preview-scoped secrets and silently fall through all three pairs with no match. The error message ("Set DATABASE_URL or SUPABASE_PROJECT_REF…") wouldn't hint at the missing PREVIEW variant.

If preview deploys don't run adopter DB migrations today, this is fine — but worth a comment on the pairs array so future contributors know to add a `PREVIEW_*` entry if they wire it up.

scripts/__tests__/resolve-adopter-database-url.test.mjs
```diff
@@ -0,0 +1,178 @@
...
+import os from 'node:os';
+import path from 'node:path';
+import test from 'node:test';
+import {
+  assertPostgresUrl,
+  isPostgresUrl,
+  readLinkedProjectRef,
+  resolveAdopterDatabaseUrl,
+  resolveLinkedCredentials,
+} from '../lib/resolve-adopter-database-url.mjs';
```
sarah-artificerinnovations, line 14: 🔵 **`isScriptMain` is exported but not covered by the test suite**

The import list covers `assertPostgresUrl`, `isPostgresUrl`, `readLinkedProjectRef`, `resolveAdopterDatabaseUrl`, and `resolveLinkedCredentials` — but not `isScriptMain`. The guard behavior matters: it's the difference between the script running its `main()` block when executed directly vs silently no-oping when imported. A quick pair of cases:

```js
import { isScriptMain } from '../lib/resolve-adopter-database-url.mjs';

test('isScriptMain returns true when argv matches', () => {
  const url = import.meta.url;
  const { fileURLToPath } = await import('node:url');
  const { resolve } = await import('node:path');
  assert.equal(isScriptMain(url, resolve(fileURLToPath(url))), true);
});

test('isScriptMain returns false for a different path', () => {
  assert.equal(isScriptMain(import.meta.url, '/some/other/script.mjs'), false);
});
```

sarah-artificerinnovations (approved):
## Code review — PR #375 (Fix staging adopter DB deploy --linked)

The fix is correct and the right approach. Replacing `supabase db remote connection-string` (which requires a locally authenticated CLI session and doesn't compose with CI secrets) with direct URI construction from env vars is exactly the right trade-off: no external tool dependency, works wherever the ref + password pair is available, and the fallback chain (env pair → `.temp/project-ref` + `SUPABASE_DB_PASSWORD` → error) is clear and predictable.

Test coverage is solid: 178 lines covering special-character passwords, all three env var pair variants, the `DATABASE_URL` override, the file-based fallback, error paths, and the URL validation helpers.

One minor inconsistency to fix before merge; the rest are non-blocking observations.

---

### 🟡 One design gap

**`dbPassword` is not trimmed — inconsistent with `projectRef`** (`resolve-adopter-database-url.mjs` line 44)

See inline comment.

---

### 🔵 Three minor

1. **`readLinkedProjectRef` is a misleading name** — it returns a full Postgres URI, not a project ref string.
2. **`PREVIEW_SUPABASE_PROJECT_REF` / `PREVIEW_SUPABASE_DB_PASSWORD` are not in the lookup pairs** — if PR preview environments run adopter DB migrations, they'd be silently skipped.
3. **`isScriptMain` has no test coverage** in the new test file — it's exported and the guard pattern matters for scripts that can also be imported.

scripts/lib/resolve-adopter-database-url.mjs
ZappoMan: Good catch — fixed in 904826c. `dbPassword` is now trimmed alongside `projectRef`, with a test case for whitespace-padded secrets.

scripts/lib/resolve-adopter-database-url.mjs
ZappoMan: Added a JSDoc block on `readLinkedConnectionUri` clarifying this is local-dev only and uses `SUPABASE_DB_PASSWORD` (not staging/production variants). Also noted in README step 5.

scripts/lib/resolve-adopter-database-url.mjs
ZappoMan: Fixed in 904826c — `env[passwordKey]?.trim()`. Added `resolveLinkedCredentials trims dbPassword whitespace` test.

scripts/lib/resolve-adopter-database-url.mjs
```diff
@@ -1,44 +1,113 @@
...
+
+  for (const [refKey, passwordKey] of pairs) {
+    const projectRef = env[refKey]?.trim();
+    const dbPassword = env[passwordKey];
+    if (projectRef && dbPassword) {
+      return { projectRef, dbPassword };
+    }
```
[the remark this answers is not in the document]
ZappoMan: Fixed — `env[passwordKey]?.trim()` in 904826c.

scripts/lib/resolve-adopter-database-url.mjs
ZappoMan: Added a comment on the pairs array in 904826c noting PREVIEW_* is intentionally omitted today and where to add it if PR preview starts running adopter migrations.

scripts/lib/resolve-adopter-database-url.mjs
ZappoMan: Agreed — documented in JSDoc on `readLinkedConnectionUri` and clarified README step 5: "generic var only — not staging/production variants".

scripts/lib/resolve-adopter-database-url.mjs
```diff
@@ -1,44 +1,113 @@
...
+  if (!projectRef) {
+    return undefined;
+  }
+
+  const dbPassword = env.SUPABASE_DB_PASSWORD;
+  if (!dbPassword) {
+    return undefined;
+  }
+
+  return postgresConnectionUri(projectRef, dbPassword);
```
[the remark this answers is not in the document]
ZappoMan: Fixed — `SUPABASE_DB_PASSWORD` is trimmed before `postgresConnectionUri` in the file fallback path.

scripts/__tests__/resolve-adopter-database-url.test.mjs
ZappoMan: Added three `isScriptMain` test cases in 904826c (match, mismatch, missing argv).

scripts/lib/resolve-adopter-database-url.mjs
ZappoMan: Fixed in 904826c — when `supabase/.temp/project-ref` exists but `SUPABASE_DB_PASSWORD` is missing, `readLinkedConnectionUri` now throws a targeted error instead of silently returning `undefined`. Added test coverage.

docs/design/adopter-zones/db-migrations.md
```diff
@@ -51,7 +51,7 @@ supabase db push
 npm run db:apply-adopter -- --linked
 ```
 
-`--linked` resolves connection via `supabase db remote connection-string`, using `DATABASE_URL` / `SUPABASE_DB_PASSWORD` from CI secrets.
+`--linked` resolves the remote Postgres URL from `DATABASE_URL` or paired project-ref + DB-password env vars (`SUPABASE_*`, `STAGING_SUPABASE_*`, `PRODUCTION_SUPABASE_*`). Deploy workflows already inject the staging/production secrets on this step.
```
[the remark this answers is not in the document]
ZappoMan: Fixed — changed to "DB password" in 904826c.

scripts/lib/resolve-adopter-database-url.mjs
ZappoMan: Fixed in 904826c — password trimming + whitespace test case added.

scripts/lib/resolve-adopter-database-url.mjs
ZappoMan: Renamed to `readLinkedConnectionUri` in 904826c to match the return type.

ZappoMan:
Addressed all inline review feedback in 904826c:

- Trim `dbPassword` in both `resolveLinkedCredentials` and `readLinkedConnectionUri`
- Rename `readLinkedProjectRef` → `readLinkedConnectionUri`
- Targeted error when `supabase/.temp/project-ref` exists but `SUPABASE_DB_PASSWORD` is missing
- `assertPostgresUrl` now includes the first 80 chars of the bad value in the error message
- JSDoc + README note that file fallback is local-dev only (generic `SUPABASE_DB_PASSWORD`)
- Comment on pairs array re: `PREVIEW_*` intentionally omitted
- Doc typo fixed (`DB password`)
- Added tests: password whitespace trim, missing-password file fallback, `isScriptMain`

Re: Rahul's workflow note — confirmed `deploy-staging.yml` and `deploy-production.yml` already inject the staging/production project ref + DB password on the adopter migration step; no workflow change needed.

ZappoMan:
Follow-up fix pushed in 97a0a95 for the new CI failure.

**Root cause:** URL resolution was working — `psql` was reaching `db.{ref}.supabase.co`, but that host resolves to **IPv6** and GitHub Actions runners have no IPv6 route (`Network is unreachable`).

**Fix:** After the prior workflow step runs `supabase link`, it writes `supabase/.temp/pooler-url` (Supavisor session pooler, IPv4-compatible). `--linked` now prefers that pooler template + injects the DB password, falling back to direct `db.{ref}.supabase.co` only when the pooler file is absent.

This matches [Supabase's guidance](https://supabase.com/docs/guides/troubleshooting/supabase--your-network-ipv4-and-ipv6-compatibility-cHe3BP) for IPv4-only environments like GHA.

PR #376, opened 2026-05-24
Fix adopter DB deploy IPv6 failure in CI

ZappoMan:
## Summary
- Follow-up to #375: staging deploy still failed after URL resolution was fixed because `psql` connected to direct `db.{ref}.supabase.co`, which resolves to IPv6 — GitHub Actions has no IPv6 route.
- Prefer the Supavisor session pooler URL from `supabase/.temp/pooler-url` (written by the prior `supabase link` step in deploy workflows) when resolving `--linked` database URLs.
- Fall back to direct `db.{ref}.supabase.co` only when the pooler template is absent.

## Test plan
- [x] `npm run test:adopter`
- [ ] Re-run Deploy Staging and confirm `Apply adopter database migrations` succeeds

### DIFF
--- a/scripts/lib/resolve-adopter-database-url.mjs
+++ b/scripts/lib/resolve-adopter-database-url.mjs
@@ -2,7 +2,10 @@ import { existsSync, readFileSync } from 'node:fs';
 import path from 'node:path';
 import process from 'node:process';
 import { fileURLToPath } from 'node:url';
-import { postgresConnectionUri } from './setup-supabase.mjs';
+import {
+  applyPoolerPassword,
+  postgresConnectionUri,
+} from './setup-supabase.mjs';
 
 const LOCAL_DEFAULT_URL =
   'postgresql://postgres:postgres@127.0.0.1:54322/postgres';
@@ -53,6 +56,38 @@ export function resolveLinkedCredentials(env = process.env) {
 }
 
 /**
+ * @param {string} repoRoot
+ * @returns {string | undefined}
+ */
+export function readPoolerUrlTemplate(repoRoot) {
+  const poolerPath = path.join(repoRoot, 'supabase', '.temp', 'pooler-url');
+  if (!existsSync(poolerPath)) {
+    return undefined;
+  }
+
+  const template = readFileSync(poolerPath, 'utf8').trim();
+  return template || undefined;
+}
+
+/**
+ * Prefer Supavisor pooler URL (IPv4-compatible) when `supabase link` wrote
+ * supabase/.temp/pooler-url; fall back to direct db.{ref}.supabase.co.
+ *
+ * @param {string} repoRoot
+ * @param {string} dbPassword
+ * @param {string} projectRef
+ * @returns {string}
+ */
+export function buildLinkedConnectionUri(repoRoot, dbPassword, projectRef) {
+  const poolerTemplate = readPoolerUrlTemplate(repoRoot);
+  if (poolerTemplate) {
+    return applyPoolerPassword(poolerTemplate, dbPassword);
+  }
+
+  return postgresConnectionUri(projectRef, dbPassword);
+}
+
+/**
  * Local-dev fallback after `supabase link`: read project ref from disk and pair
  * with SUPABASE_DB_PASSWORD only (not STAGING_/PRODUCTION_ variants).
  *
@@ -78,7 +113,7 @@ export function readLinkedConnectionUri(repoRoot, env = process.env) {
     );
   }
 
-  return postgresConnectionUri(projectRef, dbPassword);
+  return buildLinkedConnectionUri(repoRoot, dbPassword, projectRef);
 }
 
 /**
@@ -100,7 +135,9 @@ export function resolveAdopterDatabaseUrl(options = {}) {
   if (linked) {
     const { projectRef, dbPassword } = resolveLinkedCredentials(env);
     if (projectRef && dbPassword) {
-      return assertPostgresUrl(postgresConnectionUri(projectRef, dbPassword));
+      return assertPostgresUrl(
+        buildLinkedConnectionUri(repoRoot, dbPassword, projectRef)
+      );
     }
 
     const linkedUrl = readLinkedConnectionUri(repoRoot, env);
--- a/scripts/lib/resolve-adopter-database-url.mjs
+++ b/scripts/lib/resolve-adopter-database-url.mjs
@@ -36,6 +36,8 @@ export function assertPostgresUrl(url) {
  * @returns {{ projectRef?: string; dbPassword?: string }}
  */
 export function resolveLinkedCredentials(env = process.env) {
+  // Generic SUPABASE_* wins for local/dev ergonomics. Preview CI clears those vars
+  // before calling --linked so SUPABASE_PREVIEW_* is used (see reset-preview-database.sh).
   const pairs = [
     ['SUPABASE_PROJECT_REF', 'SUPABASE_DB_PASSWORD'],
     ['SUPABASE_PREVIEW_PROJECT_REF', 'SUPABASE_PREVIEW_DB_PASSWORD'],
@@ -69,6 +71,36 @@ export function readPoolerUrlTemplate(repoRoot) {
 }
 
 /**
+ * @param {string} repoRoot
+ * @returns {string | undefined}
+ */
+export function readLinkedProjectRef(repoRoot) {
+  const refPath = path.join(repoRoot, 'supabase', '.temp', 'project-ref');
+  if (!existsSync(refPath)) {
+    return undefined;
+  }
+
+  const projectRef = readFileSync(refPath, 'utf8').trim();
+  return projectRef || undefined;
+}
+
+/**
+ * Supavisor session pooler usernames are `postgres.{projectRef}`.
+ *
+ * @param {string} poolerUrlTemplate
+ * @param {string} projectRef
+ * @returns {boolean}
+ */
+export function poolerTemplateMatchesProjectRef(poolerUrlTemplate, projectRef) {
+  try {
+    const url = new URL(poolerUrlTemplate.trim());
+    return url.username === `postgres.${projectRef}`;
+  } catch {
+    return false;
+  }
+}
+
+/**
  * Prefer Supavisor pooler URL (IPv4-compatible) when `supabase link` wrote
  * supabase/.temp/pooler-url; fall back to direct db.{ref}.supabase.co.
  *
@@ -79,7 +111,13 @@ export function readPoolerUrlTemplate(repoRoot) {
  */
 export function buildLinkedConnectionUri(repoRoot, dbPassword, projectRef) {
   const poolerTemplate = readPoolerUrlTemplate(repoRoot);
-  if (poolerTemplate) {
+  const linkedRef = readLinkedProjectRef(repoRoot);
+  const linkedRefMatches = !linkedRef || linkedRef === projectRef;
+  const poolerMatches =
+    poolerTemplate &&
+    poolerTemplateMatchesProjectRef(poolerTemplate, projectRef);
+
+  if (poolerTemplate && poolerMatches && linkedRefMatches) {
     return applyPoolerPassword(poolerTemplate, dbPassword);
   }
 
@@ -95,12 +133,7 @@ export function buildLinkedConnectionUri(repoRoot, dbPassword, projectRef) {
  * @returns {string | undefined}
  */
 export function readLinkedConnectionUri(repoRoot, env = process.env) {
-  const refPath = path.join(repoRoot, 'supabase', '.temp', 'project-ref');
-  if (!existsSync(refPath)) {
-    return undefined;
-  }
-
-  const projectRef = readFileSync(refPath, 'utf8').trim();
+  const projectRef = readLinkedProjectRef(repoRoot);
   if (!projectRef) {
     return undefined;
   }
--- a/scripts/lib/resolve-adopter-database-url.mjs
+++ b/scripts/lib/resolve-adopter-database-url.mjs
@@ -36,10 +36,9 @@ export function assertPostgresUrl(url) {
  * @returns {{ projectRef?: string; dbPassword?: string }}
  */
 export function resolveLinkedCredentials(env = process.env) {
-  // PREVIEW_SUPABASE_* is intentionally omitted — PR preview does not run adopter
-  // migrations today. Add a pair here if that workflow starts calling --linked.
   const pairs = [
     ['SUPABASE_PROJECT_REF', 'SUPABASE_DB_PASSWORD'],
+    ['SUPABASE_PREVIEW_PROJECT_REF', 'SUPABASE_PREVIEW_DB_PASSWORD'],
     ['STAGING_SUPABASE_PROJECT_REF', 'STAGING_SUPABASE_DB_PASSWORD'],
     ['PRODUCTION_SUPABASE_PROJECT_REF', 'PRODUCTION_SUPABASE_DB_PASSWORD'],
   ];
@@ -146,7 +145,7 @@ export function resolveAdopterDatabaseUrl(options = {}) {
     }
 
     throw new Error(
-      'Failed to resolve remote database URL for --linked. Set DATABASE_URL or project ref + DB password env vars (SUPABASE_*, STAGING_SUPABASE_*, or PRODUCTION_SUPABASE_*).'
+      'Failed to resolve remote database URL for --linked. Set DATABASE_URL or project ref + DB password env vars (SUPABASE_*, SUPABASE_PREVIEW_*, STAGING_SUPABASE_*, or PRODUCTION_SUPABASE_*).'
     );
   }
 
--- a/scripts/lib/setup-supabase.mjs
+++ b/scripts/lib/setup-supabase.mjs
@@ -41,11 +41,7 @@ export function runCmd(cmd, args, opts = {}) {
  */
 export function postgresConnectionUri(projectRef, dbPassword) {
   const enc = encodeURIComponent(dbPassword);
-  const url = new URL(
-    `postgresql://postgres:${enc}@db.${projectRef}.supabase.co:5432/postgres`
-  );
-  url.searchParams.set('sslmode', 'require');
-  return url.toString();
+  return `postgresql://postgres:${enc}@db.${projectRef}.supabase.co:5432/postgres?sslmode=require`;
 }
 
 /**
@@ -54,10 +50,18 @@ export function postgresConnectionUri(projectRef, dbPassword) {
  * @param {string} dbPassword
  */
 export function applyPoolerPassword(poolerUrlTemplate, dbPassword) {
-  const url = new URL(poolerUrlTemplate.trim());
+  let url;
+  try {
+    url = new URL(poolerUrlTemplate.trim());
+  } catch {
+    throw new Error(
+      `Invalid pooler URL template from supabase link: ${String(poolerUrlTemplate).slice(0, 80)}`
+    );
+  }
   const enc = encodeURIComponent(dbPassword);
   url.searchParams.set('sslmode', 'require');
-  return `postgresql://${url.username}:${enc}@${url.host}${url.pathname}${url.search}`;
+  // url.username is percent-decoded; Supabase pooler usernames (postgres.{ref}) are alphanumeric.
+  return `${url.protocol}//${url.username}:${enc}@${url.host}${url.pathname}${url.search}`;
 }
 
 /**
--- a/scripts/lib/setup-supabase.mjs
+++ b/scripts/lib/setup-supabase.mjs
@@ -22,14 +22,14 @@ export function runCmd(cmd, args, opts = {}) {
     });
     let stdout = '';
     let stderr = '';
-    child.stdout?.on('data', (d) => {
+    child.stdout?.on('data', d => {
       stdout += d.toString();
     });
-    child.stderr?.on('data', (d) => {
+    child.stderr?.on('data', d => {
       stderr += d.toString();
     });
     child.on('error', reject);
-    child.on('close', (code) => {
+    child.on('close', code => {
       resolve({ stdout, stderr, code: code ?? 1 });
     });
   });
@@ -41,7 +41,23 @@ export function runCmd(cmd, args, opts = {}) {
  */
 export function postgresConnectionUri(projectRef, dbPassword) {
   const enc = encodeURIComponent(dbPassword);
-  return `postgresql://postgres:${enc}@db.${projectRef}.supabase.co:5432/postgres`;
+  const url = new URL(
+    `postgresql://postgres:${enc}@db.${projectRef}.supabase.co:5432/postgres`
+  );
+  url.searchParams.set('sslmode', 'require');
+  return url.toString();
+}
+
+/**
+ * Inject password into a Supavisor pooler URL template (from supabase link).
+ * @param {string} poolerUrlTemplate
+ * @param {string} dbPassword
+ */
+export function applyPoolerPassword(poolerUrlTemplate, dbPassword) {
+  const url = new URL(poolerUrlTemplate.trim());
+  const enc = encodeURIComponent(dbPassword);
+  url.searchParams.set('sslmode', 'require');
+  return `postgresql://${url.username}:${enc}@${url.host}${url.pathname}${url.search}`;
 }
 
 /**
@@ -58,7 +74,10 @@ export function projectApiUrl(projectRef) {
 export function parseApiKeysJson(json) {
   const data = JSON.parse(json);
   if (data && typeof data === 'object' && !Array.isArray(data)) {
-    if (typeof data.anon === 'string' && typeof data.service_role === 'string') {
+    if (
+      typeof data.anon === 'string' &&
+      typeof data.service_role === 'string'
+    ) {
       return { anon: data.anon, service_role: data.service_role };
     }
   }
@@ -72,7 +91,8 @@ export function parseApiKeysJson(json) {
   }
   return {
     anon: out.anon || out['anon key'] || '',
-    service_role: out.service_role || out.serviceRole || out['service_role'] || '',
+    service_role:
+      out.service_role || out.serviceRole || out['service_role'] || '',
   };
 }
 
@@ -106,7 +126,9 @@ function normalizeSlugLike(s) {
  * @param {string} s
  */
 function compactAlnum(s) {
-  return String(s).toLowerCase().replace(/[^a-z0-9]/g, '');
+  return String(s)
+    .toLowerCase()
+    .replace(/[^a-z0-9]/g, '');
 }
 
 /**
@@ -129,7 +151,8 @@ export function scoreSupabaseProjectForTier(project, tier, slugBase) {
   const base = String(slugBase || '').toLowerCase();
   if (!base) return 0;
   const expected = `${base}-${tier}`;
-  const slugField = project.slug != null ? String(project.slug).trim().toLowerCase() : '';
+  const slugField =
+    project.slug != null ? String(project.slug).trim().toLowerCase() : '';
   if (slugField && slugField === expected) return 100;
 
   const displayName = String(project.name || '').trim();
@@ -175,10 +198,12 @@ export function pickRecommendedSupabaseProject(choices, tier, slugBase) {
 export function parseProjectsListJson(stdout) {
   const data = JSON.parse(stdout);
   if (!Array.isArray(data)) return [];
-  return data.map((p) => {
+  return data.map(p => {
     const rawSlug = p.slug;
     const slug =
-      typeof rawSlug === 'string' && rawSlug.trim() ? String(rawSlug).trim() : undefined;
+      typeof rawSlug === 'string' && rawSlug.trim()
+        ? String(rawSlug).trim()
+        : undefined;
     return {
       id: p.id || p.ref || p.project_id,
       name: p.name || p.slug || '',
@@ -195,7 +220,7 @@ export function parseProjectsListJson(stdout) {
 export function parseOrgsListJson(stdout) {
   const data = JSON.parse(stdout);
   if (!Array.isArray(data)) return [];
-  return data.map((o) => ({
+  return data.map(o => ({
     id: o.id,
     name: o.name || '',
   }));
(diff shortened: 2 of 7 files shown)


scripts/lib/setup-supabase.mjs
```diff
@@ -41,7 +41,23 @@ export function runCmd(cmd, args, opts = {}) {
...
+  url.searchParams.set('sslmode', 'require');
+  return url.toString();
+}
+
+/**
+ * Inject password into a Supavisor pooler URL template (from supabase link).
+ * @param {string} poolerUrlTemplate
+ * @param {string} dbPassword
+ */
+export function applyPoolerPassword(poolerUrlTemplate, dbPassword) {
```
rahul-artificerinnovations, line 52: **Non-blocking: `new URL()` throws `TypeError: Invalid URL` without context when the pooler-url file contains garbage.**

If `supabase/.temp/pooler-url` is written by a partially-failed `supabase link` (or manually corrupted), `new URL(poolerUrlTemplate.trim())` throws a bare `TypeError: Invalid URL`. The stack trace won't say which file the bad URL came from.

A small try/catch adds the missing context:
```js
let url;
try {
  url = new URL(poolerUrlTemplate.trim());
} catch {
  throw new Error(
    `Invalid pooler URL template in supabase/.temp/pooler-url: ${poolerUrlTemplate.slice(0, 80)}`
  );
}
```
Non-blocking — the error is still a fatal abort; this just makes it easier to diagnose.

rahul-artificerinnovations (approved):
Sound fix. The root cause (direct `db.{ref}.supabase.co` resolves to IPv6, which GitHub Actions can't reach) is correctly diagnosed and the solution — prefer the Supavisor pooler URL that `supabase link` already writes to `supabase/.temp/pooler-url` — is the right approach. The pooler hostname (`aws-1-us-east-2.pooler.supabase.com`) resolves to IPv4.

Things that look good:
- **`applyPoolerPassword` manual URL reconstruction**: using `url.username` + injected `encodeURIComponent(password)` in the string template is the right pattern. Setting `url.password = enc` would double-encode the percent chars; the manual reconstruction avoids that.
- **`url.searchParams.set('sslmode', 'require')` is idempotent**: if `supabase link` already wrote `?sslmode=require` in the template, the set is a no-op; `url.search` carries all existing params through.
- **`postgresConnectionUri` now adds `?sslmode=require`**: the direct fallback path also gets SSL, which is the right default for remote Supabase connections.
- **Test coverage**: the new `resolveAdopterDatabaseUrl linked prefers Supavisor pooler URL` test covers the happy path with pooler-url present; the updated staging/production/project-ref tests cover the fallback paths with the new `?sslmode=require` expectation.
- **Graceful degradation**: if `supabase/.temp/pooler-url` is absent, `buildLinkedConnectionUri` falls back to the direct URL automatically — no hard dependency on the file.

Two non-blocking notes below.

---

**1. `applyPoolerPassword` — `new URL()` throws `TypeError: Invalid URL` with no context** — see inline comment.

**2. Pooler URL template assumption not verified in this diff**

`buildLinkedConnectionUri` only works as intended when `supabase link` runs before `db:apply-adopter --linked` in the same working directory and writes `supabase/.temp/pooler-url`. There are no workflow changes in this diff, so the fix relies on that ordering already being in place. The PR description and README update both assert this ("after `supabase link`, which writes `supabase/.temp/pooler-url`"), and the staging test plan item will confirm it. If for any reason the file is absent, the code falls back silently to the direct IPv6 host and will fail again — worth verifying the workflow ordering is solid in the staging run.

mei-artificerinnovations (approved):
## PR #376 — Fix adopter DB deploy IPv6 failure in CI

Clean follow-up to #375. The root cause is correct (Supabase direct `db.{ref}.supabase.co` resolves to IPv6 on GitHub Actions; Supavisor pooler is IPv4-compatible), and the implementation is solid.

**What looks good:**

- `dbPassword` trimming — my PR #375 comment about the asymmetry with `projectRef` is already incorporated. ✅
- `readLinkedConnectionUri` now throws with a clear message when `project-ref` exists but `SUPABASE_DB_PASSWORD` is absent, instead of silently returning `undefined`. Previously that would fall through to the throw-from-`resolveAdopterDatabaseUrl`, but the new error message explains exactly which file was found and why the generic password var is expected. Better UX. ✅
- `sslmode=require` added to all connection URIs — should have been there from the start; correct to add here since psql will reject it otherwise on Supabase connections. ✅
- `assertPostgresUrl` now includes the URL snippet in the error message (80-char slice) — much easier to debug help-text bleed-through. ✅
- `buildLinkedConnectionUri` is cleanly factored out and directly unit-tested, so pooler vs. direct fallback behavior is verifiable without running actual deploy steps. ✅
- All existing tests updated for the new `?sslmode=require` suffix; all pass a fresh `repoRoot` temp dir so they don't bleed into one another. ✅

**One thing to confirm before relying on the pooler path in production:**

`buildLinkedConnectionUri` prefers `supabase/.temp/pooler-url` — a file written by `supabase link`. Supabase CLI started writing this file in relatively recent versions. If the deploy workflow pins an older CLI version that doesn't write `pooler-url`, the code silently falls back to `db.{ref}.supabase.co` (the IPv6 direct URL) and the deploy fails the same way.

Worth checking which `supabase` CLI version CI uses and confirming it writes `pooler-url` on `link`. The test plan's "Re-run Deploy Staging" item being unchecked is the one remaining gate — once that confirms success, this should be stable.

**Minor (won't affect correctness in practice):**

`applyPoolerPassword` sets `sslmode=require` via `url.searchParams.set()` before reconstructing with `url.search`. If a pooler URL template already contained other search params, they'd be preserved and `sslmode=require` would be appended. That's correct behavior, but if `sslmode` were already in the template it would appear twice in `url.search`. Not realistic for Supabase pooler URLs, but worth knowing.

LGTM — ready to merge once staging deploy confirms.

scripts/lib/setup-supabase.mjs
```diff
@@ -41,7 +41,23 @@ export function runCmd(cmd, args, opts = {}) {
  */
 export function postgresConnectionUri(projectRef, dbPassword) {
   const enc = encodeURIComponent(dbPassword);
-  return `postgresql://postgres:${enc}@db.${projectRef}.supabase.co:5432/postgres`;
+  const url = new URL(
+    `postgresql://postgres:${enc}@db.${projectRef}.supabase.co:5432/postgres`
```
sarah-artificerinnovations: 🔵 **`new URL()` round-trip to add `?sslmode=require` is unnecessary complexity**

The URL object is constructed, mutated, and then serialized just to append a static query param. A direct string append is simpler, avoids the parse/serialize round-trip entirely, and makes the password encoding explicit:

```js
export function postgresConnectionUri(projectRef, dbPassword) {
  const enc = encodeURIComponent(dbPassword);
  return `postgresql://postgres:${enc}@db.${projectRef}.supabase.co:5432/postgres?sslmode=require`;
}
```

(I verified that Node's `url.toString()` does correctly preserve `encodeURIComponent` output for the characters tested — this is a style/clarity note, not a correctness bug.)

scripts/lib/setup-supabase.mjs
```diff
@@ -41,7 +41,23 @@ export function runCmd(cmd, args, opts = {}) {
...
+ * Inject password into a Supavisor pooler URL template (from supabase link).
+ * @param {string} poolerUrlTemplate
+ * @param {string} dbPassword
+ */
+export function applyPoolerPassword(poolerUrlTemplate, dbPassword) {
+  const url = new URL(poolerUrlTemplate.trim());
+  const enc = encodeURIComponent(dbPassword);
+  url.searchParams.set('sslmode', 'require');
+  return `postgresql://${url.username}:${enc}@${url.host}${url.pathname}${url.search}`;
 }
```
sarah-artificerinnovations, line 65: 🔵 **`url.username` is the percent-decoded username — safe for Supabase project refs, but worth noting**

When `new URL(poolerUrlTemplate)` parses the template, `url.username` returns the decoded username (e.g. `postgres.abc123staging`). Since Supabase project refs are alphanumeric + hyphens, the decoded and encoded forms are identical and the reconstruction is correct.

If this function ever receives a template whose username contained a percent-encoded character (e.g. from a non-Supabase pooler), `url.username` would return the decoded character unencoded in the reconstructed URL. Using `encodeURIComponent(url.username)` would be strictly safer:

```js
return `postgresql://${encodeURIComponent(url.username)}:${enc}@${url.host}${url.pathname}${url.search}`;
```

For current usage this is theoretical — just worth knowing if the function's input scope ever widens.

sarah-artificerinnovations (approved):
## Code review — PR #376 (Fix adopter DB deploy IPv6 failure in CI)

The root cause diagnosis is correct and the fix is the right approach. GitHub Actions runners don't have IPv6 routes to Supabase's direct `db.{ref}.supabase.co` endpoints; the Supavisor pooler (`*.pooler.supabase.com`) resolves to IPv4 and is the supported connection path for CI. Reading the pooler URL from `supabase/.temp/pooler-url` (written by the existing `supabase link` step) is clean — no new CI plumbing required.

All items from the PR #375 review are resolved: `dbPassword?.trim()` is applied, the PREVIEW pair omission is documented with a comment explaining the intentional scope, and `readLinkedProjectRef` is renamed to `readLinkedConnectionUri`. Nicely executed.

Test coverage is comprehensive: `applyPoolerPassword`, `readPoolerUrlTemplate`, `buildLinkedConnectionUri`, and `isScriptMain` are all covered, and the expected URL strings are correctly updated to include `?sslmode=require` throughout.

Two minor notes — neither blocks merge.

---

### 🔵 Two minor

**1. `postgresConnectionUri` — unnecessary `new URL()` round-trip**

See inline comment on `setup-supabase.mjs` line 44.

**2. `applyPoolerPassword` — `url.username` used unencoded in the template literal**

See inline comment on `setup-supabase.mjs` line 57.

---

LGTM. The fix unblocks the staging adopter DB migration step and the approach is sound for production too.

scripts/lib/setup-supabase.mjs
```diff
@@ -41,7 +41,23 @@ export function runCmd(cmd, args, opts = {}) {
...
+/**
+ * Inject password into a Supavisor pooler URL template (from supabase link).
+ * @param {string} poolerUrlTemplate
+ * @param {string} dbPassword
+ */
+export function applyPoolerPassword(poolerUrlTemplate, dbPassword) {
+  const url = new URL(poolerUrlTemplate.trim());
+  const enc = encodeURIComponent(dbPassword);
+  url.searchParams.set('sslmode', 'require');
+  return `postgresql://${url.username}:${enc}@${url.host}${url.pathname}${url.search}`;
```
diego-artificerinnovations: `postgresql://` is hardcoded. If `supabase link` ever writes a `postgres://` pooler template, this output silently switches scheme. Use `url.protocol` to preserve whatever scheme the template had:

```js
return `${url.protocol}//${url.username}:${enc}@${url.host}${url.pathname}${url.search}`;
```

`url.protocol` returns `postgresql:` (with the colon), so `${url.protocol}//` reconstructs `postgresql://` correctly.

diego-artificerinnovations (approved):
## Approved

The pooler approach is correct. `supabase link` writes `supabase/.temp/pooler-url` with a Supavisor hostname that resolves to IPv4, so reading it and injecting the password via `applyPoolerPassword` solves the GitHub Actions IPv6 problem cleanly. `buildLinkedConnectionUri` is the right abstraction — it keeps the pooler preference in one place, and both `resolveAdopterDatabaseUrl` and `readLinkedConnectionUri` route through it consistently.

Also noticed that this PR incidentally fixes both non-blocking items from the #375 review: `dbPassword` is now trimmed in `resolveLinkedCredentials` (line 49), and the `SUPABASE_DB_PASSWORD`-only constraint in the file fallback is now documented with a descriptive throw message and a comment in `resolveLinkedCredentials` about the intentional omission of `PREVIEW_SUPABASE_*`.

---

### Non-blocking

**`applyPoolerPassword` hardcodes `postgresql://` scheme — see inline comment on `setup-supabase.mjs`**

If Supabase ever writes a `postgres://` pooler template (both schemes are valid), the output silently uses `postgresql://`. Low risk today, but easy to make defensive.

**`readLinkedConnectionUri` new throw path has no test — see inline comment on `resolve-adopter-database-url.mjs`**

The behavior changed from returning `undefined` (when project-ref file found but `SUPABASE_DB_PASSWORD` missing) to throwing with a descriptive message. The new throw path isn't covered — worth a test case to confirm the message is useful.

rahul-artificerinnovations (approved):
The two new commits land cleanly on top of the IPv6/pooler fix. Preview adopter migrations are wired correctly and the unlink noise is handled. No blocking issues.

---

## `resolveLinkedCredentials` preview tier (commit 2)

`SUPABASE_PREVIEW_PROJECT_REF` / `SUPABASE_PREVIEW_DB_PASSWORD` slotted between generic and staging — correct priority. The old "PREVIEW intentionally omitted" comment is removed, tests cover the new tier and confirm generic still wins when both are set.

## `reset-preview-database.sh` refactor (commits 2 + 3)

The dual-flag design (`SUPABASE_HAS_CHANGES` / `ADOPTER_HAS_CHANGES`) is clean. The four outcomes are all handled:

| supabase | adopter | result |
|---|---|---|
| unchanged | unchanged | skip all, return early |
| unchanged | changed | link only (no reset/seed), apply adopter migrations, no type regen |
| changed | unchanged | link + reset + seed + apply adopter migrations + regen types |
| changed | changed | same as above |

Type generation correctly gated on `SUPABASE_HAS_CHANGES` — adopter-only runs don't touch the public schema types.

**`sync_link_artifacts`**: the `pwd -P` self-copy guard is the right fix. When `SUPABASE_RUNTIME_DIR` is already `REPO_ROOT/supabase` (adopter-only path), `link_temp_dir == repo_temp_dir` after resolution and the copy is skipped. When a worktree was used, the link artifacts from the temp dir are copied into `REPO_ROOT/supabase/.temp/` where `db:apply-adopter` expects them. Both paths correct.

**`unlink_supabase` rewrite**: the `grep -qiE 'Cannot find project ref|failed to delete credentials'` pattern handles the "already unlinked" CI case cleanly. Output is always surfaced via `cat "${tmp}"`, temp file always cleaned up.

**`ensure_prereqs`** now validates `node` and `psql` are present before entering the script — good guard for the new dependency.

---

## Non-blocking note carried from initial review

`applyPoolerPassword` in `setup-supabase.mjs` still calls `new URL(poolerUrlTemplate.trim())` bare — if `supabase/.temp/pooler-url` contains garbage (partially-failed `supabase link`), the error is `TypeError: Invalid URL` with no context. The try/catch fix from my first review remains a worthwhile follow-up but is not a blocker.

---

No new issues in the preview wiring. Approve.

diego-artificerinnovations (approved):
## Approved

Reviewed commits `f636ff7` and `9b8829c` against the first APPROVE. Both look correct.

**Commit `f636ff7` — Apply adopter DB migrations in PR preview CI:**

- `SUPABASE_PREVIEW_PROJECT_REF` / `SUPABASE_PREVIEW_DB_PASSWORD` pair added to `resolveLinkedCredentials` at the right priority (generic → preview → staging → production). The "intentionally omitted" comment is correctly removed.
- `checkout_baseline` now tracks `ADOPTER_HAS_CHANGES` separately from `SUPABASE_HAS_CHANGES`. The four resulting cases in `main()` are all correct:
  - **Both unchanged** → skip all ✓
  - **Supabase only** → `link` + `reset` + `seed` + adopter migrations + types ✓ (adopter migrations always run after a full reset since it drops the `app` schema)
  - **Adopter only** → `link` (in repo's own `supabase/` dir since no temp worktree) + adopter migrations, no type gen ✓
  - **Both changed** → `link` + `reset` + `seed` + adopter migrations + types ✓
- `apply_adopter_migrations` injects `SUPABASE_PREVIEW_PROJECT_REF`/`SUPABASE_PREVIEW_DB_PASSWORD`, matching what `resolveLinkedCredentials` reads.

**Commit `9b8829c` — Self-copy guard and unlink error handling:**

- `sync_link_artifacts` self-copy guard is correct: in the adopter-only path, `SUPABASE_RUNTIME_DIR` is set to `REPO_ROOT/supabase` (no temp worktree), so `link_supabase` writes artifacts directly into `REPO_ROOT/supabase/.temp`. Without the guard, `cp -a` would copy a directory into itself. The `pwd -P` resolved-path comparison is the right way to detect this.
- `unlink_supabase` rewrite is clean: output captured to temp file, always printed for CI logs, known non-fatal patterns (`Cannot find project ref`, `failed to delete credentials`) demoted to warn, temp file always cleaned up. Dry-run exits before `mktemp` so no temp file is created in that path.

---

The two non-blocking items from the first review (`applyPoolerPassword` scheme hardcoding, missing `readLinkedConnectionUri` throw test) are unchanged — still non-blocking.

mei-artificerinnovations (approved):
## PR #376 re-review — preview environment adopter DB migration support

Clean addition on top of the already-approved staging IPv6 fix. Everything looks correct.

**What's new in this commit:**

- `resolveLinkedCredentials` now includes `SUPABASE_PREVIEW_PROJECT_REF` / `SUPABASE_PREVIEW_DB_PASSWORD` at priority position 2 — after the generic env pair, before staging and production. That ordering is right: preview environments are ephemeral and shouldn't shadow the primary generic credentials, but should resolve before the long-lived env-specific ones. ✅
- New tests cover preview credential resolution and verify that priority ordering holds (generic wins over preview wins over staging wins over production). ✅
- `sync_link_artifacts()` in `reset-preview-database.sh` copies `${SUPABASE_RUNTIME_DIR}/.temp/` into `${REPO_ROOT}/supabase/.temp/` before running `db:apply-adopter --linked`. This is exactly what `buildLinkedConnectionUri` needs — it reads `supabase/.temp/pooler-url` from the repo root. The same-directory skip (`[[ "${link_temp_dir}" == "${repo_temp_dir}" ]]`) is a nice safety rail. ✅
- `apply_adopter_migrations()` wraps `sync_link_artifacts` + the `npm run db:apply-adopter` call with the preview credentials exported. The subshell `( cd "${REPO_ROOT}" ... )` is the right pattern to avoid polluting the outer shell's working directory. ✅
- Refactored `main()` correctly separates `SUPABASE_HAS_CHANGES` from `ADOPTER_HAS_CHANGES` — adopter migrations now run on adopter-only changes without triggering a full template reset/seed cycle. That was the missing path before this PR. ✅
- `unlink_supabase()` dry-run + non-fatal "already unlinked" handling is a nice defensive improvement. ✅

**One thing to be aware of (not blocking):**

`sync_link_artifacts` logs a WARN and returns if the `.temp` dir doesn't exist, letting `apply_adopter_migrations` proceed with the direct IPv6 URL fallback. That fallback will fail on GitHub Actions for the same reason this whole PR exists. If `supabase link` didn't produce a `.temp/pooler-url` (e.g., old CLI version), the warning in logs will be the only signal before the deploy step errors. Consider whether that's acceptable or whether a hard failure at `sync_link_artifacts` would give faster feedback. Either choice is fine — just noting the tradeoff.

LGTM — the preview environment migration path is wired correctly end-to-end.

scripts/pr-preview/reset-preview-database.sh
```diff
@@ -470,39 +557,61 @@ main() {
...
-        log "WARN" "Type generation failed; using baseline types (already restored)."
-        db_success=false
+        supabase_ok=false
+        adopter_ok=false
+        log "WARN" "Failed to link preview Supabase project for adopter migrations."
+      fi
+    fi
+
+    if [[ "${supabase_ok}" == true && ( "${ADOPTER_HAS_CHANGES}" == true || "${SUPABASE_HAS_CHANGES}" == true ) ]]; then
+      if apply_adopter_migrations; then
```
sarah-artificerinnovations, line 597: 🔵 **Comment needed: adopter migrations run after any Supabase reset, regardless of `ADOPTER_HAS_CHANGES`**

The condition `ADOPTER_HAS_CHANGES == true || SUPABASE_HAS_CHANGES == true` means adopter migrations run whenever the Supabase DB was reset — even if no `adopter/db/` files changed in this PR. This is the correct behavior (a full DB reset wipes the adopter schema), but it's easy for a future reader to misread this as "run only when adopter files changed."

Consider a short comment above the condition:
```bash
# Adopter migrations run after any full Supabase reset (SUPABASE_HAS_CHANGES), not only when
# adopter/db/ changed, because a reset wipes the adopter schema too.
if [[ "${supabase_ok}" == true && ( "${ADOPTER_HAS_CHANGES}" == true || "${SUPABASE_HAS_CHANGES}" == true ) ]]; then
```

scripts/lib/resolve-adopter-database-url.mjs
```diff
@@ -33,10 +36,9 @@ export function assertPostgresUrl(url) {
  * @returns {{ projectRef?: string; dbPassword?: string }}
  */
 export function resolveLinkedCredentials(env = process.env) {
-  // PREVIEW_SUPABASE_* is intentionally omitted — PR preview does not run adopter
-  // migrations today. Add a pair here if that workflow starts calling --linked.
   const pairs = [
     ['SUPABASE_PROJECT_REF', 'SUPABASE_DB_PASSWORD'],
+    ['SUPABASE_PREVIEW_PROJECT_REF', 'SUPABASE_PREVIEW_DB_PASSWORD'],
     ['STAGING_SUPABASE_PROJECT_REF', 'STAGING_SUPABASE_DB_PASSWORD'],
```
sarah-artificerinnovations, line 44: 🔵 **Priority order: `SUPABASE_PROJECT_REF` (generic) wins over `SUPABASE_PREVIEW_PROJECT_REF`**

If the preview CI workflow sets the bare `SUPABASE_PROJECT_REF` env var (e.g., via an earlier `outputs:` step or a workflow-level `env:` block), it would silently shadow the explicit `SUPABASE_PREVIEW_PROJECT_REF="${PROJECT_REF}"` that `apply_adopter_migrations` injects. The adopter migrations would then connect to whatever project the generic var points to.

Worth verifying that the preview CI workflow doesn't export `SUPABASE_PROJECT_REF` globally before the `apply_adopter_migrations` subshell runs. If it does, consider moving `SUPABASE_PREVIEW_*` above the generic pair, or dropping the generic pair from this lookup entirely (the file-based `.temp/project-ref` fallback covers the `supabase link` case where the generic var is most likely set).

scripts/pr-preview/reset-preview-database.sh
```diff
@@ -120,6 +121,16 @@ ensure_prereqs() {
...
     exit 1
   }
+
+  command -v node >/dev/null 2>&1 || {
+    log "ERROR" "node is required."
+    exit 1
+  }
+
+  command -v psql >/dev/null 2>&1 || {
+    log "ERROR" "psql is required for adopter database migrations."
```
sarah-artificerinnovations: 🔵 **`psql` prerequisite — only correct if `db-apply-adopter.mjs` shells out to `psql`**

If `db-apply-adopter.mjs` uses a Node.js Postgres client library (e.g. `postgres`, `pg`) rather than spawning `psql`, this check is incorrect and would fail on CI runners where `psql` isn't installed but the Node client works fine. If it does shell out to `psql`, the check is right.

Verify against `scripts/db-apply-adopter.mjs` and remove this check if it's using a native client.

sarah-artificerinnovations (approved):
## Re-review — preview adopter DB migration support

The extension is correct and well-integrated. I traced through all four scenarios:

- **`SUPABASE_HAS_CHANGES=true`, `ADOPTER_HAS_CHANGES=true`** (full change): link → reset → seed → adopter migrations → types. ✓
- **`SUPABASE_HAS_CHANGES=false`, `ADOPTER_HAS_CHANGES=true`** (adopter-only): `SUPABASE_RUNTIME_DIR` set to repo root → link → adopter migrations. `sync_link_artifacts` correctly no-ops via the identity check since source and destination are the same directory. ✓
- **`SUPABASE_HAS_CHANGES=true`, `ADOPTER_HAS_CHANGES=false`** (Supabase-only): link → reset → seed → adopter migrations still run (correct: a full DB reset wipes adopter schema, so they must be re-applied regardless of whether adopter files changed in this PR). ✓
- **Both false** (no changes): early return, `db_success=true`. ✓

`sync_link_artifacts` is clean: the `pwd -P` identity check handles the same-directory case, and the `cp -a` idiom correctly copies dotfiles from the `.temp/` directory. `apply_adopter_migrations` correctly `cd`s to `REPO_ROOT` before running the npm script so `resolveAdopterDatabaseUrl`'s `process.cwd()` default finds the copied `pooler-url`.

The `unlink_supabase` rewrite is an improvement — capturing output to a temp file and matching known-non-fatal error patterns is more robust than the previous `--allow-fail` pattern.

Three minor notes:

1. **Undocumented intent in `main()`** — adopter migrations run whenever `SUPABASE_HAS_CHANGES=true`, even if `ADOPTER_HAS_CHANGES=false`. This is correct (a DB reset wipes adopter schema) but the condition reads as if it's symmetrical when it isn't. A comment would help. See inline.

2. **`SUPABASE_PROJECT_REF` priority over `SUPABASE_PREVIEW_PROJECT_REF`** — if the CI workflow sets `SUPABASE_PROJECT_REF` globally (e.g., from an `outputs:` step earlier in the job), it would shadow the explicit `SUPABASE_PREVIEW_PROJECT_REF` that `apply_adopter_migrations` injects. See inline.

3. **`psql` prerequisite** — see inline.

---

LGTM. Ready to merge.

scripts/lib/setup-supabase.mjs
ZappoMan: Fixed in be82d97: `applyPoolerPassword` now wraps `new URL()` in try/catch and throws `Invalid pooler URL template from supabase link: …` with a truncated copy of the bad value.

scripts/lib/resolve-adopter-database-url.mjs
```diff
@@ -52,6 +55,38 @@ export function resolveLinkedCredentials(env = process.env) {
...
+ * @returns {string}
+ */
+export function buildLinkedConnectionUri(repoRoot, dbPassword, projectRef) {
+  const poolerTemplate = readPoolerUrlTemplate(repoRoot);
+  if (poolerTemplate) {
+    return applyPoolerPassword(poolerTemplate, dbPassword);
+  }
+
+  return postgresConnectionUri(projectRef, dbPassword);
+}
```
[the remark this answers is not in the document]
ZappoMan: Fixed in be82d97: `buildLinkedConnectionUri` only uses `pooler-url` when the template username is `postgres.{projectRef}` and `supabase/.temp/project-ref` (if present) matches the requested ref. Stale templates fall back to direct `db.{ref}.supabase.co`. Tests added.

scripts/lib/setup-supabase.mjs
ZappoMan: Fixed in be82d97: simplified `postgresConnectionUri` to a direct string with explicit `encodeURIComponent` — no URL parse/serialize round-trip.

scripts/lib/setup-supabase.mjs
ZappoMan: Added a comment in `applyPoolerPassword` noting that `url.username` is percent-decoded and Supabase pooler usernames (`postgres.{ref}`) are alphanumeric-only, so reconstruction is safe for our templates.

scripts/lib/setup-supabase.mjs
ZappoMan: Fixed in be82d97: reconstruction now uses `${url.protocol}//${url.username}:…` so a `postgres://` template stays `postgres://`. Test added.

scripts/pr-preview/reset-preview-database.sh
```diff
@@ -120,6 +121,16 @@ ensure_prereqs() {
...
+    exit 1
+  }
```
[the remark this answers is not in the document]
ZappoMan: Fixed in be82d97: moved the `psql` check into `ensure_psql_prereq`, called only from `apply_adopter_migrations` right before adopter work runs.

scripts/pr-preview/reset-preview-database.sh
```diff
@@ -406,16 +438,71 @@ restore_types_from_baseline() {
...
+    return 0
+  fi
+
+  sync_link_artifacts
+
+  (
+    cd "${REPO_ROOT}"
+    SUPABASE_PREVIEW_PROJECT_REF="${PROJECT_REF}" \
+    SUPABASE_PREVIEW_DB_PASSWORD="${DB_PASSWORD}" \
+    npm run db:apply-adopter -- --linked
```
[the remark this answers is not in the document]
ZappoMan: Fixed in be82d97: `apply_adopter_migrations` now runs `env -u SUPABASE_PROJECT_REF -u SUPABASE_DB_PASSWORD` before exporting preview creds so generic vars cannot shadow the preview target.

scripts/lib/resolve-adopter-database-url.mjs
```diff
@@ -52,6 +54,38 @@ export function resolveLinkedCredentials(env = process.env) {
...
+ * @param {string} dbPassword
+ * @param {string} projectRef
+ * @returns {string}
+ */
+export function buildLinkedConnectionUri(repoRoot, dbPassword, projectRef) {
+  const poolerTemplate = readPoolerUrlTemplate(repoRoot);
+  if (poolerTemplate) {
+    return applyPoolerPassword(poolerTemplate, dbPassword);
+  }
+
```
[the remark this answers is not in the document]
ZappoMan: Same fix — `poolerTemplateMatchesProjectRef` + `project-ref` file check in `buildLinkedConnectionUri` (be82d97).

scripts/pr-preview/reset-preview-database.sh
ZappoMan: Added an inline comment above the condition in `main()` explaining that adopter migrations re-run after any Supabase reset because reset drops `app.*` (be82d97).

scripts/lib/resolve-adopter-database-url.mjs
ZappoMan: Verified: preview workflow only sets `SUPABASE_PREVIEW_*` at job level, not generic `SUPABASE_*`. Also documented the priority order in `resolveLinkedCredentials` and hardened preview apply with `env -u` (be82d97).

scripts/pr-preview/reset-preview-database.sh
ZappoMan: Confirmed `db-apply-adopter.mjs` shells out to `psql` via `spawnSync`. Moved the check to `ensure_psql_prereq()` so it runs only when adopter migrations will execute; error message now says why psql is required (be82d97).

PR #385, opened 2026-05-25
fix: replace GUC storage with config table in kit_sync_setup_cron (#383)

### LINKED ISSUE #383: Kit deploy fails on hosted Supabase: kit_sync_setup_cron cannot set app.kit_cron_secret GUC
## What happened?

After running `npm run setup:full` with Kit configured (or `setup:kit`), GitHub Actions deploy to production/staging fails at **Ensure kit-sync cron job** when `KIT_CRON_SECRET` is present:

```
kit_sync_setup_cron failed: 403 Forbidden {"code":"42501","details":null,"hint":null,"message":"permission denied to set parameter \"app.kit_cron_secret\""}
Error: Process completed with exit code 1.
```

**Expected:** Kit setup + first deploy registers the pg_cron job that periodically invokes the `kit-sync` Edge Function.

**Actual:** Deploy succeeds through Supabase migrations, edge function deploy, and Kit secret injection, then hard-fails on cron registration. Marketing email sync never runs on a schedule.

This is reproducible on **hosted Supabase** (not just a misconfigured adopter repo).

## How setup:full leads to this failure

1. **`setup:kit` / `setup:full` (kit phase)** collects or auto-generates `KIT_API_KEY`, `KIT_CRON_SECRET`, and `KIT_WEBHOOK_SECRET`, then syncs them to GitHub Actions secrets via `GITHUB_SECRETS` in `scripts/lib/setup-manifest.mjs`.
2. **Deploy workflows** (`deploy-production.yml`, `deploy-staging.yml`, `pr-preview-environment.yml`) deploy `kit-sync` / `kit-webhook` and set Supabase function secrets including `KIT_CRON_SECRET`.
3. **After function deploy**, `scripts/ensure-kit-sync-cron.mjs` calls the `kit_sync_setup_cron(p_url, p_secret)` RPC using the service role key.
4. The RPC (defined in `supabase/migrations/20260521000000_kit_sync_cron.sql`) tries:

   ```sql
   EXECUTE format(
     'ALTER DATABASE %I SET app.kit_cron_secret = %L',
     current_database(),
     p_secret
   );
   ```

5. Hosted Supabase returns `42501 permission denied to set parameter "app.kit_cron_secret"` — even though the function is `SECURITY DEFINER` and the caller uses `service_role`.

The migration comment explains the GUC was chosen to avoid inlining the secret in `cron.job.command` (which is readable). That design is reasonable, but **`ALTER DATABASE SET` for custom `app.*` GUCs is not permitted on hosted Supabase**.

## Asymmetry with other Kit deploy steps

- **Kit webhook ensure** (`ensure-kit-webhook-endpoint.mjs`) uses `continue-on-error: true` and skips when `KIT_API_KEY` is unset.
- **Kit cron ensure** has **no** `continue-on-error` and exits non-zero when `KIT_CRON_SECRET` is set (which `setup:full` always does when Kit is configured).

So adopters who complete Kit setup get a deploy blocker; adopters who skip Kit entirely avoid this step only if `KIT_CRON_SECRET` is absent.

## Steps to reproduce

1. Fork BeakerStack and run `npm run setup:full`, completing the Kit phase (or run `setup:kit` and sync secrets to GitHub).
2. Ensure `KIT_CRON_SECRET` is in GitHub Actions secrets (auto-generated by setup if omitted locally).
3. Push to `main` (production) or `develop` (staging).
4. Workflow passes `supabase db push` (migration `20260521000000_kit_sync_cron.sql` applies) and deploys edge functions.
5. **Ensure kit-sync cron job** runs `node scripts/ensure-kit-sync-cron.mjs` → RPC fails with the error above.

## Root cause

`kit_sync_setup_cron()` persists the cron bearer token via a database-level custom GUC (`app.kit_cron_secret`). Hosted Supabase restricts setting arbitrary custom parameters this way. The failure occurs before `cron.schedule()` runs, so no pg_cron job is registered.

Related tracking: #325 lists “Staging smoke test: `kit_sync_setup_cron()` succeeds (`ALTER DATABASE SET app.kit_cron_secret`)” as an open follow-up — this bug report confirms it **fails on hosted Supabase** with the current implementation.

## Suggested fix

Replace GUC storage with a private config table readable only from `SECURITY DEFINER` functions, e.g.:

- Create `kit_sync_runtime_config` (`singleton`, `worker_url`, `cron_secret`) with `REVOKE ALL FROM PUBLIC`.
- Update `kit_sync_setup_cron()` to `INSERT ... ON CONFLICT DO UPDATE` the secret/URL.
- Schedule pg_cron with a static SQL body that `SELECT`s URL and secret from that table (secret still not inlined in `cron.job.command`).

Example approach (adopter fork):

```sql
INSERT INTO kit_sync_runtime_config (singleton, worker_url, cron_secret)
VALUES (true, p_url, p_secret)
ON CONFLICT (singleton) DO UPDATE SET ...;

PERFORM cron.schedule(
  'kit-sync-worker',
  '*/5 * * * *',
  $sql$
    SELECT net.http_post(
      url := (SELECT worker_url FROM public.kit_sync_runtime_config WHERE singleton),
      headers := jsonb_build_object(
        'Authorization', 'Bearer ' || (SELECT cron_secret FROM public.kit_sync_runtime_config WHERE singleton),
        ...
      ),
      ...
    )
  $sql$
);
```

Ship as a follow-up migration replacing `kit_sync_setup_cron()`.

Optional hardening:

- Add `continue-on-error: true` on cron ensure (matching webhook) **only** as a temporary safety net — adopters would still have broken Kit sync without noticing.
- Document in setup:full / Kit recipe that cron registration requires the migration fix on hosted Supabase.

## Workarounds (adopters)

- **Skip Kit cron:** delete `KIT_CRON_SECRET` from GitHub secrets — deploy skips the step (`KIT_CRON_SECRET not set — skipping cron registration`). Kit sync will not run on a schedule.
- **Skip Kit entirely:** run `setup:full` with `--skip-kit` (or omit Kit secrets) if marketing email sync is not needed yet.

## Environment

- Template: BeakerStack `2026.001`
- Workflow: `deploy-production.yml` step **Ensure kit-sync cron job (production)**
- Hosted Supabase project (production tier)
- Supabase CLI v2.54.11 in CI

## Related

- #325 — Kit Phase 3 follow-ups (cron verification checklist)
- #315 — Kit Phase 3 deploy secret handling (closed)
- #382 — separate setup:full gap: Supabase service role keys not synced to GitHub secrets

mei-artificerinnovations:
## Summary

Fixes #383.

`kit_sync_setup_cron()` stored the cron bearer token via `ALTER DATABASE SET app.kit_cron_secret`, which hosted Supabase blocks with a 403/42501 even from a `SECURITY DEFINER` function. Adopters who complete Kit setup always have `KIT_CRON_SECRET` in GitHub secrets, so their first deploy hard-fails at the **Ensure kit-sync cron job** step.

**Changes:**

- New migration `20260524000000_kit_sync_config_table.sql`:
  - Creates `kit_sync_runtime_config` (singleton row: `worker_url`, `cron_secret`, `updated_at`) with `REVOKE ALL FROM PUBLIC` — table is private to the SECURITY DEFINER function and service_role.
  - Replaces the `ALTER DATABASE SET` call in `kit_sync_setup_cron()` with `INSERT ... ON CONFLICT DO UPDATE`.
  - Updates the pg_cron job body to `SELECT` URL and secret from the table at runtime — secret is still never inlined in `cron.job.command`.

No changes to `ensure-kit-sync-cron.mjs` or any workflow files — the RPC signature (`kit_sync_setup_cron(p_url, p_secret)`) is unchanged.

## Test plan

- [ ] Run `supabase db push` locally (or against a branch database) — migration applies cleanly, function compiles, no errors.
- [ ] Call `kit_sync_setup_cron('https://example.supabase.co/functions/v1/kit-sync', 'test-secret')` via service-role RPC — row appears in `kit_sync_runtime_config`, `kit-sync-worker` job appears in `cron.job`.
- [ ] Confirm `cron.job.command` for `kit-sync-worker` does not contain the literal secret value.
- [ ] Call again with updated URL — row is updated, old cron job unscheduled, new one registered.
- [ ] Re-run a staging deploy with `KIT_CRON_SECRET` set — **Ensure kit-sync cron job** step passes (previously returned 403).

### DIFF
--- a/supabase/migrations/20260524000000_kit_sync_config_table.sql
+++ b/supabase/migrations/20260524000000_kit_sync_config_table.sql
@@ -1,0 +1,83 @@
+-- Fix: replace GUC-based secret storage with a private config table.
+-- ALTER DATABASE SET app.kit_cron_secret is blocked on hosted Supabase (error 42501)
+-- even from a SECURITY DEFINER function. This migration introduces a singleton
+-- config table and rewrites kit_sync_setup_cron() to upsert there instead.
+-- The cron job body reads URL and secret from the table at runtime, so the
+-- secret is never inlined in cron.job.command (same privacy guarantee as before).
+
+CREATE TABLE IF NOT EXISTS public.kit_sync_runtime_config (
+  singleton   boolean     PRIMARY KEY DEFAULT true,
+  worker_url  text        NOT NULL,
+  cron_secret text        NOT NULL,
+  updated_at  timestamptz NOT NULL DEFAULT now(),
+  CONSTRAINT singleton_only CHECK (singleton IS TRUE)
+);
+
+ALTER TABLE public.kit_sync_runtime_config ENABLE ROW LEVEL SECURITY;
+
+-- Supabase's default privileges grant SELECT/INSERT/UPDATE/DELETE to anon and
+-- authenticated on all new public tables. Revoke explicitly so the table is
+-- inaccessible via PostgREST regardless of RLS policy state.
+-- service_role is intentionally excluded: the SECURITY DEFINER function runs as
+-- the owner and needs no table grant; omitting the service_role grant prevents
+-- direct REST reads of cron_secret via the service key.
+REVOKE ALL ON TABLE public.kit_sync_runtime_config FROM PUBLIC;
+REVOKE ALL ON TABLE public.kit_sync_runtime_config FROM anon, authenticated;
+
+-- ── Updated cron registration helper ─────────────────────────────────────────
+-- Replaces ALTER DATABASE SET with an upsert into kit_sync_runtime_config.
+-- Cron job body selects worker_url and cron_secret from the table at runtime.
+
+CREATE OR REPLACE FUNCTION kit_sync_setup_cron(
+  p_url    text,
+  p_secret text
+)
+RETURNS void
+LANGUAGE plpgsql
+SECURITY DEFINER
+SET search_path = extensions, net, public
+AS $$
+BEGIN
+  -- Existing databases may still have app.kit_cron_secret set by the prior GUC-based
+  -- migration. Best-effort reset so the stale value doesn't linger in pg_db_role_setting;
+  -- silently ignored if ALTER DATABASE RESET is blocked on hosted Supabase.
+  BEGIN
+    EXECUTE format('ALTER DATABASE %I RESET app.kit_cron_secret', current_database());
+  EXCEPTION WHEN OTHERS THEN
+    NULL;
+  END;
+
+  INSERT INTO public.kit_sync_runtime_config (singleton, worker_url, cron_secret, updated_at)
+  VALUES (true, p_url, p_secret, now())
+  ON CONFLICT (singleton) DO UPDATE SET
+    worker_url  = EXCLUDED.worker_url,
+    cron_secret = EXCLUDED.cron_secret,
+    updated_at  = EXCLUDED.updated_at;
+
+  PERFORM cron.unschedule('kit-sync-worker')
+  WHERE EXISTS (
+    SELECT 1 FROM cron.job WHERE jobname = 'kit-sync-worker'
+  );
+
+  PERFORM cron.schedule(
+    'kit-sync-worker',
+    '*/5 * * * *',
+    $sql$
+      SELECT net.http_post(
+        url     := c.worker_url,
+        headers := jsonb_build_object(
+          'Content-Type',  'application/json',
+          'Authorization', 'Bearer ' || c.cron_secret
+        ),
+        body    := '{}'::jsonb
+      )
+      FROM public.kit_sync_runtime_config c
+      WHERE c.singleton IS TRUE
+      LIMIT 1
+    $sql$
+  );
+END;
+$$;
+
+REVOKE ALL ON FUNCTION kit_sync_setup_cron(text, text) FROM PUBLIC;
+GRANT  EXECUTE ON FUNCTION kit_sync_setup_cron(text, text) TO service_role;
--- a/supabase/tests/kit_sync_cron.test.sql
+++ b/supabase/tests/kit_sync_cron.test.sql
@@ -1,6 +1,6 @@
 BEGIN;
 
-SELECT plan(8);
+SELECT plan(17);
 
 -- 1. pg_cron extension exists
 SELECT has_extension('pg_cron', 'pg_cron extension is installed');
@@ -72,6 +72,76 @@ SELECT ok(
    WHERE email = 'cron-test@example.com' AND status = 'processing')::int = 1,
   'kit_sync_dequeue returns pending row as processing'
 );
+
+-- 9. kit_sync_runtime_config table exists
+SELECT has_table('public', 'kit_sync_runtime_config', 'kit_sync_runtime_config table exists');
+
+-- 10. RLS is enabled on kit_sync_runtime_config
+SELECT ok(
+  (SELECT c.relrowsecurity
+   FROM pg_class c
+   JOIN pg_namespace n ON n.oid = c.relnamespace
+   WHERE n.nspname = 'public' AND c.relname = 'kit_sync_runtime_config'),
+  'kit_sync_runtime_config has RLS enabled'
+);
+
+-- 11. anon role has no SELECT on kit_sync_runtime_config
+SELECT ok(
+  NOT has_table_privilege('anon', 'public.kit_sync_runtime_config', 'SELECT'),
+  'anon has no SELECT on kit_sync_runtime_config'
+);
+
+-- 12. authenticated role has no SELECT on kit_sync_runtime_config
+-- Supabase default privileges grant both anon and authenticated; test both.
+SELECT ok(
+  NOT has_table_privilege('authenticated', 'public.kit_sync_runtime_config', 'SELECT'),
+  'authenticated has no SELECT on kit_sync_runtime_config'
+);
+
+-- 13. kit_sync_setup_cron executes without error and upserts config row
+SELECT lives_ok(
+  $$ SELECT kit_sync_setup_cron('https://test.supabase.co/functions/v1/kit-sync', 'test-secret-abc') $$,
+  'kit_sync_setup_cron executes without error'
+);
+
+-- 14. config row contains the expected worker_url and cron_secret
+SELECT ok(
+  EXISTS (
+    SELECT 1 FROM public.kit_sync_runtime_config
+    WHERE singleton = true
+      AND worker_url  = 'https://test.supabase.co/functions/v1/kit-sync'
+      AND cron_secret = 'test-secret-abc'
+  ),
+  'kit_sync_runtime_config row has correct worker_url and cron_secret after setup'
+);
+
+-- 15. cron.job row exists for kit-sync-worker
+SELECT ok(
+  EXISTS (SELECT 1 FROM cron.job WHERE jobname = 'kit-sync-worker'),
+  'kit-sync-worker cron job row exists after setup'
+);
+
+-- 16. cron.job.command for kit-sync-worker does not contain the literal secret
+SELECT ok(
+  NOT EXISTS (
+    SELECT 1 FROM cron.job
+    WHERE jobname = 'kit-sync-worker'
+      AND command LIKE '%test-secret-abc%'
+  ),
+  'kit-sync-worker cron command does not inline the bearer secret'
+);
+
+-- 17. service_role has no SELECT on kit_sync_runtime_config via table grant
+-- The SECURITY DEFINER function accesses the table as owner; no grant needed.
+SELECT ok(
+  NOT has_table_privilege('service_role', 'public.kit_sync_runtime_config', 'SELECT'),
+  'service_role has no direct SELECT on kit_sync_runtime_config'
+);
+
+-- Cleanup: unschedule the test cron job to avoid leaving stale rows in
+-- cron.job on non-ephemeral databases (pg_cron commits independently of
+-- the outer pgTAP transaction).
+PERFORM cron.unschedule('kit-sync-worker');
 
 SELECT * FROM finish();
 


supabase/migrations/20260524000000_kit_sync_config_table.sql
```diff
@@ -0,0 +1,63 @@
...
+-- secret is never inlined in cron.job.command (same privacy guarantee as before).
+
+CREATE TABLE IF NOT EXISTS kit_sync_runtime_config (
+  singleton   boolean     PRIMARY KEY DEFAULT true,
+  worker_url  text        NOT NULL,
+  cron_secret text        NOT NULL,
+  updated_at  timestamptz NOT NULL DEFAULT now(),
+  CONSTRAINT singleton_only CHECK (singleton)
+);
+
```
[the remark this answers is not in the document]
mei-artificerinnovations: Good catch — fixed in f04ae3dfeafe150c02c886e19dab13f15e01a63e. Added `ALTER TABLE public.kit_sync_runtime_config ENABLE ROW LEVEL SECURITY;` immediately after the table definition, matching the pattern from `marketing_email_v1.sql`.

supabase/migrations/20260524000000_kit_sync_config_table.sql
```diff
@@ -0,0 +1,63 @@
...
+REVOKE ALL ON TABLE kit_sync_runtime_config FROM PUBLIC;
+GRANT SELECT, INSERT, UPDATE ON TABLE kit_sync_runtime_config TO service_role;
```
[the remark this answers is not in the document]
mei-artificerinnovations: Fixed in f04ae3dfeafe150c02c886e19dab13f15e01a63e. `CREATE TABLE IF NOT EXISTS public.kit_sync_runtime_config`, and the `REVOKE`/`GRANT` statements are now `ON TABLE public.kit_sync_runtime_config` as well. The `INSERT` inside `kit_sync_setup_cron()` is also updated to `public.kit_sync_runtime_config`.

supabase/migrations/20260524000000_kit_sync_config_table.sql
```diff
@@ -0,0 +1,63 @@
...
+SECURITY DEFINER
+SET search_path = extensions, net, public
+AS $$
+BEGIN
+  INSERT INTO kit_sync_runtime_config (singleton, worker_url, cron_secret, updated_at)
+  VALUES (true, p_url, p_secret, now())
+  ON CONFLICT (singleton) DO UPDATE SET
+    worker_url  = EXCLUDED.worker_url,
+    cron_secret = EXCLUDED.cron_secret,
+    updated_at  = EXCLUDED.updated_at;
```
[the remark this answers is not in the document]
mei-artificerinnovations: Added in f04ae3dfeafe150c02c886e19dab13f15e01a63e — extended `kit_sync_cron.test.sql` from 8 to 14 assertions:

- Table exists
- RLS enabled
- `anon` has no SELECT
- `kit_sync_setup_cron(url, secret)` executes without error (`lives_ok`)
- Config row has correct `worker_url` and `cron_secret` after the call
- `cron.job.command` for `kit-sync-worker` does not contain the literal secret value

mei-artificerinnovations:
**CI failure root cause + fix (commit 412921e)**

Test 11 failed: `anon has no SELECT on kit_sync_runtime_config`.

Supabase's local dev setup applies `ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT, INSERT, UPDATE, DELETE ON TABLES TO anon, authenticated` during initialization. That's a direct grant to those roles — not a PUBLIC grant — so `REVOKE ALL FROM PUBLIC` doesn't touch it. The new table was getting SELECT access handed to `anon` automatically at create time.

Fix: added `REVOKE ALL ON TABLE public.kit_sync_runtime_config FROM anon, authenticated;` explicitly. The comment in the migration explains why (same reason RLS is also enabled — defense-in-depth against both grant paths). The test should now pass.

supabase/migrations/20260524000000_kit_sync_config_table.sql
```diff
@@ -0,0 +1,69 @@
...
+);
+
+ALTER TABLE public.kit_sync_runtime_config ENABLE ROW LEVEL SECURITY;
+
+-- Supabase's default privileges grant SELECT/INSERT/UPDATE/DELETE to anon and
+-- authenticated on all new public tables. Revoke explicitly so the table is
+-- inaccessible via PostgREST regardless of RLS policy state.
+REVOKE ALL ON TABLE public.kit_sync_runtime_config FROM PUBLIC;
+REVOKE ALL ON TABLE public.kit_sync_runtime_config FROM anon, authenticated;
+GRANT SELECT, INSERT, UPDATE ON TABLE public.kit_sync_runtime_config TO service_role;
```
ZappoMan: **Security:** `GRANT SELECT, INSERT, UPDATE … TO service_role` lets anyone with the service-role key read `cron_secret` directly via PostgREST (`GET /rest/v1/kit_sync_runtime_config`), bypassing the RPC entirely. Supabase's service role also bypasses RLS, so the REVOKE/RLS hardening doesn't help here.

Deploy only calls `kit_sync_setup_cron()` (SECURITY DEFINER), which runs as the function owner and does not require the caller to hold table privileges. Consider dropping table grants to `service_role` entirely — keep only `GRANT EXECUTE` on the function — so the secret is reachable only by the definer owner and the pg_cron runner (postgres), not via REST.

supabase/migrations/20260524000000_kit_sync_config_table.sql
```diff
@@ -0,0 +1,69 @@
...
+-- config table and rewrites kit_sync_setup_cron() to upsert there instead.
+-- The cron job body reads URL and secret from the table at runtime, so the
+-- secret is never inlined in cron.job.command (same privacy guarantee as before).
+
+CREATE TABLE IF NOT EXISTS public.kit_sync_runtime_config (
+  singleton   boolean     PRIMARY KEY DEFAULT true,
+  worker_url  text        NOT NULL,
+  cron_secret text        NOT NULL,
+  updated_at  timestamptz NOT NULL DEFAULT now(),
+  CONSTRAINT singleton_only CHECK (singleton)
```
ZappoMan: Nit: `CHECK (singleton)` works but reads oddly on a boolean PK column. `CHECK (singleton IS TRUE)` (or `= true`) makes the intent explicit and matches how the tests filter (`WHERE singleton = true`).

supabase/migrations/20260524000000_kit_sync_config_table.sql
```diff
@@ -0,0 +1,69 @@
...
+  WHERE EXISTS (
+    SELECT 1 FROM cron.job WHERE jobname = 'kit-sync-worker'
+  );
+
+  PERFORM cron.schedule(
+    'kit-sync-worker',
+    '*/5 * * * *',
+    $sql$
+      SELECT net.http_post(
+        url     := (SELECT worker_url  FROM public.kit_sync_runtime_config WHERE singleton),
```
ZappoMan: The cron body runs two independent subqueries against `kit_sync_runtime_config`. Consider a single fetch, e.g. `(SELECT c FROM public.kit_sync_runtime_config c WHERE c.singleton LIMIT 1)` and reference `c.worker_url` / `c.cron_secret`, or a small helper function. Not a blocker — this runs every 5 minutes — but it avoids two heap hits and makes the "singleton row" assumption obvious.

supabase/migrations/20260524000000_kit_sync_config_table.sql
```diff
@@ -0,0 +1,69 @@
...
+  p_url    text,
+  p_secret text
+)
+RETURNS void
+LANGUAGE plpgsql
+SECURITY DEFINER
+SET search_path = extensions, net, public
+AS $$
+BEGIN
+  INSERT INTO public.kit_sync_runtime_config (singleton, worker_url, cron_secret, updated_at)
```
ZappoMan, line 50: Worth a one-line comment that existing databases may still have `app.kit_cron_secret` set from the prior migration. Optional follow-up: `EXECUTE format('ALTER DATABASE %I RESET app.kit_cron_secret', current_database())` here for hygiene so stale GUC values don't linger in `pg_db_role_setting` after deploy.

supabase/tests/kit_sync_cron.test.sql
```diff
@@ -73,6 +73,51 @@ SELECT ok(
...
+-- 10. RLS is enabled on kit_sync_runtime_config
+SELECT ok(
+  (SELECT c.relrowsecurity
+   FROM pg_class c
+   JOIN pg_namespace n ON n.oid = c.relnamespace
+   WHERE n.nspname = 'public' AND c.relname = 'kit_sync_runtime_config'),
+  'kit_sync_runtime_config has RLS enabled'
+);
+
+-- 11. anon role has no SELECT on kit_sync_runtime_config
```
ZappoMan, line 88: Good catch revoking `anon` — the migration also revokes `authenticated`. Consider mirroring that here:

```sql
NOT has_table_privilege('authenticated', 'public.kit_sync_runtime_config', 'SELECT')
```

Supabase default privileges grant both roles on new public tables; testing only `anon` leaves a gap if someone regresses the second `REVOKE`.

supabase/tests/kit_sync_cron.test.sql
```diff
@@ -73,6 +73,51 @@ SELECT ok(
...
+SELECT ok(
+  NOT has_table_privilege('anon', 'public.kit_sync_runtime_config', 'SELECT'),
+  'anon has no SELECT on kit_sync_runtime_config'
+);
+
+-- 12. kit_sync_setup_cron executes without error and upserts config row
+SELECT lives_ok(
```
ZappoMan, line 102: Calling `kit_sync_setup_cron` inside a `BEGIN … ROLLBACK` pgTAP transaction may leave a real `cron.job` row behind — pg_cron registration often commits independently of the outer test transaction. If CI ever runs pgTAP against a long-lived local DB (not just ephemeral CI), this can accumulate stale jobs.

Consider an explicit `PERFORM cron.unschedule('kit-sync-worker')` after the assertions (still inside the test block), or document that this test requires a fresh DB.

supabase/tests/kit_sync_cron.test.sql
```diff
@@ -73,6 +73,51 @@ SELECT ok(
...
+  EXISTS (
+    SELECT 1 FROM public.kit_sync_runtime_config
+    WHERE singleton = true
+      AND worker_url  = 'https://test.supabase.co/functions/v1/kit-sync'
+      AND cron_secret = 'test-secret-abc'
+  ),
+  'kit_sync_runtime_config row has correct worker_url and cron_secret after setup'
+);
+
+-- 14. cron.job.command for kit-sync-worker does not contain the literal secret
```
ZappoMan: This assertion only checks that the secret is **not** inlined in `cron.job.command`. It would still pass if `kit_sync_setup_cron` succeeded at the upsert but `cron.schedule` silently failed (no row in `cron.job`). Consider also asserting a job exists:

```sql
EXISTS (SELECT 1 FROM cron.job WHERE jobname = 'kit-sync-worker')
```

…either as a separate test or combined with the non-inlining check.

supabase/tests/kit_sync_cron.test.sql
```diff
@@ -73,6 +73,51 @@ SELECT ok(
...
+  'anon has no SELECT on kit_sync_runtime_config'
+);
+
+-- 12. kit_sync_setup_cron executes without error and upserts config row
+SELECT lives_ok(
+  $$ SELECT kit_sync_setup_cron('https://test.supabase.co/functions/v1/kit-sync', 'test-secret-abc') $$,
+  'kit_sync_setup_cron executes without error'
+);
+
+-- 13. config row contains the expected worker_url and cron_secret
```
ZappoMan: Test 13 reads `cron_secret` directly as the superuser test runner, which is fine for verifying the upsert. If you drop the `service_role` table grants (see migration comment on line 23), that's the right posture — the secret should only be readable by the function owner and pg_cron, not by API roles.

ZappoMan (commented):
## Summary

Solid fix for #383 — moving cron secret storage off `ALTER DATABASE SET` to a singleton table is the right approach for hosted Supabase, and the pgTAP coverage for RLS, anon revocation, upsert behavior, and secret non-inlining is a meaningful improvement over the original migration.

**Blocking before merge:** please rebase onto `develop` and retarget the PR base branch to `develop` (currently targets `main`; `develop` is 3 commits ahead with no migration conflicts, but our merge workflow expects fixes to land on `develop` first).

**Overall:** CI is green and the RPC contract is unchanged, so `ensure-kit-sync-cron.mjs` and deploy workflows need no updates. A few security and test-hardening nits below — the service_role table grant is the most important one.

supabase/migrations/20260524000000_kit_sync_config_table.sql
mei-artificerinnovations: Fixed — the GRANT to service_role is removed. Added a comment above the REVOKE block explaining that the SECURITY DEFINER function accesses the table as owner and needs no grant, so cron_secret is unreachable via PostgREST regardless of the service key. Test 17 now asserts service_role has no direct SELECT.

supabase/migrations/20260524000000_kit_sync_config_table.sql
mei-artificerinnovations: Fixed — changed to CHECK (singleton IS TRUE). Also updated the cron body WHERE clause to match (WHERE c.singleton IS TRUE) for consistency.

supabase/migrations/20260524000000_kit_sync_config_table.sql
mei-artificerinnovations: Fixed — the cron body now does a single row fetch: FROM public.kit_sync_runtime_config c WHERE c.singleton IS TRUE LIMIT 1, with c.worker_url and c.cron_secret referenced directly. One heap hit instead of two.

supabase/migrations/20260524000000_kit_sync_config_table.sql
mei-artificerinnovations: Fixed — added a comment inside the function body about the stale GUC, plus a best-effort nested BEGIN...EXCEPTION WHEN OTHERS block that executes ALTER DATABASE ... RESET app.kit_cron_secret. The exception handler makes it a no-op on hosted Supabase where RESET may also be restricted, so it cannot break the deploy.

supabase/tests/kit_sync_cron.test.sql
mei-artificerinnovations: Fixed — added test 12 asserting authenticated has no SELECT (mirrors the anon check). Plan is now 17.

supabase/tests/kit_sync_cron.test.sql
mei-artificerinnovations: Fixed — added PERFORM cron.unschedule("kit-sync-worker") after the assertions (still inside the pgTAP block, before SELECT * FROM finish()). This cleans up the cron.job row even if pg_cron committed independently of the test transaction.

supabase/tests/kit_sync_cron.test.sql
mei-artificerinnovations: Fixed — added test 15 asserting EXISTS (SELECT 1 FROM cron.job WHERE jobname = "kit-sync-worker") before the non-inlining check. Now the suite verifies both that the job was registered and that the secret is not in its command.

supabase/tests/kit_sync_cron.test.sql
mei-artificerinnovations: Acknowledged — the direct read in test 14 is correct as a superuser assertion. With the service_role table grant now removed, test 17 also explicitly asserts service_role cannot SELECT, covering that posture.

mei-artificerinnovations (commented):
Rebased onto `develop`, PR retargeted, and all review feedback addressed in a single clean commit (ad05441).

---

### What changed

**Security (blocking)**
- Removed `GRANT SELECT, INSERT, UPDATE ... TO service_role`. The SECURITY DEFINER function accesses the table as its owner — no table grant is needed, and omitting it prevents direct REST reads of `cron_secret` via the service key. Comment added above the REVOKE block to document the intent. Test 17 now asserts `service_role` has no direct SELECT.

**Migration nits**
- `CHECK (singleton IS TRUE)` — explicit form, matches test filter.
- Cron body consolidated from two subqueries to a single `FROM public.kit_sync_runtime_config c WHERE c.singleton IS TRUE LIMIT 1` row reference.
- Best-effort `RESET app.kit_cron_secret` inside a nested `BEGIN...EXCEPTION WHEN OTHERS` block — clears the stale GUC from the prior migration; silently no-ops if `ALTER DATABASE RESET` is also blocked on hosted Supabase.

**Test nits**
- Test 12: `authenticated` has no SELECT (mirrors the `anon` check; plan 14 → 17).
- Test 15: `cron.job` row exists after `kit_sync_setup_cron` (was only checking secret non-inlining).
- Test 17: `service_role` has no direct SELECT.
- `PERFORM cron.unschedule('kit-sync-worker')` added before `finish()` to clean up any cron.job rows committed independently by pg_cron.

PR #392, opened 2026-06-01
feat(auth): sync OAuth display name and avatar into user_profiles

ZappoMan:
## Summary

- Add DB triggers to copy Google (and Apple) OAuth display name and profile photo from `auth.users.raw_user_meta_data` into `user_profiles.display_name` and `user_profiles.avatar_url`
- Seed both fields on first signup; backfill empty profile fields on later OAuth sign-in without overwriting user edits in ProfileEditor
- Extend pgTAP coverage (22 assertions) and document the mapping in `docs/OAUTH.md`
- Add `provider` prop and accessibility labels to `SocialLoginButton`

## Problem

Google social sign-in authenticates users but only partially seeds `user_profiles`: `display_name` may fall back to email and `avatar_url` is never set. The UI reads `user_profiles` only, so users see initials/email instead of their OAuth name and photo.

## Test plan

- [x] `npm run test:db` (migration filenames + pgTAP, including `test_user_creation_trigger.sql` — 22/22)
- [x] `SocialLoginButton` unit tests (5/5)
- [x] Pre-commit type-check
- [ ] Manual: Google sign-in on web → header shows Google photo + name
- [ ] Manual: edit display name on `/profile` → sign out/in with Google → edited name preserved
- [ ] Manual: mobile Google sign-in → same profile row updated

### DIFF
--- a/supabase/migrations/20260531120200_oauth_profile_sync.sql
+++ b/supabase/migrations/20260531120200_oauth_profile_sync.sql
@@ -1,0 +1,84 @@
+-- Sync OAuth provider display name and avatar into user_profiles on signup and sign-in.
+-- Fill-empty-only: never overwrite user-edited profile fields.
+
+CREATE OR REPLACE FUNCTION public.oauth_profile_from_metadata(meta jsonb)
+RETURNS TABLE (display_name text, avatar_url text)
+LANGUAGE sql
+IMMUTABLE
+SET search_path = public
+AS $$
+  SELECT
+    NULLIF(trim(COALESCE(meta->>'full_name', meta->>'name')), '') AS display_name,
+    NULLIF(trim(COALESCE(meta->>'avatar_url', meta->>'picture')), '') AS avatar_url;
+$$;
+
+REVOKE ALL ON FUNCTION public.oauth_profile_from_metadata(jsonb) FROM PUBLIC;
+
+CREATE OR REPLACE FUNCTION public.handle_new_user()
+RETURNS TRIGGER AS $$
+DECLARE
+  oauth_profile record;
+BEGIN
+  SELECT * INTO oauth_profile
+  FROM public.oauth_profile_from_metadata(NEW.raw_user_meta_data);
+
+  INSERT INTO public.user_profiles (user_id, username, display_name, avatar_url)
+  VALUES (
+    NEW.id,
+    public.generate_username(),
+    COALESCE(oauth_profile.display_name, NEW.email),
+    oauth_profile.avatar_url
+  );
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql SECURITY DEFINER SET search_path = public;
+
+CREATE OR REPLACE FUNCTION public.sync_oauth_profile_from_auth()
+RETURNS TRIGGER AS $$
+DECLARE
+  oauth_profile record;
+  profile_display_name text;
+  profile_avatar_url text;
+  resolved_display_name text;
+BEGIN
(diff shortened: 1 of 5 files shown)


supabase/migrations/20260531120200_oauth_profile_sync.sql
```diff
@@ -0,0 +1,84 @@
...
+
+CREATE OR REPLACE FUNCTION public.oauth_profile_from_metadata(meta jsonb)
+RETURNS TABLE (display_name text, avatar_url text)
+LANGUAGE sql
+IMMUTABLE
+SET search_path = public
+AS $$
+  SELECT
+    NULLIF(trim(COALESCE(meta->>'full_name', meta->>'name')), '') AS display_name,
+    NULLIF(trim(COALESCE(meta->>'avatar_url', meta->>'picture')), '') AS avatar_url;
```
sarah-artificerinnovations, line 12: ✅ NULLIF(trim(...)) correctly normalises whitespace-only strings (e.g. "  ") to NULL, consistent with handle_new_user's COALESCE(email) fallback.

supabase/migrations/20260531120200_oauth_profile_sync.sql
```diff
@@ -0,0 +1,84 @@
...
+CREATE OR REPLACE FUNCTION public.sync_oauth_profile_from_auth()
+RETURNS TRIGGER AS $$
+DECLARE
+  oauth_profile record;
+  profile_display_name text;
+  profile_avatar_url text;
+  resolved_display_name text;
+BEGIN
+  IF NEW.raw_user_meta_data IS NOT DISTINCT FROM OLD.raw_user_meta_data THEN
+    RETURN NEW;
```
sarah-artificerinnovations, line 45: ✅ IS NOT DISTINCT FROM is the right guard here — handles NULL = NULL correctly (unlike =), so a user whose metadata was never set (both OLD and NEW are NULL) correctly skips the profile update.

sarah-artificerinnovations (approved):
## Migration review — OAuth display name / avatar sync

Clean, focused migration. The design is correct.

**Billing deployment failure** — I pulled CI for the current commit (`cd894a9`) and all 38 checks pass. The earlier failure appears to have been transient or fixed in this commit; no action needed on that front.

---

**Migration `20260531120200_oauth_profile_sync.sql`**

- `oauth_profile_from_metadata` correctly handles the Google (`full_name`/`picture`) and Apple (`name`/`avatar_url`) metadata keys, with REVOKE ALL from PUBLIC. IMMUTABLE is correct — pure function of the input jsonb.
- `handle_new_user` updated with CREATE OR REPLACE — existing `on_auth_user_created` trigger wiring is preserved without re-creation. OAuth fields seeded on first signup; email fallback for display_name consistent with original behavior.
- `sync_oauth_profile_from_auth`: fill-empty-only semantics (two independent NULL guards), `IS NOT DISTINCT FROM` early exit prevents spurious updates on unrelated `auth.users` modifications, `NOT FOUND` guard handles profileless users. SECURITY DEFINER + SET search_path on both trigger functions.
- DROP TRIGGER IF EXISTS before CREATE TRIGGER — idempotent.

**Nit (non-blocking):** Two separate UPDATE statements for display_name and avatar_url could be combined into one. The two-UPDATE approach is correct as written; this is style only.

**pgTAP** — 22 assertions covering helper edge cases (empty metadata, fallback keys), email signup flow, Google signup seeding, backfill-without-overwrite, skip-on-unchanged-metadata, and user-edit preservation. Comprehensive for the feature scope.

PR #393, opened 2026-06-01
feat: add @beakerstack/connections package

ZappoMan:
## Summary
- Adds `@beakerstack/connections` (Supabase RPCs, React hooks, search/row UI components) ported from SkeinAI `feat/beakerstack-connections` with generic Beaker Stack template content.
- Wires template integration: protected `/connections` page, profile discoverability settings, user menu link, Vite alias, and regenerated database types.
- Adds four Supabase migrations + pgTAP suite, CI coverage shard, classify/merge-coverage paths, and spec doc `docs/specs/beakerstack-connections-v1.md`.

## Test plan
- [x] `npm run test:unit:connections` (56 tests)
- [x] `COVERAGE_MERGE=1 npm run test:coverage:connections` (100% thresholds)
- [x] `apps/web` ConnectionsPage tests (4 tests)
- [x] `npm run type-check` (full monorepo)
- [x] `./scripts/ci/classify-pr-push-changes.sh --self-test`
- [ ] CI supabase pgTAP (`connections.test.sql`) on PR run
- [ ] Manual: sign in as two users, search/connect/accept/disconnect on `/connections`


supabase/migrations/20260601160000_bs_connections_v1.sql
```diff
@@ -0,0 +1,859 @@
...
+                THEN public.bs_connections.disconnected_at
+            ELSE NULL
+        END,
+        blocked_by_user_id = CASE
+            WHEN public.bs_connections.status = 'blocked'
+                THEN public.bs_connections.blocked_by_user_id
+            ELSE NULL
+        END
+    WHERE public.bs_connections.status NOT IN ('blocked', 'accepted', 'declined')
+       OR (
```
sarah-artificerinnovations, line 393: ✅ Terminal-state guard is correct — `WHERE status NOT IN ('blocked', 'accepted', 'declined')` on the ON CONFLICT prevents overwriting terminal states in the narrow race between SELECT FOR UPDATE (NOT FOUND) and INSERT. This is the right approach.

supabase/migrations/20260603140000_connection_discoverability_enum.sql
```diff
@@ -0,0 +1,383 @@
...
+UPDATE public.user_profiles
+SET connection_discoverability = CASE
+    WHEN is_discoverable = false THEN 'username_only'::public.connection_discoverability
+    ELSE 'searchable'::public.connection_discoverability
+END
+WHERE connection_discoverability IS NULL
+  AND EXISTS (
+      SELECT 1
+      FROM information_schema.columns
+      WHERE table_schema = 'public'
```
sarah-artificerinnovations, line 27: ✅ Portable backfill guard: `information_schema.columns` check ensures the UPDATE only runs when `is_discoverable` actually exists. Safe to re-run on a fresh schema that never had the boolean column.

supabase/migrations/20260603140000_connection_discoverability_enum.sql
```diff
@@ -0,0 +1,383 @@
...
+WHERE connection_discoverability IS NULL;
+
+ALTER TABLE public.user_profiles
+    ALTER COLUMN connection_discoverability
+    SET DEFAULT 'searchable'::public.connection_discoverability;
+
+ALTER TABLE public.user_profiles
+    ALTER COLUMN connection_discoverability
+    SET NOT NULL;
+
```
sarah-artificerinnovations, line 43: ✅ Correctly sequenced: column added nullable (line 17), fully backfilled (lines 19–35), then SET NOT NULL here. No risk of constraint violation mid-migration.

sarah-artificerinnovations (approved):
## Migration review — @beakerstack/connections port

Sound migration approach throughout. Confirming the four key correctness points.

**Migrations**

1. `20260601155900_user_profiles_is_discoverable.sql` — safe additive: `NOT NULL DEFAULT true` correctly defaults existing rows to discoverable; partial index on `is_discoverable = true` matches the query filter.

2. `20260601160000_bs_connections_v1.sql` — correct schema and state machine: `user_low`/`user_high` GENERATED ALWAYS pair-uniqueness; `IF NOT EXISTS` throughout; SECURITY DEFINER + SET search_path on all RPCs; internal helpers REVOKE ALL from PUBLIC; RLS blocks hidden on both table and SECURITY DEFINER belt-and-suspenders.

3. `20260602130000_connections_search_exact_username.sql` — CREATE OR REPLACE preserves existing grants; no DDL; idempotent.

4. `20260603140000_connection_discoverability_enum.sql` — data migration is correctly sequenced (see inline). CREATE TYPE EXCEPTION handler makes it idempotent. Discoverability enforcement in `connections_request` and `connections_search_users` is correct for all three enum values.

**Non-blocking observation**

In `connections_request`'s INSERT path (reached only when SELECT FOR UPDATE returns NOT FOUND), if the ON CONFLICT `WHERE` clause prevents the update due to a concurrent terminal-state transition in the narrow window between SELECT and INSERT, `RETURNING * INTO v_row` yields NULL. The subsequent audit write then records `connection_id = NULL` and the function returns a NULL-status row. This is an extremely narrow race (requires a concurrent session to both insert and transition to a terminal state in that window), not a security issue, and the `WHERE` guard itself is the correct fix for terminal-state protection. Worth noting for future hardening, but not a blocker.

**pgTAP coverage** — 23 tests covering request/accept, mutual auto-accept, decline and flip, disconnect re-request, block/unblock, expired pending renewal, rate limit, RLS isolation, search filtering, and all three discoverability enum values. Comprehensive.

PR #399, opened 2026-06-03
Fix AI code quality findings in setup, help, and mobile profile

ZappoMan:
## Summary
- Remove stale `@param {CliFlags} flags` JSDoc from `ghAuthOk` and `easWhoamiOk` in `setup-full.mjs`, and fix main-module detection to compare resolved paths via `__filename`.
- Parse help markdown with `marked.lexer` + `marked.parser` so build output is always a synchronous HTML string (avoids `Promise`/`[object Promise]` edge cases and satisfies TypeScript).
- Store lazy-loaded `ProfileEditor` in component state instead of a module-level mutable, and drop the unused `navigation` prop from `ProfileScreenContent`.

## Test plan
- [x] `cd packages/help && npx tsc --noEmit`
- [x] `cd packages/help && npx vitest run build-help.test.ts`
- [x] Pre-commit hook type-check passed (mobile, web, shared, help, etc.)

### DIFF
--- a/apps/mobile/src/screens/ProfileScreen.tsx
+++ b/apps/mobile/src/screens/ProfileScreen.tsx
@@ -24,7 +24,6 @@ import { ProfileStats } from '@beakerstack/shared/components/profile/ProfileStat
 // @ts-ignore - Dynamic imports are supported by Metro, TypeScript error is a false positive
 import type { ProfileEditorProps } from '@beakerstack/shared/components/profile/ProfileEditor.native';
 import { loadProfileEditorModule } from './profileEditorLoader';
-let ProfileEditor: React.ComponentType<ProfileEditorProps> | null = null;
 
 type RootStackParamList = {
   Home: undefined;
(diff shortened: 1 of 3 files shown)


sarah-artificerinnovations (approved):
## Review — PR #399 (AI code quality findings fix)

No blockers. Approved.

---

**`apps/mobile/src/screens/ProfileScreen.tsx`**

Module-level `let ProfileEditor = null` mutable replaced with component state via `useState`. The `setProfileEditor(() => module.ProfileEditor)` functional updater form is the right call here — React would otherwise invoke a component type (a function) as a state initializer, producing the wrong value. `componentsLoaded` sentinel collapsed into the null-check on `ProfileEditor` itself. `_navigation` prop dropped from `ProfileScreenContent` since it was never used. ✅

**`packages/help/scripts/build-help.ts`**

`marked.parse(body) as string` → `marked.parser(marked.lexer(body))`. `marked.parse()` can return `Promise<string>` when async hooks are active; the `as string` cast was hiding the type issue. `marked.lexer()` and `marked.parser()` are always synchronous, so the output is a guaranteed `string`. ✅

**`scripts/setup-full.mjs`**

Stale `@param {CliFlags} flags` JSDoc removed from `ghAuthOk` and `easWhoamiOk` — both are zero-argument functions. Main-module guard changed from `import.meta.url === url.pathToFileURL(process.argv[1] || '').href` to `process.argv[1] && path.resolve(process.argv[1]) === __filename`. Confirmed `__filename` is defined at line 106 as `url.fileURLToPath(import.meta.url)` — the comparison is correct and the `argv[1]` null-guard is an improvement over the prior `|| ''` fallback. ✅

PR #403, opened 2026-06-24
fix(deps): resolve npm audit to zero vulnerabilities

ZappoMan:
## Summary
- Bumps patched direct dependencies (`ws`, `dompurify`) and expands root npm overrides for transitive Expo/EAS and dev-tooling packages (tar, esbuild, js-yaml, joi, form-data, and others).
- Adds Skein-aligned `patch-package` fixes for `gray-matter`, `read-yaml-file`, and `cosmiconfig` so js-yaml 4.2.0 works without rewriting article frontmatter parsing.
- Regenerates `package-lock.json`; `npm audit` now reports **0 vulnerabilities**.

## Test plan
- [x] `npm audit` — 0 vulnerabilities
- [x] `npm run test:unit:articles` — 73 passed
- [x] `npm run test:unit:help` — 42 passed
- [x] `npm run type-check:articles` — passed
- [ ] CI green on PR

### DIFF
--- a/patches/gray-matter+4.0.3.patch
+++ b/patches/gray-matter+4.0.3.patch
@@ -1,0 +1,15 @@
+diff --git a/node_modules/gray-matter/lib/engines.js b/node_modules/gray-matter/lib/engines.js
+index 38f993d..1dfec8d 100644
+--- a/node_modules/gray-matter/lib/engines.js
++++ b/node_modules/gray-matter/lib/engines.js
+@@ -13,8 +13,8 @@ const engines = exports = module.exports;
+  */
+ 
+ engines.yaml = {
+-  parse: yaml.safeLoad.bind(yaml),
+-  stringify: yaml.safeDump.bind(yaml)
++  parse: yaml.load.bind(yaml),
++  stringify: yaml.dump.bind(yaml)
+ };
+ 
+ /**
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,7 +41,7 @@
         "tsup": "^8.0.0",
         "tsx": "^4.19.4",
         "typescript": "^5.3.0",
-        "ws": "^8.20.1"
+        "ws": "^8.21.0"
       },
       "engines": {
         "node": ">=20.19.0",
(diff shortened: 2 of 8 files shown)


patches/gray-matter+4.0.3.patch
```diff
@@ -0,0 +1,15 @@
+diff --git a/node_modules/gray-matter/lib/engines.js b/node_modules/gray-matter/lib/engines.js
+index 38f993d..1dfec8d 100644
+--- a/node_modules/gray-matter/lib/engines.js
++++ b/node_modules/gray-matter/lib/engines.js
+@@ -13,8 +13,8 @@ const engines = exports = module.exports;
+  */
+ 
+ engines.yaml = {
```
sarah-artificerinnovations, line 8: Non-blocking — `load`/`dump` are the correct js-yaml v4 replacements (the `safe` prefix was dropped as misleading; the behavior is identical). If gray-matter ever releases 4.0.4+, check whether they've fixed this themselves before keeping the patch.

sarah-artificerinnovations (approved):
**Triage: fast** / Reasoning: security dep sweep, no logic changes, CI fully green.

Approved. Comprehensive sweep — audit goes from N vulnerabilities to zero and CI is clean.

**Direct bumps:** `ws ^8.21.0`, `dompurify ^3.4.11` — both minor patch increments, straightforward.

**npm overrides** for transitive deps (`shell-quote`, `tar`, `js-yaml`, `joi`, `esbuild`, `form-data`, `ts-deepmerge`) — right approach when you can't bump the consuming packages. The nested vite/vite-node → esbuild overrides are a bit gnarly but correct.

**patch-package patches** for `gray-matter`, `cosmiconfig`, `read-yaml-file` — all three just swap `safeLoad → load` / `safeDump → dump` which js-yaml v4 removed as deprecated aliases. The semantics are identical for safe YAML parsing. Minimal and correct.

**`@types/dompurify` removal** — right call, dompurify ships its own types since 3.x.

One thing: the PR body shows `[ ] CI green on PR` unchecked, but CI is actually all green (38 checks, all passed). Probably opened before CI ran.

Maintenance note (non-blocking): the three patch-package patches add a small ongoing cost — if gray-matter/cosmiconfig/read-yaml-file ever see new releases, patches will need re-evaluation. Given how dormant those packages are, low risk in practice.

PR #405, opened 2026-06-30
feat: port VIP comp grants and waitlist-billing integration

ZappoMan:
## Summary

- Add complimentary billing grants foundation: `billing_comp_grants` table, admin grant/revoke RPCs, hidden `beakerstack_vip` plan (Max-equivalent), and `comped` billing UI state
- Add `@beakerstack/waitlist-billing` integration package for per-invite VIP provisioning at waitlist conversion time
- Wire admin waitlist UI (invite, approve, retroactive save), fix `waitlist-ops` consume to fulfill on `already_converted` (OAuth path), and add ops CLI scripts

## Test plan

- [ ] Apply migrations locally (`supabase db reset`) and run `./scripts/test-db-pgtap.sh`
- [ ] Admin → Users: grant/revoke complimentary VIP; verify `comped` status and billing UI banner
- [ ] Admin → Waitlist: invite/approve with VIP + reason; verify user auto-comped on signup
- [ ] Admin → Waitlist: Save VIP intent on approved entry; verify retroactive fulfillment
- [ ] Set `WAITLIST_COMP_PLAN_IDS=beakerstack_vip` on deployed `waitlist-ops` edge function
- [ ] Run `npm run db:apply-adopter` for `beakerstack_vip` plan seed
- [x] Unit tests pass for admin, billing, waitlist, waitlist-billing packages


sarah-artificerinnovations (commented):
**Sarah — product/velocity lens.** Well-spec'd, the spec doc + deploy checklist are genuinely good, and the SQL idempotency (the `unchanged` short-circuits in `waitlist_billing_fulfill_conversion`) is the right call. Leaving correctness/perf/UX to Rahul/Mei/Diego. Two product concerns and a scope note:

**1. Silent provisioning failure is the real risk here.** Your own deploy checklist documents it: config drift → `plan_not_allowed` at conversion "while admin invite/approve still succeeds." For a feature whose entire job is comping design partners, that's the failure mode that'll actually bite — you promise someone VIP, the invite shows green, they sign up, fulfill no-ops or errors, and they land on Free with *nobody alerted*. Two asks:
  - In `waitlist-ops`, `plan_not_allowed`/`invalid_plan` at conversion currently get `console.error`'d and lumped into a generic 500. Surface those distinctly (alert/structured log) — a misconfigured allowlist failing every conversion should be loud.
  - Admin has no way to confirm it worked. The waitlist drawer shows that *intent was set*, not that provisioning *succeeded*. Post-conversion, show the outcome — "VIP provisioned ✓" / "⚠ provisioning failed."

**2. Two sources of truth for the comp-plan allowlist.** `compPlanIds` in `adopter/config/waitlist-billing.ts` + `WAITLIST_COMP_PLAN_IDS` edge env. You've documented the sync requirement honestly, but docs aren't a mechanism — this drifts the first time someone's in a hurry, and the symptom is silent conversion failures (see #1). Can the edge derive the allowlist from the same config, or can fulfill just trust `is_public=false AND product_id matches` and drop the separate env allowlist? Fewer knobs, fewer silent failures.

**3. Scope (non-blocking).** Comp-grants foundation + waitlist-billing integration, 63 files / ~4.4k lines. They're genuinely coupled so I won't push a split — but the foundation (table, RPCs, comped UI) is independently testable and lower-risk. If this needs iteration rounds, landing the foundation first shrinks the blast radius. Your call.

**4. Test plan — this is where I'd hold.** Only the unit-test box is checked. Every integration box (migrations applied, grant/revoke, auto-comp on signup, retroactive fulfill) is unchecked — and that end-to-end conversion path is exactly what delivers the value and exactly what units can't cover. Before merge I'd want signup→auto-comp and retroactive-fulfill actually run once against a real `db reset`, with output pasted here.

None of this is blocking your other three reviewers — it's product-shaped, not correctness-shaped. Happy to re-look once the e2e path is exercised.

packages/admin/src/adminClient.ts
```diff
@@ -119,6 +119,52 @@ export async function revokeOperator(
...
+  expiresAt?: string | null;
+};
+
+export async function grantBillingComp(
+  supabase: SupabaseClient,
+  params: GrantBillingCompParams
+): Promise<void> {
+  const { data, error } = await supabase.rpc('admin_grant_billing_comp', {
+    p_user_id: params.userId,
+    p_product_id: params.productId,
```
diego-artificerinnovations, line 147: **blocking:** `params.expiresAt ?? undefined` coerces `null` to `undefined`. When the caller explicitly passes `null` (meaning no expiry), that intent is silently dropped — the RPC param just disappears instead of being `NULL`. The type is `string | null | undefined`; use `?? null` to preserve the distinction:
```ts
p_expires_at: params.expiresAt ?? null,
```
The test for `grantBillingComp` only exercises the case where `expiresAt` is a string, so this isn't caught there.

apps/web/src/admin/components/AdminUserDetailDrawer.web.tsx
```diff
@@ -107,6 +152,48 @@ export function AdminUserDetailDrawer({
   const isSelf =
     userId !== null && currentUserId !== null && userId === currentUserId;
 
+  const hasCompAccess =
```
diego-artificerinnovations: **blocking:** Two sources of truth for the same fact. `detail.comp_grant` and `detail.subscription.status === 'comped'` can transiently disagree — e.g. if the grant was recently revoked but the subscription snapshot hasn't refreshed, you'd show "Revoke complimentary access" when the grant is gone. Since `comp_grant` is the canonical field (it's `null` when no active grant exists), I'd drop the subscription-status fallback:
```ts
const hasCompAccess = Boolean(detail?.comp_grant);
```
If there's a scenario where `comp_grant` can be absent while `status` is `'comped'`, that case needs a comment explaining it.

packages/admin/src/adminClient.ts
```diff
@@ -119,6 +119,52 @@ export async function revokeOperator(
...
+}
+
+export async function revokeBillingComp(
+  supabase: SupabaseClient,
+  userId: string,
+  productId: string,
+  reason?: string
+): Promise<void> {
+  const { data, error } = await supabase.rpc('admin_revoke_billing_comp', {
+    p_user_id: userId,
```
diego-artificerinnovations: nit: `grantBillingComp` takes a params object; `revokeBillingComp` takes positional args. Paired operations should have a consistent call signature. When a `revokedBy` or `expiresAt` param gets added to revoke later, the positional pattern becomes a footgun. Consider a params object:
```ts
type RevokeBillingCompParams = { userId: string; productId: string; reason?: string };
export async function revokeBillingComp(supabase: SupabaseClient, params: RevokeBillingCompParams)
```

apps/web/src/admin/components/AdminInviteByEmailPanel.web.tsx
```diff
@@ -115,31 +157,45 @@ export function AdminInviteByEmailPanel({
...
-          disabled={busy || !email.trim()}
-          className='shrink-0 rounded-md bg-indigo-600 px-4 py-2 text-sm font-medium text-white hover:bg-indigo-700 disabled:opacity-50'
-        >
-          {busy ? 'Sending…' : 'Send invite'}
-        </button>
+
+        <WaitlistVipInviteFields
+          defaultCompPlanId={
+            waitlistBillingConfig.defaultCompPlanId ?? 'beakerstack_vip'
+          }
```
diego-artificerinnovations: nit: `waitlistBillingConfig.defaultCompPlanId` is required in `WaitlistBillingConfig` — the `?? 'beakerstack_vip'` fallback implies it can be absent, but it can't. Drop the fallback so a misconfiguration fails loudly rather than silently using a hardcoded plan ID:
```tsx
defaultCompPlanId={waitlistBillingConfig.defaultCompPlanId}
```
Same pattern in `AdminWaitlistDetailDrawer`.

diego-artificerinnovations (changes requested):
Solid PR — the overall shape is clean and the test coverage is good (parity test is a nice touch). Two blocking items around a null-coercion bug and a fragile derived state; a few nits below.

FE/DX lens: the `comped` state is handled consistently across billing UI (overview, plans, badge, label utility), which is exactly what I want to see. Error messages are user-facing and specific. The `mapProvisioningError` / `compErrorMessage` helpers are appropriately thin.

mei-artificerinnovations (changes requested):
Reviewed from an infrastructure and reliability lens. CI is green, the overall design is solid. Four issues worth addressing before merge.

---

### 1. Missing row lock in `_billing_apply_comp_grant` — concurrent conversion will 500

**File:** `supabase/migrations/20260630120200_billing_apply_comp_grant_internal.sql`

The subscription SELECT has no `FOR UPDATE`:
```sql
SELECT * INTO v_sub
FROM public.billing_subscriptions s
WHERE s.user_id = p_user_id AND s.product_id = p_product_id;
```
Under a race (same user, two tabs, sign-up from two devices at the same time):
- Both reads pass the `stripe_subscription_id IS NOT NULL` guard
- Both proceed to supersede the existing grant and INSERT a new one
- The second INSERT hits the `billing_comp_grants_active_user_product` unique index → exception propagates as a 500 to the user at the critical sign-up moment

Fix: add `FOR UPDATE` to the subscription SELECT, or at minimum add `AND stripe_subscription_id IS NULL` to the subscription UPDATE so the window collapses.

Practical likelihood is low (users very rarely convert from two devices simultaneously), but the failure mode is a 500 at the worst possible moment.

---

### 2. Comp grant UPDATE can stomp a concurrently-created Stripe subscription

**File:** `supabase/migrations/20260630120200_billing_apply_comp_grant_internal.sql`

The `billing_subscriptions` UPDATE does not guard against `stripe_subscription_id`:
```sql
UPDATE public.billing_subscriptions s
SET plan_id = p_plan_id,
    status = comped,
    stripe_subscription_id = NULL,  -- clears it
    ...
WHERE s.user_id = p_user_id AND s.product_id = p_product_id;
-- missing: AND s.stripe_subscription_id IS NULL
```
The Stripe check runs at SELECT time, but between SELECT and UPDATE a Stripe webhook could write `stripe_subscription_id`. The revoke path (`admin_revoke_billing_comp`) correctly guards with `AND s.stripe_subscription_id IS NULL`. This path should match.

---

### 3. `inviteWaitlistEmail` third-argument is a breaking API change

**File:** `packages/waitlist/src/waitlistClient.ts`

Old signature: `inviteWaitlistEmail(supabase, email, metadata?: Record<string, unknown>)`
New signature: `inviteWaitlistEmail(supabase, email, options?: { metadata?, provisioningIntent?, ... })`

Any existing caller passing `{ note: vip }` as the third arg now sets `options = { note: vip }`, but `options.metadata` is `undefined`, so `p_metadata: {}` is silently sent instead. No type error at call sites that were passing a flat metadata object directly — TypeScript won’t catch this because a `{ note: string }` satisfies `{ metadata?: ..., provisioningIntent?: ... }` structurally.

This should be in the PR description as a migration note for adopters, and ideally caught with a stricter options type (`metadata` required, or a discriminated union).

---

### 4. `admin_grant_billing_comp` defined twice across sequential migrations

**Files:** `supabase/migrations/20260630120100_admin_billing_comp_rpcs.sql` and `20260630120200_billing_apply_comp_grant_internal.sql`

120100 defines the full inline version; 120200 immediately overwrites it with a thin wrapper around `_billing_apply_comp_grant`. Both work when applied in order, but this will confuse anyone reading migration history and’s redundant weight in `supabase db reset`. Consider dropping the 120100 implementation and only defining the function once in 120200.

---

### Notes (not blocking)

- **`WAITLIST_COMP_PLAN_IDS` default hardens `beakerstack_vip`**: if this plan doesn’t exist in a product at conversion time, `waitlist_billing_fulfill_conversion` returns `invalid_plan` → 500 for the user at sign-up. The test plan calls this out (`npm run db:apply-adopter` + env var), but worth a runtime guard or a clearer error code distinction (so the edge function can fall through to the default plan instead of 500ing).
- **`already_converted` now re-runs `fulfillWaitlistConversion`**: intentional for idempotency (confirmed by the PR description), and the SQL has the right idempotency checks. Just noting it’s a behavioral change vs. the previous path that skipped billing entirely on repeat conversion.

CI is all green. Issues 1 and 2 are the ones I’d want addressed before this ships to production.

diego-artificerinnovations:
Adding one more blocking item following Mei's review:

**blocking (via Mei #3): `inviteWaitlistEmail` third-arg shape change is a silent breaking change**

The TypeScript compiler won't catch this because the old `metadata?` and new `options?` shapes are structurally compatible — callers passing `{ note: 'vip' }` as metadata silently pass it as options with no type error. Two paths to fix:
- Use a branded or discriminated type so the shapes are explicitly incompatible at compile time
- Add a runtime assertion that rejects unknown keys in `options`

If neither fits the timeline, the minimum bar is: (a) PR description calls it out as a breaking change for callers, and (b) the `AdminInviteByEmailPanel` tests have a case that asserts `provisioningIntent` is actually forwarded to the RPC — not just that the call succeeds.

apps/web/src/admin/components/AdminInviteByEmailPanel.web.tsx
```diff
@@ -77,12 +117,14 @@ export function AdminInviteByEmailPanel({
...
+      await sendInviteEmail(link, result.email, result.entry_id);
       await emitLifecycleEvent('waitlist.approved', {
         email: result.email,
         entryId: result.entry_id,
       });
       setEmail('');
+      setProvisioningIntent(null);
+      setVipTouched(false);
       onInvited?.();
     } catch (err) {
```
rahul-artificerinnovations, line 132: `setVipEnabled(false)` is missing here. After a successful invite, `vipEnabled` stays `true` but `provisioningIntent` is reset to `null`. The next submission hits the guard at line 94 and blocks with "Enter a reason" even though no VIP grant was intended. Add `setVipEnabled(false)` in this success block, or key-reset `WaitlistVipInviteFields` to force remount.

supabase/migrations/20260630120200_billing_apply_comp_grant_internal.sql
```diff
@@ -0,0 +1,193 @@
...
+BEGIN
+    IF p_user_id IS NULL OR p_product_id IS NULL OR p_plan_id IS NULL THEN
+        RETURN jsonb_build_object('error', 'not_found');
+    END IF;
+
+    v_reason := nullif(btrim(p_reason), '');
+    IF v_reason IS NULL OR char_length(v_reason) > 500 THEN
+        RETURN jsonb_build_object('error', 'invalid_reason');
+    END IF;
+
```
rahul-artificerinnovations, line 35: This UPDATE has no `AND s.stripe_subscription_id IS NULL` guard — it will overwrite a real Stripe subscription ID with NULL if a webhook fires between the SELECT at line 22 and this UPDATE. The revoke path in migration 120100 correctly has that guard. Fix: add `AND s.stripe_subscription_id IS NULL` to the WHERE clause here.

rahul-artificerinnovations (changes requested):
**REQUEST_CHANGES** — correctness/failure-envelope lens.

Other reviews have the main blockers covered. Adding one new finding and specifics on the SQL gap.

---

**Blocker (new): `vipEnabled` not reset after successful invite in `AdminInviteByEmailPanel`**

After a successful submit, `setProvisioningIntent(null)` and `setVipTouched(false)` are called, but `setVipEnabled(false)` is not. On the next submission (same admin, next invitee), `vipEnabled` is still `true` and `provisioningIntent` is `null`, which fires the guard at line 94 (`if (vipEnabled && !provisioningIntent)`) with "Enter a reason when granting VIP access" — even though no VIP grant was intended. Admin is stuck until they manually toggle VIP off.

Fix: call `setVipEnabled(false)` in the success path alongside the other resets, or key-reset `WaitlistVipInviteFields` on success.

---

**Confirming Mei's #2 with the specific fix line**

`_billing_apply_comp_grant` (migration 120200) updates the subscription without guarding on `stripe_subscription_id IS NULL`. The revoke path in migration 120100 correctly has `AND s.stripe_subscription_id IS NULL` on its UPDATE WHERE clause. The grant path doesn't. The SELECT-check-UPDATE is an unguarded TOCTOU: a Stripe webhook writing between those two statements zeros out a real subscription ID with no audit trail. Fix: mirror the revoke guard on the grant UPDATE.

---

**Migration debt: `admin_grant_billing_comp` defined twice**

Migration 120100 defines `admin_grant_billing_comp` with 145 lines of inline logic. Migration 120200 immediately replaces it with a thin wrapper. The 120100 version never runs in production. It also carries the same unguarded UPDATE bug (moot since it's replaced), suggesting the guard fix wasn't applied during the refactor. Not a blocker, but worth noting before this migration pattern gets copied.

---

Agree with team: Diego's `?? null` fix and Mei's row-lock + stripe guard fixes are must-haves. The conversion path is the entire value surface of this feature and it currently has a race, a TOCTOU, and a UI state bug all on that path. Sarah's e2e request is the right gate.

ZappoMan:
## Review feedback addressed (`6568925`)

Thanks all — pushed fixes for the blocking items and nits called out inline. Summary by reviewer:

### Mei / Rahul (SQL concurrency + TOCTOU)
- `_billing_apply_comp_grant`: subscription row is now locked with `SELECT … FOR UPDATE`, and the comp UPDATE requires `stripe_subscription_id IS NULL` (returns `stripe_subscription_active` if the guarded update affects zero rows).
- Removed duplicate `admin_grant_billing_comp` body from migration `120100`; grant RPC is defined once in `120200`.

### Rahul / Copilot (AdminInviteByEmailPanel VIP state)
- Success path now calls `setVipEnabled(false)` and bumps a `vipFormKey` to remount `WaitlistVipInviteFields`, so the next invite is not blocked by stale VIP state.

### Diego (blocking + nits)
- `grantBillingComp`: `p_expires_at: params.expiresAt ?? null` (explicit null preserved).
- `hasCompAccess` uses `detail.comp_grant` only (single source of truth).
- `revokeBillingComp` now takes a params object (`RevokeBillingCompParams`).
- Dropped `?? 'beakerstack_vip'` fallbacks in admin waitlist UI (config is required).

### Copilot (admin client + pgTAP + API)
- `grantBillingComp` / `revokeBillingComp` throw on any RPC `{ error }` payload (not just a whitelist).
- pgTAP asserts VIP plan parity vs `beakerstack_max` features/limits.
- `inviteWaitlistEmail` third arg: exported `InviteWaitlistEmailOptions` + runtime `normalizeInviteWaitlistEmailOptions()` for legacy bare-metadata callers; new callers should pass `{ metadata: … }`.

### Sarah (product — partial this round)
- `waitlist-ops` now logs structured JSON for fulfill failures and returns distinct `plan_not_allowed` / `invalid_plan` errors (vs generic `plan_provision_failed`) so allowlist drift is louder in logs/monitoring.
- Post-conversion “VIP provisioned ✓” admin UI and allowlist de-duplication (edge env vs adopter config) are noted — happy to follow up in a small PR if we want those before merge.

### Verification (local, post-fix)
- pgTAP: **15/15 files, 267/267 assertions** after `supabase db reset`
- `@beakerstack/admin` + `@beakerstack/waitlist` unit tests green
- `waitlist-flow` integration: **5/5** passing

packages/admin/src/adminClient.ts
```diff
@@ -119,6 +119,52 @@ export async function revokeOperator(
...
+  });
+  if (error) throw rpcError(error);
+  if (isNotFound(data)) throw new Error('not_found');
+  const code = errorCode(data);
+  if (code === 'invalid_reason') throw new Error('invalid_reason');
+  if (code === 'invalid_plan') throw new Error('invalid_plan');
+  if (code === 'stripe_subscription_active') {
+    throw new Error('stripe_subscription_active');
+  }
+}
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 6568925: `throwIfRpcPayloadError` now throws on any RPC `{ error }` payload after the `not_found` check, so unhandled codes like `no_free_plan` cannot slip through as success.

packages/admin/src/adminClient.ts
```diff
@@ -119,6 +119,52 @@ export async function revokeOperator(
...
+  const { data, error } = await supabase.rpc('admin_revoke_billing_comp', {
+    p_user_id: userId,
+    p_product_id: productId,
+    p_reason: reason ?? undefined,
+  });
+  if (error) throw rpcError(error);
+  if (isNotFound(data)) throw new Error('not_found');
+  const code = errorCode(data);
+  if (code === 'no_free_plan') throw new Error('no_free_plan');
+}
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 6568925: same `throwIfRpcPayloadError` helper applied to `revokeBillingComp` — any JSON error code now throws.

apps/web/src/admin/components/AdminInviteByEmailPanel.web.tsx
```diff
@@ -77,12 +117,14 @@ export function AdminInviteByEmailPanel({
...
-      await sendInviteEmail(link, result.email);
+      await sendInviteEmail(link, result.email, result.entry_id);
       await emitLifecycleEvent('waitlist.approved', {
         email: result.email,
         entryId: result.entry_id,
       });
       setEmail('');
+      setProvisioningIntent(null);
+      setVipTouched(false);
       onInvited?.();
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 6568925: success path resets `vipEnabled`, clears intent/touched, and increments `vipFormKey` to remount `WaitlistVipInviteFields`.

supabase/tests/billing_comp_grants.test.sql
```diff
@@ -0,0 +1,333 @@
...
+SELECT ok(
+    (SELECT (features ->> 'byok_enabled')::boolean FROM public.billing_plans WHERE id = 'beakerstack_vip'),
+    'beakerstack_vip includes byok_enabled'
+);
+
+SELECT is(
+    (SELECT (usage_limits ->> 'thought_capture_byok')::int FROM public.billing_plans WHERE id = 'beakerstack_vip'),
+    -1,
+    'beakerstack_vip includes unlimited thought_capture_byok meter'
+);
```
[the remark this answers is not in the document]
ZappoMan: Fixed in b4f3763/6568925: pgTAP now compares `beakerstack_vip` features/usage_limits directly to `beakerstack_max` instead of Skein-specific fields.

supabase/migrations/20260630120200_billing_apply_comp_grant_internal.sql
```diff
@@ -0,0 +1,193 @@
...
+        p_reason,
+        v_uid,
+        p_expires_at,
+        'admin'
+    );
+END;
+$$;
+
+REVOKE ALL ON FUNCTION public.admin_grant_billing_comp(uuid, text, text, text, timestamptz) FROM PUBLIC;
+GRANT EXECUTE ON FUNCTION public.admin_grant_billing_comp(uuid, text, text, text, timestamptz) TO authenticated;
```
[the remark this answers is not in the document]
ZappoMan: Fixed in 6568925: removed the inline `admin_grant_billing_comp` implementation from migration 120100; grant RPC is defined once in 120200 (wrapper around `_billing_apply_comp_grant`).

packages/admin/src/adminClient.ts
ZappoMan: Fixed in 6568925: `p_expires_at: params.expiresAt ?? null`. Added unit test asserting explicit `null` is forwarded to the RPC.

apps/web/src/admin/components/AdminUserDetailDrawer.web.tsx
ZappoMan: Fixed in 6568925: `hasCompAccess = Boolean(detail?.comp_grant)` — dropped subscription-status fallback.

packages/admin/src/adminClient.ts
ZappoMan: Fixed in 6568925: `revokeBillingComp` now uses `RevokeBillingCompParams` (params object), matching `grantBillingComp`.

apps/web/src/admin/components/AdminInviteByEmailPanel.web.tsx
ZappoMan: Fixed in 6568925: removed `?? 'beakerstack_vip'` in `AdminInviteByEmailPanel` and `AdminWaitlistDetailDrawer`; uses required `waitlistBillingConfig.defaultCompPlanId`.

apps/web/src/admin/components/AdminInviteByEmailPanel.web.tsx
ZappoMan: Fixed in 6568925: added `setVipEnabled(false)` plus `vipFormKey` remount on successful invite (same fix as Copilot thread).

supabase/migrations/20260630120200_billing_apply_comp_grant_internal.sql
ZappoMan: Fixed in 6568925: comp grant UPDATE now includes `AND s.stripe_subscription_id IS NULL`; subscription row is also locked with `FOR UPDATE` before the check/update.

diego-artificerinnovations (approved):
All my blocking items resolved. Approving.

**Fixed (my items):**
- `expiresAt ?? undefined` → `?? null` ✓
- `hasCompAccess` single source of truth (`comp_grant` only) ✓
- `revokeBillingComp` now takes `RevokeBillingCompParams` object, consistent with grant ✓
- `?? 'beakerstack_vip'` fallbacks dropped from both callers ✓

**Fixed (team items via this PR):**
- `vipEnabled` + `vipFormKey` reset on successful invite (Rahul #1) ✓
- `FOR UPDATE` on subscription SELECT in 120200 (Mei #1) ✓
- `stripe_subscription_id IS NULL` guard + `NOT FOUND` on the UPDATE, matching revoke path (Mei #2) ✓
- Legacy `inviteWaitlistEmail` metadata callers preserved via `normalizeInviteWaitlistEmailOptions` + test (Mei #3) ✓
- Migration 120100 deduped, grant RPC consolidated into 120200 ✓
- `waitlist-ops` now returns `plan_not_allowed`/`invalid_plan` as distinct error codes with structured logging (Sarah allowlist-drift concern) ✓
- Complimentary access banner added to plans page (last commit) ✓

**One observation (not blocking):** `throwIfRpcPayloadError` guards with `code !== 'not_found'`, but `not_found` is already thrown before that call — the guard is redundant but harmless. The new helper is otherwise strictly better than the original per-code checks.

Good fixes. The conversion path is solid now.

mei-artificerinnovations (approved):
All four of my blocking items are addressed. Approving from the infrastructure/reliability lens.

**Verified fixes:**
- `FOR UPDATE` added to subscription SELECT in `_billing_apply_comp_grant` (race condition closed)
- `AND s.stripe_subscription_id IS NULL` guard added to the subscription UPDATE, with an explicit `IF NOT FOUND` post-check (TOCTOU closed, now matches the revoke path)
- `inviteWaitlistEmail` backward-compat shim (`normalizeInviteWaitlistEmailOptions`) preserves legacy callers without type unsafety
- Migration 120100 stripped of the duplicate `admin_grant_billing_comp` — one definition, one place
- `expiresAt ?? null` fixed
- `plan_not_allowed` / `invalid_plan` now returned as distinct error codes with structured logging (`severity: config_mismatch`) — Sarah's alerting concern addressed

**CI:** green on the fix commit (65689256) — db-tests, supabase-tests, integration all passed. Latest commit (billing plans banner) has unit tests green; DB/integration checks not triggered for a TS-only change, expected.

**Bonus:** billing plans banner for comped users is the right thing to show in that state.

No remaining concerns from me. The SQL fixes are precisely what I asked for.

sarah-artificerinnovations (commented):
**Sarah — re-review (product/velocity).** The turnaround looks good. Confirming my items:

- **Silent-failure alerting — resolved, and cleanly.** `waitlist-ops` now emits a structured `waitlist_billing_fulfill_conversion_failed` log with `severity: config_mismatch` for `plan_not_allowed`/`invalid_plan` and surfaces the specific error instead of a generic 500. That's exactly the operator-facing signal I wanted — a misconfigured allowlist is now loud instead of a partner silently landing on Free.
- **Allowlist dual-source (TS config + `WAITLIST_COMP_PLAN_IDS` env) — mitigated, not eliminated.** With the loud failure above I'm fine shipping it as-is for v1; the drift now fails visibly at conversion rather than silently. Collapsing to one exported constant both the seed and the edge consume is worth a follow-up issue, not a merge blocker.
- **Admin provisioning-outcome visibility** (showing whether a conversion actually got comped, not just that intent was set) — not in this pass. Also fine as a follow-up; the alerting covers the urgent version of it.

All the cross-team blockers I was tracking landed too — `?? null`, `hasCompAccess` single-source, `FOR UPDATE` + stripe guard, the `inviteWaitlistEmail` back-compat shim, and the `vipEnabled` reset.

**One thing still open from me — the e2e gate.** The pgTAP (267 assertions) + integration suites now exercise the SQL conversion logic, which covers a lot of what I was worried about. But every manual box in the test plan is still unchecked — specifically *invite-with-VIP → user signs up → auto-comped* and *retroactive fulfill on an approved entry*. That end-to-end wiring (edge → fulfill → billing state) is the value surface and the part the unit/pgTAP layers can't fully prove. Before this merges I'd like either those boxes checked or a one-line confirmation you ran the two conversion paths against a real `db reset`.

Minor: the preview deploy comment shows **💳 Billing ❌ Failed** — probably stale from an earlier commit, but since this PR is billing-heavy, worth a glance to confirm it's unrelated.

Not blocking your other reviewers' re-reviews. Everything I raised is addressed or acceptably deferred — just want the conversion path confirmed before it goes in. (Merge is yours regardless.)

rahul-artificerinnovations (approved):
**APPROVE** — all my blockers and the team's are resolved.

**What landed in the fix commits:**

- `vipEnabled` + `vipFormKey` reset on successful invite ✓ (my bug, clean fix)
- `expiresAt ?? null` ✓
- `hasCompAccess` reduced to `comp_grant` single source ✓
- `revokeBillingComp` params object (consistent with grant) ✓
- `?? 'beakerstack_vip'` fallbacks dropped ✓
- `FOR UPDATE` on subscription SELECT in `_billing_apply_comp_grant` ✓ — serializes concurrent calls; the idempotency check handles the second caller cleanly
- `stripe_subscription_id IS NULL` guard on the UPDATE + `IF NOT FOUND` error path ✓ — belt-and-suspenders with the row lock; correct
- 120100 dead code removed, grant RPC consolidated into 120200 wrapper ✓
- `inviteWaitlistEmail` backward compat via `normalizeInviteWaitlistEmailOptions` ✓
- `plan_not_allowed` surfaced distinctly with structured logging + severity field ✓
- Plans page banner for comped accounts ✓
- pgTAP tests now do full `features` / `usage_limits` equality against `beakerstack_max` rather than spot-checking one field — strictly better ✓

**Remaining non-blocker:** The split allowlist (TS config vs. edge env `WAITLIST_COMP_PLAN_IDS`) is mitigated — errors are now surfaced and logged — but the structural fix (single exported constant consumed by both) is outstanding. Fine to track as follow-on; the silent-failure mode is gone.

From my end: clean. Sarah's e2e gate still stands before Brad merges.

mei-artificerinnovations (approved):
Re-reviewed BeakerStack #405 (68321ac6). All review concerns from both rounds are now addressed.

**SQL/reliability (my original blockers):**
- `FOR UPDATE` on subscription SELECT in `_billing_apply_comp_grant` ✅
- `AND s.stripe_subscription_id IS NULL` guard + `IF NOT FOUND THEN RETURN 'stripe_subscription_active'` ✅
- `expiresAt ?? null` in `grantBillingComp` ✅
- `revokeBillingComp` params object (not positional args) ✅

**New in this round:**
- `BillingPlansPage` comp banner — was missing, now added with test ✅
- `hasCompAccess = Boolean(detail?.comp_grant)` single source of truth ✅
- `vipFormKey` reset on success ✅
- `defaultCompPlanId` fallback `?? 'beakerstack_vip'` removed ✅
- `waitlist_billing_fulfill_conversion` now explicitly revokes from `authenticated` and `anon` in addition to `REVOKE ALL FROM PUBLIC` — belt-and-suspenders, correct ✅
- pgTAP now does a DB-level `SELECT is(features ..., features ...)` comparison: `beakerstack_vip` features === `beakerstack_max` features from the actual `billing_plans` table. This closes the SQL JSONB drift gap — the invariant is now enforced end-to-end, not just at the TypeScript layer ✅

Approved.

rahul-artificerinnovations (approved):
Reviewed commit 68321ac6. Clean addition.

Comped banner added to `BillingPlansPage` using the existing `Banner` component (`variant='info'`), gated on `isComped`. Message is correct: "You have complimentary access. Plan changes and Stripe checkout are not available on this account." Test verifies it appears in the comped subscription state.

No concerns. Approved.

sarah-artificerinnovations (approved):
**Sarah — re-review of 68321ac6 (comped banner on plans page). Approving.**

This closes a small product gap in my lane: a comped user on the plans page already had disabled CTAs, but no *explanation*. Now they get the `isComped` info banner — "You have complimentary access. Plan changes and Stripe checkout are not available on this account." — so the disabled buttons read as intentional rather than broken. Gated on `isComped`, tested, and consistent with the overview banner and the same treatment we shipped on Glidepath #11. No concerns.

Issue #299, opened 2026-05-20
Automate Sentry CI/CD setup with setup:sentry wizard (follow-up to #288)


ZappoMan:
## Summary

Follow-up to #288 / PR #288. v1 shipped `@beakerstack/observability`, local web wiring (`VITE_SENTRY_DSN` in `.env.local`), and docs for manual GitHub secret + workflow configuration. Adopters who enable Sentry today must hand-copy DSNs, create auth tokens, edit three deploy workflows, fix environment tagging, and add source-map upload steps themselves.

This issue adds a **setup wizard** (`npm run setup:sentry`, plus optional phase in `npm run setup:full`) that automates **CI/CD Sentry configuration end-to-end** for preview, staging, and production web deploys — matching the ergonomics of existing phases (`setup-stripe`, GitHub manifest sync, etc.).

**Scope for this ticket:** steps **2, 3, 4, and 6** from the manual deployment checklist (GitHub secrets, workflow env wiring, environment tagging, source maps). Local `.env.local` DSN setup remains covered by existing docs; mobile/EAS is out of scope (see #295 and a future mobile follow-up).

**Parent:** #288 (observability package)

## Business reasons

| Pain today | Impact |
|------------|--------|
| Manual GitHub secret creation across preview/staging/production | High setup friction; adopters skip Sentry or misconfigure one environment |
| `vite build` sets `MODE=production` for all deploys | Preview/staging/prod errors collapse into one Sentry environment — noisy alerts, wrong triage |
| No source map upload in CI | Production issues show minified stacks; slower MTTR, lower trust in observability |
| Secrets/workflow drift from template | Forks diverge; upgrades from upstream BeakerStack miss new Sentry keys |
| Inconsistent with BeakerStack’s “setup wizard” product promise | Observability feels bolted-on vs Stripe/Supabase/AWS phases |

**Goal:** An adopter with a Sentry account runs one wizard flow and ends with GitHub Actions secrets synced, deploy workflows passing the correct build-time vars, distinct Sentry environments per deploy tier, and source maps uploaded on every web deploy — without editing workflow YAML by hand.

## Why this is feasible now

- Sentry exposes REST APIs for **projects**, **DSNs**, and **internal auth tokens** (org auth token with `project:releases` scope for `@sentry/cli`).
[...]

PR #406, opened 2026-07-21
ci: wire Sentry into web deploys

ZappoMan:
## Summary
- Port Skein's Sentry CI wiring into the BeakerStack template so adopters get exception reporting without a separate wizard step
- Deploy workflows pass `VITE_SENTRY_*` for staging, production, and PR preview
- Add `@sentry/vite-plugin` for staging/production source map upload (`filesToDeleteAfterUpload`)
- Update `docs/observability.md` + `env.example` (closes the “until #299” gap for web deploys)

## Adopter setup
```bash
gh secret set STAGING_SENTRY_DSN
gh secret set PRODUCTION_SENTRY_DSN
gh secret set SENTRY_AUTH_TOKEN
gh variable set SENTRY_ORG --body '<org-slug>'
gh variable set SENTRY_PROJECT --body '<project-slug>'
```

Preview reuses `STAGING_SENTRY_DSN` with `VITE_SENTRY_ENVIRONMENT=preview` (same as Skein). CSP report-URI / enforced CSP Sentry hosts remain a separate follow-up.

## Test plan
- [ ] Set secrets/vars on a test adopter or BeakerStack itself
- [ ] Merge → staging deploy embeds DSN; force an error and check Sentry
- [ ] Confirm source maps upload when auth token + org/project vars are set
- [ ] Confirm local builds stay no-op without DSN

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -188,6 +188,9 @@ jobs:
       STACK_NAME: ${{ vars.PR_PREVIEW_STACK_NAME }}
       VITE_SUPABASE_URL: ${{ secrets.PRODUCTION_SUPABASE_URL }}
       VITE_SUPABASE_ANON_KEY: ${{ secrets.PRODUCTION_SUPABASE_ANON_KEY }}
+      VITE_SENTRY_DSN: ${{ secrets.PRODUCTION_SENTRY_DSN }}
+      VITE_SENTRY_ENVIRONMENT: production
+      VITE_SENTRY_RELEASE: ${{ github.sha }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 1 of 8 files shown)


ZappoMan:
Follow-up for deploy reliability (same as Glidepath #32): `@sentry/vite-plugin` now soft-fails source-map upload via `errorHandler` (warn + continue) so Sentry blips cannot block a ship.

sarah-artificerinnovations (approved):
Reviewed on my dimension (product/velocity) — **approve.** This is the template-side twin of Glidepath #32, and it already ships with the `errorHandler` soft-fail baked in (source-map upload logs-and-continues, so a bad token / Sentry outage never fails an adopter's staging/prod deploy). Good — that was my one substantive concern on the Glidepath side and it's pre-addressed here.

Because this lands in the **template**, the load-bearing property is "does nothing until an adopter opts in," and it holds:
- `enableSentrySourceMaps` requires all four of token/org/project/release; absent them the plugin isn't added and `sourcemap` stays `false`.
- DSN is only injected via workflow env from `*_SENTRY_DSN` secrets — unset secret = empty = SDK no-ops.
So an adopter who never touches Sentry gets normal builds with zero new failure surface. That's the right default for a template, and it closes the "until #299" wizard gap without forcing a setup step. Nice.

**Minor (defer to @mei, cosmetic):**
- Same duplicate `@sentry/cli` (2.39.1 nested under `bundler-plugin-core`) from pinning the plugin at `^2.23.1` (3.x exists) — since this is the template, every adopter inherits the second native-binary download in CI. Worth a look but not blocking.
- `package-lock.json` is +299/**-310** — the removals are unrelated version churn (a few `6.3.1`/`7.5.x` entries dropped), i.e. a full lockfile regen swept in some pruning beyond the Sentry adds. Benign, but worth a glance to confirm it's just regen and nothing intentional got reverted.

Test plan's unchecked because it depends on setting secrets/vars on a test adopter. CI green needs the human Actions-tab eyeball. Only @ZappoMan merges.

Issue #234, opened 2026-05-17
[bug]: Mobile billing screen missing AppHeader, shows back link instead


ZappoMan:
## What happened?

On the **mobile app**, the **Billing** screens do not show the standard app header bar (brand icon, title, user menu). Instead they show a **← Back** link at the top of the page.

This is inconsistent with other authenticated screens such as **Dashboard** and **Profile**, which render `AppHeader` and keep global navigation available from the header/user menu.

## Expected

Billing should use the same top navigation pattern as Dashboard and Profile: `AppHeader` with access to Home, Dashboard, Profile, Billing (via user menu), and sign-out — not a standalone back affordance as the primary chrome.

## Actual

`BillingLayout` renders a custom back control and page title, but no `AppHeader`:

- `apps/mobile/src/screens/billing/BillingLayout.tsx` — `Pressable` with `← Back` that calls `navigation.goBack()` / parent `goBack()`
- Billing overview/usage screens wrap content in `BillingLayout` instead of including `AppHeader`

By contrast:

- `apps/mobile/src/screens/DashboardScreen.tsx` — `<AppHeader supabaseClient={supabase} />`
- `apps/mobile/src/screens/ProfileScreen.tsx` — `<AppHeader supabaseClient={supabase} />`

Billing is also mounted via a nested stack (`apps/mobile/src/navigation/BillingNavigator.tsx`) under the root `Billing` route (`apps/mobile/src/navigation/AppNavigator.tsx`), which may reinforce a “pushed screen” feel compared to top-level Dashboard/Profile.

## Suggested fix

- Add `AppHeader` to billing screens (likely in `BillingLayout` or each billing screen), matching Dashboard/Profile layout.
- Remove or demote the `← Back` link unless there is a deliberate sub-flow that needs it (internal billing tabs probably should not require going “back” to leave billing).
- Align safe-area / scroll structure with Dashboard/Profile for visual consistency.
- Update `apps/mobile/__tests__/screens/billing/BillingScreens.test.tsx` if assertions depend on the back link.

## Steps to reproduce

1. Run the mobile app (`apps/mobile`) and sign in.
2. Open **Billing** from the user menu (same path as on web).
3. Observe: no `AppHeader`; **← Back** at top.
4. Navigate to **Dashboard** or **Profile** and compare: `AppHeader` is present with app branding and menu.

## Environment

- Platform: iOS / Android (React Native / Expo)
- Area: `apps/mobile` billing screens

PR #410, opened 2026-08-01
feat(deps): audit-zero React Router 8 via React 19 + Expo 57

ZappoMan:
## Summary
- Bump engines/CI/docs to **Node ≥22.22** so `react-router@8.3.0` and Expo 57 are installable in local + GitHub Actions.
- Upgrade monorepo to **React/React DOM 19.2.8** and **Expo SDK 57 / RN 0.86**, drop obsolete Expo 50 patches, and refresh Jest harnesses (`expo/virtual/env`, winter mocks, adopter-tests env).
- Replace `react-router-dom` with **`react-router@8.3.0`** (import codemod, remove v7 `future` flags) so root `npm audit` reports **0 vulnerabilities**.

## Dependency pinning
- Audit-related pins use exact versions in `overrides` (and root deps only when npm needs them to materialize the graph).
- `package-lock.json` + `npm ci` remain the install contract; CI runs `npm audit`.

## Test plan
- [x] `npm audit` → `found 0 vulnerabilities`
- [x] Typecheck: mobile, web, shared, admin, articles (+ pre-commit type-check suite)
- [x] Unit: web, admin, articles, mobile, shared-tests (web+native), adopter-tests
- [x] Smoke web routing: admin splat, basename/PR preview, auth redirects (manual on PR preview)
- [ ] Smoke mobile: `npx expo-doctor`, dev-client boot / critical auth flows
- [ ] Confirm Dependabot React Router alerts close after merge


sarah-artificerinnovations (commented):
**Review from Sarah (product/velocity lens)** — approving the *approach*; my one merge gate is runtime verification, not the code.

### Scope: one atomic PR is correct here (unusual for me to say)
My reflex on a 176-file PR is "split this," but these four moves are genuinely coupled — there's no green intermediate state:
- `react-router@8.3.0` needs the React 19 peer,
- Expo SDK 57 brings RN 0.86 (and lets us drop the obsolete `expo-dev-menu`/`react-native@0.73` patches — that's most of the -14k lines, not scary code deletion),
- Node ≥22.22 is required for both to even install.

Landing them separately would ship broken intermediate commits. So one PR is right **provided it merges as a single squash** — that's also our rollback story: if prod breaks, `git revert` the merge commit in one shot rather than untangling a stack.

### Blocking-for-merge: the whole PR's value rides on 3 unchecked boxes
99.27% coverage + typecheck prove the tree compiles and units pass with `MemoryRouter` mocks — they prove **nothing** about RR8 runtime behavior. RR7/8 changed relative-splat path resolution and removed the v6 `future` flags this PR drops; the codemod swaps imports but can't catch a behavior diff. The three unchecked items are exactly the ones that matter:
- [ ] Smoke web routing: admin splat, basename/PR-preview, **auth redirects**
- [ ] Smoke mobile: `expo-doctor`, dev-client boot / critical auth flows
- [ ] Confirm Dependabot RR alerts actually close post-merge

Shipping a routing migration that silently breaks auth redirects in order to close a Dependabot alert is a bad trade. **Please don't merge until those three are green.** Once they are, I'm happy.

### Non-blocking
1. **Durability of audit-zero.** React 19 + Expo 57 pulled a fresh transitive graph; `npm audit → 0` is a snapshot. Re-run after the lockfile settles post-merge. Related: the new direct devDeps (`asynckit`, `combined-stream`, `mime-types`, `es-set-tostringtag`, `@babel/highlight`) look like manual transitive pins to zero out audit — that works, but as bare devDependencies the intent is invisible and a future `npm i` can drift them. Prefer moving those into `overrides` (where "we pin this to kill CVE-X" is explicit) or a one-line note in the PR body.
2. Copilot's 3 nits are real and cheap — README still says React 18.2, admin README peer ranges still say `^6.0.0`, one combinable import. Batch them in.

Codemod itself: verified clean — zero `react-router-dom` references remain in source or manifests, dep correctly replaced with `react-router ^8.3.0`.

— Sarah

mei-artificerinnovations (changes requested):
This is a big-blast-radius migration (React 19, react-router 8, Expo SDK 57/RN 0.86, Node engine bump to ≥22.22 across CI) — 176 files, +4655/-14415. Infra side checks out: `deploy-production.yml`/`deploy-staging.yml` node-version bumped consistently in all 3 jobs, and CloudFront Functions (`infra/aws/*`) run their own `cloudfront-js-2.0` runtime so they're unaffected by the Node bump. No mismatch there.

But I traced why the PR's own test plan has "Smoke web routing: admin splat, basename/PR preview, auth redirects" unchecked, and it's not just an oversight — **the web PR preview never deployed for the current head commit.** Check-run `deploy-preview-web` is `skipped`, and the `preview-scope` job log (run [30723155531](https://github.com/Artificer-Innovations/BeakerStack/actions/runs/30723155531/job/91430148554)) shows why:

```
[INFO] classify a6fa14e6..05011a89: run_deploy=true app_code=true web=false mobile=true auth_sync=false
```

`pr-preview-environment.yml` classifies scope via `git diff before_sha..head_sha` (the *last push* only), not the full PR diff against `develop` — see `scripts/ci/classify-pr-event.sh`. The last push in this PR apparently only touched mobile-related files, so `web=false` and the web preview build was skipped, even though the cumulative PR diff clearly rewrites `apps/web` (react-router 6→8 import codemod, React 19). Net effect: there is currently no deployed preview reflecting this PR's actual web bundle, so the routing/auth-redirect smoke test the author flagged as needed literally cannot be run against current head right now.

RR splat-route and basename semantics changed between v6 and v8 (that's exactly the kind of thing a typecheck+unit-test pass won't catch, per the same lesson from #234's sharp discussion). For a migration this size I don't want to approve on green unit/coverage checks alone when the one check that would catch a routing regression silently didn't run.

Asking for, before merge:
1. Force a full-diff preview redeploy (push a no-op commit, or manually re-run `pr-preview-environment` against the full `develop...head` diff) so `deploy-preview-web` actually builds against current head.
2. Run the web + mobile smoke checklist against that real deployment.
3. Separately: worth a follow-up issue on `classify-pr-event.sh` itself — incremental-diff-only classification means any multi-commit PR whose *last* push happens to skip an app's files will silently skip that app's preview even when earlier commits in the same PR touched it heavily. That's a general footgun, not specific to this PR.

mei-artificerinnovations (changes requested):
This is a big-blast-radius migration (React 19, react-router 8, Expo SDK 57/RN 0.86, Node engine bump to ≥22.22 across CI) — 176 files, +4655/-14415. Infra side checks out: `deploy-production.yml`/`deploy-staging.yml` node-version bumped consistently in all 3 jobs, and CloudFront Functions (`infra/aws/*`) run their own `cloudfront-js-2.0` runtime so they're unaffected by the Node bump. No mismatch there.

But I traced why the PR's own test plan has "Smoke web routing: admin splat, basename/PR preview, auth redirects" unchecked, and it's not just an oversight — **the web PR preview never deployed for the current head commit.** Check-run `deploy-preview-web` is `skipped`, and the `preview-scope` job log (run [30723155531](https://github.com/Artificer-Innovations/BeakerStack/actions/runs/30723155531/job/91430148554)) shows why:

```
[INFO] classify a6fa14e6..05011a89: run_deploy=true app_code=true web=false mobile=true auth_sync=false
```

`pr-preview-environment.yml` classifies scope via `git diff before_sha..head_sha` (the *last push* only), not the full PR diff against `develop` — see `scripts/ci/classify-pr-event.sh`. The last push in this PR apparently only touched mobile-related files, so `web=false` and the web preview build was skipped, even though the cumulative PR diff clearly rewrites `apps/web` (react-router 6→8 import codemod, React 19). Net effect: there is currently no deployed preview reflecting this PR's actual web bundle, so the routing/auth-redirect smoke test the author flagged as needed literally cannot be run against current head right now.

RR splat-route and basename semantics changed between v6 and v8 (that's exactly the kind of thing a typecheck+unit-test pass won't catch, per the same lesson from #234's sharp discussion). For a migration this size I don't want to approve on green unit/coverage checks alone when the one check that would catch a routing regression silently didn't run.

Asking for, before merge:
1. Force a full-diff preview redeploy (push a no-op commit, or manually re-run `pr-preview-environment` against the full `develop...head` diff) so `deploy-preview-web` actually builds against current head.
2. Run the web + mobile smoke checklist against that real deployment.
3. Separately: worth a follow-up issue on `classify-pr-event.sh` itself — incremental-diff-only classification means any multi-commit PR whose *last* push happens to skip an app's files will silently skip that app's preview even when earlier commits in the same PR touched it heavily. That's a general footgun, not specific to this PR.

apps/mobile/jest/expo-virtual-env-mock.cjs
```diff
@@ -1,14 +1,19 @@
...
+    ? process.env
+    : {};
 
-module.exports = Object.assign(
-  { __esModule: true, default: d },
-  d
-);
+module.exports = {
+  __esModule: true,
+  env,
```
rahul-artificerinnovations, line 17: This mock now exports a named `env` object, but the real runtime shim `apps/mobile/shims/expo-virtual-env.js` (untouched by this PR) only has a default export — no named `env`. Expo SDK 57's babel-preset-expo auto-rewrites `process.env.EXPO_PUBLIC_*` reads to `import { env } from "expo/virtual/env"` (confirmed against the plugin source), so this passes under Jest but resolves `env` to `undefined` at real Metro runtime in dev/staging/preview builds (prod inlines literals and never imports this module). Recommend adding a matching named `env` export to the real shim, or reverting this mock to the shim's actual shape.

packages/shared/package.json
```diff
@@ -21,10 +21,10 @@
     "zod": "^3.22.0"
   },
   "peerDependencies": {
-    "react": "18.2.0",
-    "react-dom": "18.2.0",
-    "react-native": "0.73.x",
-    "react-router-dom": "^6"
+    "react": "19.2.8",
+    "react-dom": "19.2.8",
+    "react-native": ">=0.73.0",
```
rahul-artificerinnovations: Nit: `react-native` peerDependency widened from pinned `"0.73.x"` to unbounded `">=0.73.0"` in the same PR bumping the app to RN 0.86.0 — no peer-dep warning would catch a consumer on an older RN. Consider `>=0.86.0` if intentional, otherwise non-blocking.

rahul-artificerinnovations (changes requested):
Reviewed the diff (176 files, focused on the 83-file source subset excluding lockfile/snapshots/docs) with emphasis on routing, auth, and Expo/RN native config given the unchecked web-routing and mobile-auth smoke tests in the test plan.

**One confirmed correctness bug** in the mobile env-var plumbing (inline comment below) — worth fixing or explicitly smoke-testing before merge, since it lands exactly in the blast radius this PR's own test plan flags as unverified.

I also ran down two other plausible-looking leads and want to note them as **refuted**, not just silently dropped:
- `basename` + absolute `<Navigate>` interaction under RR8 in `PublicShell.tsx`/auth guards: no documented behavior change from v6→v8 for basename-prepending on absolute navigate targets — v7/v8 changelog explicitly calls this a minimal breaking-change surface. Not a real risk.
- `packages/shared/tsconfig.json` `moduleResolution: node → bundler` combined with `declaration: true`: not actually restricted in TS 5.x (this repo's floor is `^5.3.0`) — `bundler` mode has supported declaration emission since it was introduced in TS 5.0. No compile-time conflict.

Not independently re-verified but worth calling out since they sit in the same unchecked-smoke-test blast radius: AsyncStorage 1.21→2.2 (New Architecture requirements), Google Sign-In 13→16 (3 majors, no visible call-site diff in this PR), and RN 0.73→0.86 — none of these show a concrete bug in the diff itself, but they're exactly what "smoke web routing / mobile auth" in the test plan should be catching before merge.

---

**apps/mobile/jest/expo-virtual-env-mock.cjs:17** — Mock now exports a named `env` object (`{ __esModule: true, env, default: { env } }`), but the real Metro runtime shim it stands in for (`apps/mobile/shims/expo-virtual-env.js`, untouched by this PR) only does `export default env` — no named `env` export exists there.

Expo SDK 57's babel-preset-expo rewrites every `process.env.EXPO_PUBLIC_*` read via an auto-injected `import { env } from 'expo/virtual/env'` (confirmed against babel-preset-expo's `inline-env-vars` plugin source — this is implicit, no app code opts in explicitly). Against the real shim, that named import resolves to `undefined` at runtime in dev/staging/preview Metro bundles (production release builds inline literals at build time and never hit this module, so the blast radius is dev/CI/preview builds specifically — but that includes exactly the kind of environment a "smoke test before merge" would run in).

Net effect: Jest passes (mock supplies the named export) while a real dev/staging build silently gets `undefined` for every `EXPO_PUBLIC_*` var — e.g. `EXPO_PUBLIC_SUPABASE_URL`, breaking Supabase client init with no compile or lint signal. Recommend either updating `apps/mobile/shims/expo-virtual-env.js` to also export a named `env`, or reverting the mock to not diverge from the real shim's export shape.

**packages/shared/package.json:26** (nit, non-blocking) — `react-native` peerDependency range widened from pinned `"0.73.x"` to open-ended `">=0.73.0"` in the same PR that bumps the app's own RN to `0.86.0`. Other packages (`articles`, `connections`) already used `>=0.73.0` before this PR, so this just brings `shared` in line — but an unbounded floor means a consumer could install `shared` against a much older RN (e.g. 0.74) without a peer-dep warning catching the mismatch. Consider pinning to `>=0.86.0` to match the RN version this PR actually targets, or leave as-is if intentionally permissive for downstream consumers on older RN.

apps/mobile/jest/expo-virtual-env-mock.cjs
ZappoMan: Addressed in 3a6d35c: the Metro alias to the broken shim is removed (Expo SDK 57 supplies the real `expo/virtual/env`), and `apps/mobile/shims/expo-virtual-env.js` now exports a named `env` matching the Jest mock / babel rewrite shape as a fallback.

packages/shared/package.json
ZappoMan: Addressed in 3a6d35c: tightened `react-native` peerDependency to `>=0.86.0` on shared/observability/billing/connections to match Expo SDK 57 / RN 0.86.

ZappoMan:
### Reviewer follow-up

**Rahul** — both inline items addressed in `3a6d35c` (threads resolved):
- Metro no longer aliases over Expo’s real `expo/virtual/env`; shim also exports named `env` as fallback
- `react-native` peers tightened to `>=0.86.0` on shared/observability/billing/connections

**Mei** — web preview redeploy triggered on current head (`f9ba677` includes observability runtime → `run_web=true`). Classifier footgun tracked in #411.

**Lint** — fixed unused eslint-disable in `packages/observability/src/init.native.ts` (`f9ba677`); babysitting CI now.

ZappoMan:
### Update

- Fixed lint (`f9ba677`) and a flaky adopter `DashboardScreen` timeout (`b278118`).
- **Mei:** preview scope now classifies against `origin/<base>..head` (full PR) so last-push-only diffs cannot skip web/mobile — see #411. This push should deploy both web + mobile previews at current head.
- **Rahul:** env-shim + RN peer items remain addressed; threads resolved.

Babysitting CI on `b278118`.

ZappoMan:
### Smoke test (Mei)

Web PR preview at current head was manually smoke-tested (routing / auth redirects / general app behavior) — worked well. That covers the unchecked “smoke web routing” item from the review.

Mobile Expo 57 boot/auth was also exercised locally earlier on this PR (iOS + Android). Remaining Android Google Sign-In `DEVELOPER_ERROR` is a local debug-keystore SHA-1 config mismatch, not an upgrade regression.

ZappoMan:
### Dependency pinning (Sarah + policy)

Agreed — lockfile + explicit pins should be the contract, not floating caret “audit helpers.”

In `b278118` follow-up (this push):
- Exact-pinned the form-data tree helpers / `@babel/highlight` / `ws` (no `^`)
- Mirrored those pins in `overrides` (nested under `form-data` for `mime-types` so Metro’s 3.x isn’t forced down)
- Added CI job `npm-audit` (`npm audit --audit-level=low`)
- Documented the policy in `SECURITY.md`

Also marked **smoke web routing** checked in the PR body after your preview verification.

mei-artificerinnovations (approved):
Re-reviewed at head `90bf228`. My blocker is resolved — clearing CHANGES_REQUESTED.

`.github/workflows/pr-preview-environment.yml` now calls `classify-pr-push-changes.sh` against `origin/<base>..head` unconditionally instead of the incremental `before..head` diff, with a comment referencing a follow-up issue (#411) for the general classifier gap. Confirmed on this head: `deploy-preview-web`, `deploy-preview-infra`, `deploy-preview-finalize`, and `lighthouse` all show `completed success` (previously `skipped`), and the test plan's "Smoke web routing" box is now checked with a real preview to have run it against.

Also confirms rahul's bug is fixed properly, not papered over: rather than patching the Jest mock to match a broken shim, `metro.config.js` dropped the `expo/virtual/env` alias entirely so Metro resolves its own virtual module (matching what the SDK 57 babel rewrite expects), and `shims/expo-virtual-env.js` now exports the named `env` for any straggler import — root cause fixed, not just the test double.

And Sarah's overrides nit — `asynckit`/`combined-stream`/`es-set-tostringtag`/`mime-types` moved from bare top-level `devDependencies`/`overrides` pins into a scoped `form-data` override object, plus a dedicated `npm-audit` CI job added to `test.yml` to keep this from drifting silently. Good call adding that as a permanent gate rather than a one-time snapshot.

Remaining open items (mobile `expo-doctor`/dev-client boot smoke, Dependabot alert closure) are outside what I can verify from here — deferring to diego/rahul on those before merge, per their standing review. Approving on the infra/CI side.

mei-artificerinnovations (approved):
Re-reviewed at head `90bf228`. My blocker is resolved — clearing CHANGES_REQUESTED.

`.github/workflows/pr-preview-environment.yml` now calls `classify-pr-push-changes.sh` against `origin/<base>..head` unconditionally instead of the incremental `before..head` diff, with a comment referencing a follow-up issue (#411) for the general classifier gap. Confirmed on this head: `deploy-preview-web`, `deploy-preview-infra`, `deploy-preview-finalize`, and `lighthouse` all show `completed success` (previously `skipped`), and the test plan's "Smoke web routing" box is now checked with a real preview to have run it against.

Also confirms rahul's bug is fixed properly, not papered over: rather than patching the Jest mock to match a broken shim, `metro.config.js` dropped the `expo/virtual/env` alias entirely so Metro resolves its own virtual module (matching what the SDK 57 babel rewrite expects), and `shims/expo-virtual-env.js` now exports the named `env` for any straggler import — root cause fixed, not just the test double.

And Sarah's overrides nit — `asynckit`/`combined-stream`/`es-set-tostringtag`/`mime-types` moved from bare top-level `devDependencies`/`overrides` pins into a scoped `form-data` override object, plus a dedicated `npm-audit` CI job added to `test.yml` to keep this from drifting silently. Good call adding that as a permanent gate rather than a one-time snapshot.

Remaining open items (mobile `expo-doctor`/dev-client boot smoke, Dependabot alert closure) are outside what I can verify from here — deferring to diego/rahul on those before merge, per their standing review. Approving on the infra/CI side.

sarah-artificerinnovations (approved):
**Re-review from Sarah** — everything I flagged is resolved; approving from the product/velocity lens.

- **Audit durability (my main non-blocker):** better than I asked. New `npm-audit` CI job runs `npm audit --audit-level=low` as a gate (`test.yml`), so audit-zero can't silently regress on a future `npm i`. And the bare transitive devDeps are now exact-pinned *and* scoped under `form-data` in `overrides` — intent is explicit ("pinned to kill the form-data chain CVE") instead of invisible. (Minor: they're now duplicated across direct devDeps + top-level overrides + the scoped `form-data` block — harmless, and the CI gate makes it self-verifying, so not worth churning.)
- **Web routing / auth (my hard gate last round):** `deploy-preview-web` now builds **success** against the full PR diff — Mei's classifier fix (`b278118d`) closed the preview-skip gap, so the routing/basename/auth-redirect smoke was actually runnable, and the test-plan box is checked against a real preview. Good.
- **Scope:** unchanged from last round — still correctly one atomic, revertable PR.

**Sole remaining merge gate is not mine:** the mobile dev-client boot + auth smoke box is still unchecked. `deploy-preview-mobile` going green proves it *builds*, not that it *boots and authenticates* on RN 0.86 — that's Diego/Rahul's blast radius (Expo 50→57, AsyncStorage 1→2, Google Sign-In 13→16). I also see Rahul's confirmed `expo/virtual/env` bug looks fixed — the shim now does `export { env }` (named) alongside the default, matching the SDK-57 babel rewrite — but I'll defer to Rahul to close his own finding.

My concerns are cleared. Don't merge until the mobile smoke runs and Rahul re-confirms.

— Sarah

rahul-artificerinnovations (approved):
Re-reviewed at head `90bf228e` (previous review was at `05011a89`). My blocking finding is fixed, and the other gaps the team raised look genuinely addressed rather than just checked off — verified each:

- **My CONFIRMED bug** (`apps/mobile/jest/expo-virtual-env-mock.cjs` named-export mismatch): fixed properly. `apps/mobile/shims/expo-virtual-env.js` now exports `{ env }` to match the mock's shape. Even better — `metro.config.js`'s `resolveRequest` alias that pointed `expo/virtual/env` at the shim file was removed entirely, so Metro now resolves the real virtual module directly and the shim is just a documented fallback. That's a root-cause fix, not a patch on the symptom.
- **My nit** (`packages/shared/package.json` unbounded `react-native` peerDep): fixed, now `>=0.86.0` matching the app's actual RN version.
- **mei's blocker** (preview never deployed — classifier scoped off last-push, not full PR diff): fixed. New `scripts/ci/classify-pr-push-changes.sh` diffs `base..head` cumulatively (with its own `--self-test` suite), referenced from `pr-preview-environment.yml`. Confirmed on the actual check-runs for this head: `deploy-preview-web` and `deploy-preview-mobile` both completed successfully this time.
- **sarah's ask** (manual transitive pins as bare devDeps): fixed, moved into `overrides` (including a properly nested `form-data` override), consistent with the PR's own "Dependency pinning" note.
- Web routing smoke test checkbox is now checked in the PR description.

**Still open, non-blocking:** the mobile smoke-test checkbox (`expo-doctor`, dev-client boot / critical auth flows) is still unchecked — given this is a 7-major Expo jump (50→57) touching auth persistence (AsyncStorage 1→2) and sign-in (google-signin 13→16), I'd still want that run before merge, same ask diego made originally. Diego's `metro-config/private/defaults/exclusionList` fragility note also still applies (unchanged, now with a comment explaining why) — already flagged, not re-blocking on it.

Approving — the code-level correctness issue I raised is resolved and the verification-gap issues the team raised are resolved with real fixes, not just re-runs. Merge is gated on the team's judgment call re: the still-unchecked mobile smoke test.

diego-artificerinnovations (commented):
Re-reviewed the latest revision against the three CHANGES_REQUESTED threads plus my own earlier nit. All check out:

- **rahul's `expo/virtual/env` bug**: fixed, and more completely than the minimum — not just the jest mock (`apps/mobile/jest/expo-virtual-env-mock.cjs` now exports a named `env`), but `apps/mobile/metro.config.js` no longer aliases `expo/virtual/env` to the custom shim in real builds, letting Metro's native SDK-57 virtual module handle it. The fallback shim (`shims/expo-virtual-env.js`) was updated to match the same `{ env, default: { env } }` shape for anywhere still pointing at it. Test mock and runtime behavior now agree.
- **rahul's `react-native` peerDep nit**: tightened from unbounded `>=0.73.0` to `>=0.86.0` across shared/observability/billing/connections.
- **mei's preview-classification bug**: `.github/workflows/pr-preview-environment.yml` now calls `classify-pr-push-changes.sh` (base..head, full PR diff) instead of the incremental `classify-pr-event.sh` for preview scope, so a later push touching only one app can no longer suppress another app's preview. Test workflow keeps incremental scoping, which is the right split. Follow-up issue #411 filed and closed as addressed-in-410, and the web preview was confirmed redeployed + manually smoke-tested at current head.
- **Sarah's overrides-visibility ask**: the bare devDeps (`@babel/highlight`, `asynckit`, `combined-stream`, `es-set-tostringtag`, `mime-types`, `ws`) are now exact-pinned (no caret) and mirrored in `overrides`, `form-data`'s sub-deps nested so Metro's own `mime-types` range isn't force-downgraded. New `SECURITY.md` documents the pinning policy and there's now a dedicated `npm-audit` CI job.

One thing still open, non-blocking: `apps/mobile/metro.config.js` still does `require('metro-config/private/defaults/exclusionList').default` — that's an undocumented/private Metro path with no compile-time guardrail if it moves again in a future Metro bump. Worth a one-line follow-up comment or issue, not worth holding this PR for.

Nothing else raised from my side. Good to go from the web/routing surface I reviewed.

PR #413, opened 2026-08-02
fix(ci): force decimal when bumping CalVer past .008

ZappoMan:
## Summary
- Release Template failed computing the next tag after `2026.008`: bash `$((008 + 1))` treats leading-zero values as octal (`value too great for base`).
- Force decimal with `10#${LAST_N}` so `2026.008` → `2026.009`.

## Test plan
- [x] Locally: `LAST_N=008` with `10#` yields `009`; `007` → `008`; empty → `001`
- [ ] Merge to `develop`, then promote / merge to `main`
- [ ] Re-run **Release Template** on `main` after the fix is there

### DIFF
--- a/.github/workflows/release-template.yml
+++ b/.github/workflows/release-template.yml
@@ -44,8 +44,9 @@ jobs:
             YEAR=$(date +%Y)
             # CalVer YYYY.NNN only (exclude semver-shaped tags like 2026.1.0)
             LAST_N=$(git tag -l | grep -E "^${YEAR}\\.[0-9]{3}$" | sed "s/${YEAR}\.//" | sort -n | tail -1)
-            NEXT_N=$(( ${LAST_N:-0} + 1 ))
-            printf -v PADDED_N "%03d" $NEXT_N
+            # 10# forces decimal: bare 008 is invalid octal in bash arithmetic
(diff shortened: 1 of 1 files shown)


sarah-artificerinnovations (approved):
**Approve.** Triage: fast — single-file CI fix, obvious, targets develop.

Correct root cause and correct fix. Bash arithmetic reads a leading-zero literal as octal, so `$((008 + 1))` blows up on 8/9 (`value too great for base`). `10#${LAST_N:-0}` forces base-10 — resolves it.

Verified the edge cases hold:
- `10#008`→8→009, `10#009`→9→010 (the ones that were failing) ✓
- empty `LAST_N` → `:-0` default preserved → `10#0`→0→001 ✓
- The `grep -E "^${YEAR}\.[0-9]{3}$"` upstream guarantees `LAST_N` is exactly 3 digits or empty, so there's no non-numeric input to worry about.
- Quoting `"$NEXT_N"` on printf is good hygiene — non-blocking bonus.

Can't confirm CI green independently (PAT lacks `checks:read`) — the real proof is re-running Release Template on main past `2026.008`, which your test plan already calls out.

Not merging (only @ZappoMan merges).

